PR #1, opened 2022-06-23
migrate DEFAULT_SERVER into config

zancas:


dannasessha (commented):
Looks pretty good!

I'm going to build locally and check out a little more of the "wiring" before merging, but should happen shortly.

zancas:
So, if you approve this code, can you please merge it @dannasessha ?

dannasessha:
Built locally for me, and worked, but I had to move aside `zingo-wallet.dat` (but that was also true when I tried to run from `dev` as well).

There were some warnings associated with the build, but I think that is understood and we're moving fast, we'll circle back to those. I am going to merge in.

PR #2, opened 2022-06-24
Even morchard

AloeareV:
We have orchard note decryption!!
It doesn't happen on sync/resync, only if an (orchard) action is received while zingo-cli is running in interactive mode, and all it does is print the memo to the terminal, but it's enough proof that we're able to parse orchard transactions for me to move on to focusing on actually getting the _wallet_ side of this working. Writing orchard keys into the wallet (they currently need to be regenerated after startup each time), writing received orchard data down, etc.


lib/proto/compact_formats.proto
```diff
@@ -37,20 +37,30 @@ message CompactTx {
...
-message CompactOutput {
+message CompactSaplingOutput {
     bytes cmu = 1;          // note commitment u-coordinate
     bytes epk = 2;          // ephemeral public key
-    bytes ciphertext = 3;   // ciphertext and zkproof
+    bytes ciphertext = 3;   // first 52 bytes of ciphertext
+}
+
+// https://github.com/zcash/zips/blob/main/zip-0225.rst#orchard-action-description-orchardaction
+// (but not all fields are needed)
```
zancas, line 60: So again...   omit as many fields as possible to be `Compact`?

lib/proto/service.proto
```diff
@@ -109,11 +110,12 @@ message Exclude {
...
-    string network = 1; // "main" or "test"
-    uint64 height = 2;
-    string hash = 3;    // block id
-    uint32 time = 4;    // Unix epoch time when the block was mined
-    string tree = 5;    // sapling commitment tree state
+    string network = 1;     // "main" or "test"
+    uint64 height = 2;      // block height
+    string hash = 3;        // block id
+    uint32 time = 4;        // Unix epoch time when the block was mined
+    string saplingTree = 5; // sapling commitment tree state
```
zancas, line 117: Doesn't this have implications for the Service on the other side of the wire?

lib/proto/service.proto
```diff
@@ -135,38 +137,13 @@ message GetAddressUtxosReplyList {
     repeated GetAddressUtxosReply addressUtxos = 1;
 }
```
zancas, line 139: We'll need to come up with a replacement story shortly...   as a short-term solution we could delegate this entirely into the front-end app software.

lib/proto/service.proto
```diff
@@ -193,6 +166,8 @@ service CompactTxStreamer {
     // in the exclude list that don't exist in the mempool are ignored.
     rpc GetMempoolTx(Exclude) returns (stream CompactTx) {}
 
+    // Return a stream of current Mempool transactions. This will keep the output stream open while
```
zancas, line 169: Nice commentary.

zancas (commented):
This is a preliminary review.   I will need to rereview (at least once) before I can approve.

lib/proto/compact_formats.proto
```diff
@@ -37,20 +37,30 @@ message CompactTx {
...
 
-    repeated CompactSpend spends = 4;   // inputs
-    repeated CompactOutput outputs = 5; // outputs
+    repeated CompactSaplingSpend spends = 4;   // inputs
+    repeated CompactSaplingOutput outputs = 5; // outputs
+    repeated CompactOrchardAction actions = 6;
 }
 
-// CompactSpend is a Sapling Spend Description as described in 7.3 of the Zcash
+// CompactSaplingSpend is a Sapling Spend Description as described in 7.3 of the Zcash
```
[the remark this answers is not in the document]
AloeareV: `bytes` doesn't have a specified size, the number is the order the fields will be serialized/deserialized in, if I remember my protobuf syntax right

lib/proto/compact_formats.proto
```diff
@@ -37,20 +37,30 @@ message CompactTx {
...
 // protocol specification.
-message CompactSpend {
+message CompactSaplingSpend {
     bytes nf = 1;   // nullifier (see the Zcash protocol specification)
 }
 
 // output is a Sapling Output Description as described in section 7.4 of the
```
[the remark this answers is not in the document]
AloeareV: I don't think a compact action contains all the information an action does. I think it's some minimum subset needed for certain things.. There's code in here for fetching full transactions, the compact transaction is something like just enough info to figure out if we need to ask for the full one.

lib/proto/compact_formats.proto
AloeareV: Yeah, that's my impression.

lib/proto/service.proto
AloeareV: The service on the other side would need to have the same proto file I think, yes (Although I don't think the names matter). This is the latest from `zcash/lightwalletd`, at the time it was committed.

zancas: OK...   so this isn't something you generated fresh..   I am going to make this a submodule, so we can just depend directly on upstream.

Cargo.toml
```diff
@@ -6,4 +6,8 @@ members = [
 ]
 
 [profile.release]
-debug = false
\ No newline at end of file
+debug = false
+
+[patch.crates-io]
+zcash_note_encryption = { git = "https://github.com/zcash/librustzcash", rev = "73d9395c9d3c5fa81fc7becd363a2c1a51772a76" }
+orchard = { git = "https://github.com/zingolabs/orchard", rev = "22b77c028f054d11455b60ac600787c3cdf74597" }
```
zancas, line 13: If we're patching to a `rev` then we need an intelligent human-involved form of awareness (like GH watch all activity).

zancas: Note...  for some cases a simple `Watch All` is too noisy.

lib/proto/service.proto
```diff
@@ -135,38 +137,13 @@ message GetAddressUtxosReplyList {
     repeated GetAddressUtxosReply addressUtxos = 1;
 }
 
-
-message PriceRequest {
```
zancas, line 139: We should keep the price stuff for now.

lib/src/blaze/fetch_full_transaction.rs
```diff
@@ -543,11 +547,63 @@ impl FetchFullTxns {
...
+                            Some(ret) => ret,
+                            None => continue,
+                        };
+                    let memo_bytes = MemoBytes::from_bytes(memo_bytes.as_slice()).unwrap();
+                    let memo = memo_bytes
+                        .clone()
+                        .try_into()
+                        .unwrap_or(Memo::Future(memo_bytes));
+                    println!("Recieved orchard tx with memo {:?}", memo);
+                    //This logic is all sapling specific! Todo: implement orchard version
```
zancas, line 586: This seems ready to refactor.

lib/src/blaze/fetch_full_transaction.rs
```diff
@@ -543,11 +558,83 @@ impl FetchFullTxns {
...
+    key_to_ivk: impl Fn(&K) -> Result<<D as Domain>::IncomingViewingKey, E>,
+    output_to_domain: impl Fn(&O) -> D,
+    memo_slice: impl Fn(&D::Memo) -> &[u8],
+) where
+    K: Clone,
+    for<'a> &'a B2: IntoIterator<Item = &'a O>,
+    D: Domain,
+    O: ShieldedOutput<D, ENC_CIPHERTEXT_SIZE>,
+{
+    let keys = key_reader(&*keys.read().await).clone();
```
zancas, line 615: `keys` as a variable name is a little tricky.

lib/src/blaze/fetch_full_transaction.rs
```diff
@@ -543,11 +558,83 @@ impl FetchFullTxns {
...
+        scan_bundle(
+            config,
+            transaction,
+            height,
+            unconfirmed,
+            block_time,
+            keys,
+            wallet_transactions,
+            is_outgoing_transaction,
+            outgoing_metadatas,
```
zancas, line 581: Oh...   I get it!   All the additional args are transformations represented as functions.   Very nice!

lib/src/wallet/keys.rs
```diff
@@ -535,6 +538,17 @@ impl Keys {
             .unwrap_or(false)
     }
 
+    pub fn have_orchard_spending_key(&self, ivk: &OrchardIncomingViewingKey) -> bool {
```
zancas, line 541: Issued abstracting this.

zancas: https://github.com/zingolabs/zingolib/issues/15

zancas (approved):

Issue #6, opened 2022-06-25
We need to update the version in `/cli/src/version.rs`


dannasessha:
There is a `CONST` defined in `/cli/src/version.rs` that hasn't been updated for 8 months. I don't know the details of where this is used yet, but we should update this and understand the implications.

dannasessha:
in `/cli/src/lib.rs` line 9, there is `pub mod version`, which is used on line 14.

I have confirmed this is the same string when calling `--help` with the cli invocation. I will likely adjust this string as part of a PR which is intended to support "regtest mode"

dannasessha:
This one got covered by #26 or #8 !

PR #8, opened 2022-06-25
WIP : Add regtest mode

dannasessha:
WIP!

Following up on [an older issue](https://github.com/zingolabs/zingolib_original/issues/85).

Solves #6 with first commit. 

<strike>Currently a non-functional flag that is recognized when running the compiled binary (ex, `./target/release/zingo-cli --regtest`).</strike>

### DIFF
--- a/cli/src/version.rs
+++ b/cli/src/version.rs
@@ -1,1 +1,1 @@
-pub const VERSION: &str = "1.7.7";
+pub const VERSION: &str = "0.1.1";
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ Zingo-CLI does automatic note and utxo management, which means it doesn't allow 
 
 ```
 git clone https://github.com/zingolabs/zingolib.git
+cd zingolib
 cargo build --release
 ./target/release/zingo-cli
 ```
--- a/README.md
+++ b/README.md
@@ -65,3 +65,46 @@ Here are some CLI arguments you can pass to `zingo-cli`. Please run `zingo-cli -
  * `--recover`: Attempt to recover the seed phrase from a corrupted wallet
  * `--data-dir`: uses the specified path as data directory.
     * Example: `./zingo-cli --server 127.0.0.1:9067 --data-dir /Users/ZingoRocks/my-test-wallet` will use the provided directory to store `zingo-wallet.dat` and logs. If the provided directory does not exist, it will create it.
+
+## Regtest
+This documentation to run regtest "manually" is a placeholder until the `--regtest` flag is fully functional.
+
+The CLI can work in regtest, by locally running a `zcashd` and `lightwalletd`.
+
+Create the directories `~/.zcash/zd/` and `~/.zcash/ld`
+`zcashd` works in regtest with the following invocation, run from your `zcash` directory (explicit path, substitute your own username):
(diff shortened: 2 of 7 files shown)


cli/src/version.rs
```diff
@@ -1 +1 @@
-pub const VERSION: &str = "1.7.7";
+pub const VERSION: &str = "0.1.1";
```
zancas, line 1: +1

dannasessha:
Posting here a message I've received: 
"Why not pass the zcashd and lightwalletd binaries to --regtest as arguments? Force a user to figure out where to get those binaries from.. that's not a problem that you need to solve inside this feature."

Sounds reasonable to me. I could probably easily also include default values that fail if they don't find something.

zancas:
Adding sane defaults, is beyond the scope of this task.   If the User doesn't provide binaries the `zingo-cli` should error out and exit.   The developer of this feature, @dannasessha shouldn't need to write any special logic to get this behavior.

AloeareV:
`FakeMainnet` uses the mainnet parameters defined here: https://github.com/zcash/librustzcash/blob/main/zcash_primitives/src/constants/mainnet.rs, with its only difference being all the network upgrade heights set to 1. 
the regtest parameters are different, defined here: https://github.com/zcash/librustzcash/blob/main/zcash_primitives/src/constants/regtest.rs. For minimum viable setup we can set the network upgrade heights to 1, but eventually in theory we should be parsing the `zcash.conf` and parsing the network upgrade heights defined in there.

AloeareV:
If we could get the 'press a button to boot a zcashd a, lightwalletd, and a zingo-cli that all connect to each other' for mainnet and/or testnet, that would be fantastic to get in, I don't want that functionality to be blocked on regtest.

AloeareV:
The network enum is inspired by a similar enum in librustzcash defined here: https://github.com/zcash/librustzcash/blob/0142a3db1b16d3859a013a7079587ae7c5882afc/zcash_primitives/src/consensus.rs#L274-L278

dannasessha:
Acknowledged, I think I understand, I'll look more  into it.

dannasessha:
I'd like a review but I think I have got Network:Regtest working, tested locally.

README is attempting to be very verbose with the state of this branch also.

AloeareV (approved):
This looks good! Fixes #12. Will spawn a new issue of our regtest mode expecting specific NU heights, but I don't think being able to be flexible for custom network upgrade heights is high-priority. 

Zancas was specifically requested as a reviewer, which I interpret as a desire for zancas to see this before its merged in. I won't hit merge, and will wait and allow @zancas to merge this, or @dannasessha if you wish to merge before that happens!

skyl (approved):
LGTM

skyl:
> The network enum is inspired by a similar enum in librustzcash defined here: https://github.com/zcash/librustzcash/blob/0142a3db1b16d3859a013a7079587ae7c5882afc/zcash_primitives/src/consensus.rs#L274-L278

Maybe it makes sense to eventually contribute the regtest upstream and depend on that definition?

zancas:
This looks awesome!   It's still marked `WIP`, and labeled `DRAFT`.   I am +1 on merging this, if that's correct then that mark and label should change.

zancas (approved):
LGTM!  I can't remove `DRAFT` or `WIP` but this seems like a landable chunk of work, so I am merging!

PR #5, opened 2022-06-24
update README

dannasessha:
Tested instructions of "Pre-requisites", using docker, with:
`Debian 11.3`
`Debian 10.12`
`Ubuntu 22.04`
`Ubuntu 20.04`
`Ubuntu 18.04`

I included Debian because `zcash`'s official build support includes Debian, so I'd like to see at _least_ a one distro supported setup for this repo, `lightwalletd` and `zcash`.

PR #22, opened 2022-07-01
update readme 2

### LINKED ISSUE #18: Change suggested use of, or understand and clear specifics of `0.0.0.0`
Currently, the README specifies adding `rpcallowip=0.0.0.0` to `zcash.conf` to work in "regtest mode."

`0.0.0.0` can mean different things in different contexts, IIUC, and in some cases it means "any address" ie, any connection is not ruled out. I'm honestly not sure what it means exactly given this context, but it's a question we need to either know well the answer to, and understand it poses no problem in itself, or solve by suggesting something different that also works, before a general launch.

dannasessha:
This is working for me locally. `127.0.0.1` should be less permissive about what it is allowed to connect with, or at the very least it is more clearly limited to local connections.

fixes #18 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,13 +28,10 @@ Zingo-CLI does automatic note and utxo management, which means it doesn't allow 
 
 #### Pre-requisites
 * Rust v1.37 or higher.
-    * Run `rustup update` to get the latest version of Rust if you already have it installed
 * Rustfmt
     * Run `rustup component add rustfmt` to add rustfmt
 * Build tools
     * Please install the build tools for your platform. On Ubuntu `sudo apt install build-essential gcc`
-* Protobuf Compiler
-    * Please install the protobuf compiler for your platform. On Ubuntu `sudo apt install protobuf-compiler`
 
 ```
 git clone https://github.com/zingolabs/zingolib.git
@@ -46,7 +43,7 @@ cargo build --release
 This will launch the interactive prompt. Type `help` to get a list of commands.
 
 ## Notes:
-* If you want to run your own server, please see [zingo lightwalletd](https://github.com/zingolabs/lightwalletd), and then run `./zingo-cli --server http://127.0.0.1:9067`
+* If you want to run your own server, please see [zcash lightwalletd](https://github.com/zcash/lightwalletd), and then run `./zingo-cli --server http://127.0.0.1:9067`
 * The log file is in `~/.zcash/zingo-wallet.debug.log`. Wallet is stored in `~/.zcash/zingo-wallet.dat`
 * Currently, the default, hard-coded `lightwalletd` server is https://lwdv3.zecwallet.co:443/. To change this, you can modify line 25 of `lib/src/lightclient/lightclient_config.rs`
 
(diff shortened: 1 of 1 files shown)


dannasessha:
Removing rust install instruictions following up on [this](https://github.com/zingolabs/zingolib/pull/5/files#r909064686).
Removing protobuf as I didn't need this in building in my test containers, as noted [here](https://github.com/zingolabs/zingolib/pull/5#issue-1284313480).
Changing instructions to look at `zcash/lightwalletd` based off my own test pipeline and internal chats.

dannasessha:
If protobuf is needed for some users to get this pipeline to work, let's open an issue for that and get specific.

zancas:
I am curious about evidence that `protobuf-compiler` is not required.

The `build.rs` script is in control of whether or not protobuf `.proto` "spec" files are compiled into Rust native types on any given build.

On this line:

https://github.com/zingolabs/zingolib/blob/ddf8261a2a6aabb5a1fb8b46d450d7ed5aee339c/lib/build.rs#L6

there's a conditional directive (`println!` is part of the build.rs API!), that says:

Only compile from `.proto` files if this file has changed, where "this" file is:

`proto/service.proto`

in this case.

Is it possible that you discovered that the compiler wasn't necessary in situations where it had previously run, and that file hadn't changed?

zancas:
My guess is that if you tweak that file, without a `protobuf-compiler` present the build will fail.

A pretty effective way to ensure a clean state, such that artifacts of previous builds aren't contaminating your experiment might be something like:

`git worktree add ../new_build_expt && cd ../new_build_expt && cargo build`

My `worktree` fu is fairly weak, so I may be missing important branch manipulating steps.. for more sophisticated use cases... but to experiment with building (without necessarily changing branches, I am sure that'll give you a clean working directory).

dannasessha:
> My guess is that if you tweak that file, without a `protobuf-compiler` present the build will fail.

No, that's what I was saying with [the link above](https://github.com/zingolabs/zingolib/pull/5#issue-1284313480). I ran a series of tests in docker, and was not using that in the build script at all. Also it's not present in the repo's `Dockerfile` right now. 

I could be wrong about this, I didn't _re_ confirm it works without that instruction, but this change was added in when a windows user was having an issue, IIRC. I don't think I ever got a clear answer if it was a mistaken instruction, but I thought I possibly saw something to that effect.

What would be a good way to put this to the test, other than removing that instruction and then showing it can build (which I believe I did, as noted in that link above)?

> A pretty effective way to ensure a clean state, such that artifacts of previous builds aren't contaminating your experiment might be something like:
> 
> `git worktree add ../new_build_expt && cd ../new_build_expt && cargo build`
> 
> My `worktree` fu is fairly weak, so I may be missing important branch manipulating steps.. for more sophisticated use cases... but to experiment with building (without necessarily changing branches, I am sure that'll give you a clean working directory).

I was using containers in my trials, whereas I think this kind of instruction on my local machine would include whatever I've already installed (for example I have a package `protobuf` installed.) 

I could be wrong, but I think it's good to explore this issue. Is there a reason why using `docker` wouldn't be a good test of a 'clean' build, regardless of other tools, repos, and so on?

PR #46, opened 2022-07-15
fix warnings

zancas:
We need a low-noise environment to develop in.  This commit simply addresses each compiler warning.


dannasessha (approved):
Looks good!

I'd like to see a follow-up issue addressing the changes so we can look at these places in the code to remove variable names, remove unused code, whatever might be appropriate.

PR #52, opened 2022-07-21
Patch zingolabs lrzc+orchard, finish scan_bundle generic impl

AloeareV:


lib/src/wallet/traits.rs
```diff
@@ -0,0 +1,464 @@
...
+    },
+    zip32::{
+        ExtendedFullViewingKey as SaplingExtendedFullViewingKey,
+        ExtendedSpendingKey as SaplingExtendedSpendingKey,
+    },
+};
+use zingoconfig::Network;
+
+pub(crate) trait ToBytes<const N: usize> {
+    fn to_bytes(&self) -> [u8; N];
```
zancas, line 48: Is this for the purpose of providing a standard interface?

lib/src/wallet/traits.rs
```diff
@@ -0,0 +1,464 @@
...
+    }
+}
+
+impl ToBytes<512> for Memo {
+    fn to_bytes(&self) -> [u8; 512] {
+        *self.encode().as_array()
+    }
+}
+
+impl ToBytes<512> for MemoBytes {
```
zancas, line 69: Can we factor out the `MemoBytes` type?

lib/src/wallet/traits.rs
```diff
@@ -0,0 +1,464 @@
...
+impl<const N: usize> ToBytes<N> for [u8; N] {
+    fn to_bytes(&self) -> [u8; N] {
+        *self
+    }
+}
+
+pub(crate) trait ShieldedOutputExt<D: Domain>:
+    ShieldedOutput<D, ENC_CIPHERTEXT_SIZE>
+{
+    fn cv(&self) -> D::ValueCommitment;
```
zancas, line 84: Are we forced to use this cryptic `cv` for back compatibility?

lib/src/wallet/traits.rs
```diff
@@ -0,0 +1,464 @@
...
+    }
+}
+
+impl ToBytes<512> for MemoBytes {
+    fn to_bytes(&self) -> [u8; 512] {
+        *self.as_array()
+    }
+}
+
+impl<const N: usize> ToBytes<N> for [u8; N] {
```
zancas, line 75: Huh....

lib/src/wallet/traits.rs
```diff
@@ -0,0 +1,464 @@
...
+    fn to_bytes(&self) -> [u8; N] {
+        *self
+    }
+}
+
+pub(crate) trait ShieldedOutputExt<D: Domain>:
+    ShieldedOutput<D, ENC_CIPHERTEXT_SIZE>
```
zancas, line 82: Are we sure that we don't want `COMPACT_NOTE_SIZE` here?

lib/src/wallet/traits.rs
AloeareV: Yes.

lib/src/wallet/traits.rs
AloeareV: No. Will improve this name.

lib/src/wallet/traits.rs
AloeareV: We only need this for a certain value, but there's no reason not to implement it for any N.

lib/src/wallet/traits.rs
AloeareV: `usize` is the right kind of number, as we implement it on an array, which is generic for a `usize`, and I _think_ the type of the const generic needs to match.

lib/src/wallet/traits.rs
AloeareV: Yes, we're implementing on `Action` and `OutputDescription`. This is used after we fetch full (raw) transactions. There are different types `CompactAction` and `CompactOutputDescription` that will implement this for `COMPACT_NOTE_SIZE`

lib/src/wallet/traits.rs
zancas: I understand this constraint...  testing...

lib/src/wallet/traits.rs
AloeareV: Done.

lib/src/wallet/traits.rs
```diff
@@ -0,0 +1,555 @@
...
+    fn to_bytes(&self) -> [u8; N] {
+        *self
+    }
+}
+
+pub(crate) trait ShieldedOutputExt<P: Parameters, D: Domain>:
+    ShieldedOutput<D, ENC_CIPHERTEXT_SIZE>
+{
+    fn domain(&self, height: BlockHeight, parameters: P) -> D;
+    fn out_ciphertext(&self) -> [u8; 80];
```
zancas, line 88: Why doesn't `zcash_note_encryption::ShieldedOutput` provide interfaces to these "encrypted transaction output" components?

zancas: Where by "these" I mean:

* out_ciphertext
* value_commitment

AloeareV: No reason I'm aware of. Note that adding required methods to a pub trait is a breaking change.

zancas: Issued here: https://github.com/zingolabs/zingolib/issues/63

lib/src/wallet/traits.rs
```diff
@@ -0,0 +1,590 @@
...
+    fn unspent_from_wallet_txns(transactions: &WalletTxns) -> Vec<(Self, u64, TxId)> {
+        transactions.get_unspent_orchard_nullifiers()
+    }
+}
+
+pub(crate) trait NoteData: Sized {
+    type Fvk: Clone;
+    type Diversifier;
+    type Note: PartialEq;
+    type Node: Hashable;
```
zancas, line 299: This language `Node` to describe something that has a particular coordinate in a data-structure, is from the perspective of someone who's thinking of the IncrementalMerkle note commitment tree.

I am curious about the constraints on this name, and whether a more conceptually powerful might be available.

lib/src/wallet/traits.rs
```diff
@@ -0,0 +1,590 @@
...
+}
+
+pub(crate) trait DomainWalletExt<P: Parameters>: Domain
+where
+    Self: Sized,
+    Self::Note: PartialEq,
+    Self::Recipient: Recipient,
+    Self::Note: PartialEq,
+{
+    type Fvk: Clone;
```
zancas, line 546: Given that `zcash_note_encryption::Domain` has IVK and OVK associated types..  is it really necessary to also have the FVK?

lib/src/wallet/traits.rs
```diff
@@ -0,0 +1,590 @@
...
+    >;
+    type Key: WalletKey<
+            Ovk = <Self as Domain>::OutgoingViewingKey,
+            Ivk = <Self as Domain>::IncomingViewingKey,
+            Fvk = <Self as DomainWalletExt<P>>::Fvk,
+        > + Clone;
+
+    type Bundle: Bundle<Self, P>;
+
+    fn wallet_notes_mut(_: &mut WalletTx) -> &mut Vec<Self::WalletNote>;
```
zancas, line 561: It seems like this is set up for later use, since there's no references to it, outside of the `traits.rs` mod?

lib/src/wallet/traits.rs
```diff
@@ -0,0 +1,590 @@
...
+}
+
+impl FromBytes<32> for OrchardNullifier {
+    fn from_bytes(bytes: [u8; 32]) -> Self {
+        Option::from(OrchardNullifier::from_bytes(&bytes))
+            .expect(&format!("Invalid nullifier {:?}", bytes))
+    }
+}
+
+pub(crate) trait FromCommitment {
```
zancas, line 143: I assume this is also...   "for future use" since there are no references?

lib/src/wallet/traits.rs
```diff
@@ -0,0 +1,590 @@
...
+    fn wallet_transaction_notes_mut(wallet_transaction: &mut WalletTx) -> &mut Vec<Self> {
+        &mut wallet_transaction.orchard_notes
+    }
+
+    fn witnesses(&mut self) -> &mut WitnessCache<Self::Node> {
+        &mut self.witnesses
+    }
+}
+
+pub(crate) trait DomainWalletExt<P: Parameters>: Domain
```
zancas, line 539: Would the name `WalletDomainExt` make more sense?

lib/src/wallet/traits.rs
```diff
@@ -0,0 +1,590 @@
...
+    }
+    fn wallet_nullifier(null: &Self::Nullifier) -> WalletNullifier {
+        WalletNullifier::Orchard(*null)
+    }
+}
+
+pub(crate) trait Recipient {
+    type Diversifier;
+    fn diversifier(&self) -> Self::Diversifier;
+    fn encode(&self, chain: Network) -> String;
```
zancas, line 187: How constrained are we by this name?    I think a new type wrapper might be ideal for the returned value, and the method name might indicate how/why we provide this representation?

lib/src/blaze/fetch_full_transaction.rs
```diff
@@ -577,62 +464,163 @@ impl FetchFullTxns {
...
+            Item = &'a <<D as DomainWalletExt<Network>>::Bundle as zingo_traits::Bundle<
+                D,
+                Network,
+            >>::Output,
+        >,
+    D::Memo: zingo_traits::ToBytes<512>,
 {
-    let local_keys = key_reader(&*keys.read().await).clone();
-    for output in transaction_to_bundle(transaction)
+    // Check if any of the nullifiers spent in this transaction are ours. We only need this for unconfirmed transactions,
```
zancas, line 503: Does this language match our practice?   Do we say "nullifiers spent"? 

Or rather do we say that:

 `A spend of an output generates a nullifier, and a new output.`

?

lib/src/blaze/fetch_full_transaction.rs
```diff
@@ -577,62 +464,163 @@ impl FetchFullTxns {
...
+    // because for transactions in the block, we will check the nullifiers from the blockdata
+    if unconfirmed {
+        let unspent_nullifiers =
```
zancas, line 506: These are nullifiers that are not yet in blocks?

lib/src/wallet/transactions.rs
```diff
@@ -523,7 +528,35 @@ impl WalletTxns {
...
+                    .is_none()
+                {
+                    wtx.add_spent_nullifier(WalletNullifier::Orchard(nullifier), value)
+                }
+
+                // Since this Txid has spent some funds, output notes in this Tx that are sent to us are actually change.
+                self.check_notes_mark_change(&txid);
+
+                // Mark the source note's nullifier as spent
+                if !unconfirmed {
```
zancas, line 545: This definition of `confirmed` means that the `nullifier` is in a block?  And therefore can be marked as `spent`?

lib/src/wallet.rs
```diff
@@ -567,19 +569,19 @@ impl LightWallet {
 
     async fn add_imported_spend_key<
         WKey: WalletKey + Clone,
```
zancas, line 571: Is `WKey` a useful abbreviation?   Should we just use `WalletKey` as the identifier?

lib/src/blaze/fetch_full_transaction.rs
zancas, line 503: Any reason to _not_ factor lines 503-532 into a helper?

lib/src/wallet/traits.rs
AloeareV: I matched the name of the generic type in `librustzcash`, see `zcash_primitives::merkle_tree::IncrementalWitness`. A more conceptually powerful name may well be available. Note that the sapling concrete type is named `Node`, and the orchard concrete type is named `MerkleHashOrchard`

lib/src/wallet/traits.rs
AloeareV: Line 681 of `src/wallet/transactions.rs`

lib/src/blaze/fetch_full_transaction.rs
```diff
@@ -577,62 +464,163 @@ impl FetchFullTxns {
...
-async fn scan_bundle<K, B, L, O, D, E>(
-    _config: &ZingoConfig,
+async fn scan_bundle<D>(
+    config: &ZingoConfig,
     transaction: &Transaction,
-    _height: BlockHeight,
-    _unconfirmed: bool,
-    _block_time: u32,
+    height: BlockHeight,
+    unconfirmed: bool,
```
zancas, line 476: We don't want this `unconfirmed` to be decouplable from this transaction.  Maybe we need a:
`TransactionAndMetadata` type-or-trait?

lib/src/blaze/fetch_full_transaction.rs
AloeareV: Nope! Issued!

lib/src/blaze/fetch_full_transaction.rs
AloeareV: Issued.

lib/src/blaze/fetch_full_transaction.rs
AloeareV: Yes. Whether or not they can be in the mempool is unclear.

lib/src/blaze/fetch_full_transaction.rs
AloeareV: I think `nullifiers spent` is concise, and it would be nice to be more concise than having to say 'Check if any of the nullifiers in this transaction spend notes belonging to us', but 'nullifiers spent' doesn't convey that very well.

lib/src/wallet/traits.rs
AloeareV: We call functions that return strings, and usually we pass them to a user with no further transformation, where the only reason to turn them into strings is to write them down somewhere.. I think `String` is a fine type, and we've improved the name.

lib/src/blaze/fetch_full_transaction.rs
AloeareV: Issued this nullifier spending business.

zancas (approved):
All my conversations/commments have been addressed.

Issue #36, opened 2022-07-08
Regtest 2.2 - Sleepless Launch


dannasessha:
The second half of #34 ; sleepless launch of regtest mode.

I plan to tackle this after getting a graceful shutdown of the daemons implemented.

Issue #27, opened 2022-07-04
More informative messages when cli fails to launch.


dannasessha:
While working on a feature, I saw
```
Error during startup:Error with response: Status { code: Unavailable, message: "error trying to connect: tcp connect error: Connection refused (os error 111)", source: Some(hyper::Error(Connect, ConnectError("tcp connect error", Os { code: 111, kind: ConnectionRefused, message: "Connection refused" }))) }
If you repeatedly run into this issue, you might have to restore your wallet from your seed phrase.
```

This failure had nothing at all to do with the seed phrase, and a naive user wouldn't know this.

We should look at what failure messages are printed under what conditions, and probably improve them.

zancas:
One cause of that message is missing network parameters.

Fixed by running `zcutil/fetch_params.sh`

Another confusing message is the message printed on regtest launch, that is caused by the 0-block blockchain.

zancas:
@dannasessha are you available to test this to see if the error messages have improved?

If I don't hear back from you within 48 hours, I am going to close this as won't-fix.

PR #56, opened 2022-07-23
checks for ready zcashd and lwd daemons in regtest mode

dannasessha:
Chasing #34 and #36 - probably a WIP. This no longer has a hard coded waiting duration for the daemons to start, instead there is a check every half second against the log files. I considered several more exotic solutions, for example, splitting the stdout into multiple outputs like `tee` in the shell, and I would like to move toward an async solution, but this is working for me.

I also added a simple error check for `zcashd`, which only works in regtest mode, and logging for `lightwalletd`'s `stdout`into a file.

Implementing async could or will be a further refinement, with a timeout of the attempts and so on. I'd also like to have more consideration of error detection, including if that should be ported to other "modes" like mainnet usage.

See also #27 .

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1,5 +1,6 @@
 mod regtest;
 use log::error;
+use tokio::runtime::Runtime;
 use zingo_cli::{
     attempt_recover_seed, configure_clapapp, report_permission_error, start_interactive, startup,
     version::VERSION,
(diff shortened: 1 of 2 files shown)


zancas:
Circling back to this.   It's still marked draft.  Should I build and manually test?  Is review requested?

dannasessha:
This may be a case of the intermediate state should be merged, pending and 'upgrade' to async logic, which is still a WIP. I am going to dig on this issue to get specifics right now.

AloeareV:
Closing in favor of #70

PR #69, opened 2022-08-04
Basic regtest mode elegant startup

dannasessha:
Part of the incomplete #56 PR - Still aiming for #34 and #36 .

These are not completed features (because we want to handle this with async for a variety of good reasons), but should be working for users. Therefore I advocate this should be merged.


dannasessha:
If this is merged, #56 should be closed. I think the associated issues should remain open, as those can be closed when the feature is completed, with async logic.

AloeareV:
Closed in favor of #70 (Which I think includes all of this)

Issue #71, opened 2022-08-05
Regtest mode generated blocks don't show up


AloeareV:
When I run in regtest mode, I get 
```
zcashd is starting in regtest mode, please standby...
zcashd start section completed, zcashd should be running.
lightwalletd is about to start. This should only take a moment.
lightwalletd is now started in regtest mode, please standby...
lwd start section completed, lightwalletd should be running!
Standby, Zingo-cli should be running in regtest mode momentarily...
regtest detected and network set correctly!
Creating a new wallet
Lightclient connecting to http://127.0.0.1:9067/
Error with response: Status { code: Unknown, message: "Cache is empty. Server is probably not yet ready", metadata: MetadataMap { headers: {"content-type": "application/grpc"} }, source: None }
Ready!
```
After using `zcash-cli` to generate blocks, those blocks never show up on the zingo-cli, even after running `rescan`.

zancas:
This is a significant UnknownUnknown.    Why does this occur?

The first thing to do, is to monitor the local block cache, specific to this regtest mode.

I believe it's here:

`regtest/data/zcashd/regtest/`

While `watch 'ls -la'` that directory, launch the _same_ `zcashd`,

(located here: `regtest/bin/zcashd`) (in regtest mode with the appropriate "datadir")

Attach `regtest/bin/zcash-cli` to _THAT_ zcashd, and report the output of:

  1. `zcash-cli getinfo`
  2. `zcash-cli getblockchaininfo`
  3. `zcash-cli version`

in this comment thread.

Can we generate _THIS_ error "directly" with `zcash-cli`:

`Error with response: Status { code: Unknown, message: "cache is empty, server is probably not yet ready", metadata: MetadataMap { headers: {"content-type": "application/grpc"} }, source: None }`

NO..  because @AloeareV tells me that it's a `lightwalletd` error.  ;)

Now with this context:

  0. use `tail -F regtest/logs/lightwalletd/stdout.log`
       * spend some time studying this log output (which is probably old-or-empty prior to step 1)
  1. launch `zingo-cli --regtest`   Use explicit flags `--data-dir` and `--server`directories to guarantee that the same data is being referenced, as the ones we set up above.
       * notice that the `tail` of the lightwalletd log receiving data!
   2. observe that we get the `lightwalletd` error (or don't)!!! in the `zingo-cli` output
   3a. invoke `zcash-cli generate 1` and watch the `lightwalletd` `tail` to see what happens
[...]

PR #70, opened 2022-08-05
Basic regtest mode elegant startup

zancas:

PR #74, opened 2022-08-15
patch data vs datadir, cleanup

dannasessha:
Follow-up to #70 merge-in.

#71 issue is appearing on this branch as well.

### DIFF
--- a/cli/src/regtest.rs
+++ b/cli/src/regtest.rs
@@ -1,7 +1,7 @@
 use std::process::Command;
 ///  Enforce strict expectations for tool use with current zingolib.  Relaxing these restrictions will facilitate
 ///  use in other projects.  For example, this version of regtest will only run within a git repo that is historically
-///  descended from 27e5eedc6b35759f463d43ea341ce66714aa9e01.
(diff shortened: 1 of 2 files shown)


zancas:
This looks pretty good.  Can you please add two tests that:

1.  runs in CI
2.  attempts to run zcashd with the flags supplied

Test One:
   * fails as long as any flag that zcashd does _NOT_ recognize is passed

Test Two:
   * fails unless all flags necessary for `zingolib regtest mode` are present


Test One will fail until `data` is changed to `datadir`.

cli/src/regtest.rs
```diff
@@ -79,6 +79,7 @@ pub(crate) fn launch() {
     let lightwalletd_config = confs_dir.join("lightwalletd.yaml");
     let lightwalletd_logs = logs.join("lightwalletd");
     let lightwalletd_stdout_log = lightwalletd_logs.join("stdout.log");
+    let lightwalletd_stderr_log = lightwalletd_logs.join("stderr.log");
```
zancas, line 82: Nice.

zancas:
I am not aware of any work on this branch, and external factors have elevated its priority so I am going to add the requested tests, on top of these commits on my own feature, and see if I can get it merged.

PR #75, opened 2022-08-18
Regtest patch

zancas:


cli/src/regtest.rs
```diff
@@ -1,9 +1,9 @@
-use std::process::Command;
```
AloeareV, line 1: Why remove this?

cli/src/regtest.rs
```diff
@@ -59,10 +59,45 @@ fn get_regtest_dir() -> PathBuf {
...
+            .expect("failed to start zcashd"),
+        File::create(&zcashd_stdout_log).expect("file::create Result error"),
+        zcashd_stdout_log,
+    )
+}
+
 pub(crate) fn launch() {
-    use std::fs::File;
     use std::io::Read;
-    use std::process::Stdio;
```
AloeareV, line 65: Why remove this?

cli/src/regtest.rs
```diff
@@ -82,8 +117,8 @@ pub(crate) fn launch() {
     let lightwalletd_stderr_log = lightwalletd_logs.join("stderr.log");
     let lightwalletd_datadir = data_dir.join("lightwalletd");
 
-    let mut zcashd_bin = bin_location.to_owned();
-    zcashd_bin.push("zcashd");
+    let (mut zcashd_command, mut zcashd_logfile, zcashd_stdout_log) =
```
AloeareV, line 151: Looks good!

AloeareV (approved):

cli/src/regtest.rs
zancas: Because there's a another `Command` `tokio::process::Command` and I want to preempt confusion.

zancas:
> I've approved these changes (pending addressing of my review comments), but would this make more sense to merge into #74, instead of being a separate PR that builds atop it?

No, this is a superset of #74 merging this should close that.

AloeareV:
> > I've approved these changes (pending addressing of my review comments), but would this make more sense to merge into #74, instead of being a separate PR that builds atop it?
> 
> No, this is a superset of #74 merging this should close that.

Yes, historically with your orchard prs (which are almost always a superset of mine) you've opened a PR into my branch. I'm wondering why we're doing it differently here.

zancas:
Because I don't want to block on @dannasessha .    If I open a PR against @dannasessha 's branch then they would have to merge it, then that feature would have to merge subsequently into `dev`.

We have an outstanding request to use this feature.  Should we really engage a process that requires the otherwise unnecessary intervention of another developer?   What's the benefit?

zancas:
In the case where you and I are collaborating in review, it makes sense to open a PR against your branch because the subsequent merge into `dev` is something we need to consense on anyway.

AloeareV:
> Because I don't want to block on @dannasessha . If I open a PR against @dannasessha 's branch then they would have to merge it, then that feature would have to merge subsequently into `dev`.
> 
> We have an outstanding request to use this feature. Should we really engage a process that requires the otherwise unnecessary intervention of another developer? What's the benefit?

This is a convincing argument, and I'm having trouble articulating where my hesitation is coming from. 

Merging.

PR #77, opened 2022-08-26
Regtest Mode Fix

### LINKED ISSUE #71: Regtest mode generated blocks don't show up

dannasessha:
Fixes #71 - by adding pre-generated data directory contents. That is to say, there is, included in this PR, the state of a single generated block.

Regtest Mode had some issues, which involved an inability to cope with an initial state without any blocks. Additionally, any blocks added while `zcashd` is running are not recorded / retained upon subsequent runs. Therefore a direct solution (and because we do not want Regtest Mode to 'remember' state between runs) is simply to add an initial block.

Network parameters are all set to activate at block 1, and so all network upgrades should neccessarily be enacted when using Regtest Mode when using this branch.


dannasessha:
A problem with this setup is that as soon as Regtest Mode is run, the state of the (now tracked) datadirectories changes, even when not adding a block. Therefore the `git` staging becomes immediately untidy.

There may be a clever solution using `git` related tools or files, but for now, it's a bit messy, but seems to work.

regtest/data/zcashd/regtest/blocks/index/LOG
```diff
@@ -0,0 +1,5 @@
+2022/08/26-00:13:45.135344 7f00d4cece80 Recovering log #6
```
AloeareV, line 1: I suspect we want to `.gitignore` the LOG files, and definitely the LOCK/.lock files.

AloeareV: I expected them to be a problem, but they don't seem to be. It's working locally.

AloeareV (approved):
I don't know about checking in the lock and log files, but they don't seem to break anything in my local run.

PR #86, opened 2022-09-02
Fix regtest reindexing

AloeareV:
We now need a way to have zcashd's wallet (or at least its addresses) persist through a zcashd reboot, so that we can mine to them.


AloeareV:
I had a 
```
Error with response: Status { code: Unknown, message: "Cache is empty. Server is probably not yet ready", metadata: MetadataMap { headers: {"content-type": "application/grpc"} }, source: None }
``` 
pop up again, but in my attempt to debug it I failed to reproduce it, and it started working.

zancas (approved):

AloeareV:
> I had a
> 
> ```
> Error with response: Status { code: Unknown, message: "Cache is empty. Server is probably not yet ready", metadata: MetadataMap { headers: {"content-type": "application/grpc"} }, source: None }
> ```
> 
> pop up again, but in my attempt to debug it I failed to reproduce it, and it started working.

Fixed, by not using -9 when killing child processes.

zancas (approved):

zancas:
Fixes: #80

PR #87, opened 2022-09-02
Define transaction context

zancas:


AloeareV (approved):
Looks like a nice piece of refactor! It would be good to get similar stuff going with the other blaze components, but that would be a very big PR to do it all at once!

PR #88, opened 2022-09-07
Arg macro to fn

zancas:
Simple port from macro to fn, and addition of first example scenario.

You can run examples like this:

`cargo test --example scenarios`


AloeareV (approved):
Great! Down one unnecessary macro!

PR #93, opened 2022-09-07
Move key, address management into UnifiedSpendCapability

AloeareV:


cli/tests/integration_tests.rs
```diff
@@ -112,8 +112,7 @@ fn note_selection_order() {
     Runtime::new().unwrap().block_on(async {
         utils::increase_height_and_sync_client(&regtest_manager, &client_1, 5).await;
 
-        client_2.set_server(client_1.get_server().clone());
-        let address_of_2 = client_2.do_address().await["sapling_addresses"][0].clone();
+        let address_of_2 = client_2.do_address().await[0]["recievers"]["sapling"].clone();
```
zancas, line 115: `receivers`

zancas:
Zingo! depends on a built-in Orchard address.  This ensures that there's always a seed present.  We need a test that demonstrates this behavior for:
* generation from seed
* new-wallet generation
and prevents regressions.

lib/src/wallet/keys/unified.rs
```diff
@@ -0,0 +1,406 @@
...
+                .as_ref()
+                // This is deprecated. Not sure what the alternative is,
+                // other than implementing it ourselves.
+                .map(zcash_primitives::legacy::keys::pubkey_to_address),
+        )
+        .ok_or(format!(
+            "Invalid receivers requested! At least one of sapling or orchard required"
+        ))?;
+        self.addresses.push(ua.clone());
+        Ok(ua)
```
zancas, line 152: Instead of mutating the USA in place, and also returning the same UA, why not enforce a single interface on the USA that provides the UA to interested callers?

lib/src/lightclient.rs
```diff
@@ -1173,156 +1021,23 @@ impl LightClient {
...
-                let e = addr;
-                error!("{}", e);
-                return Err(e);
-            }
-
-            addr
-        };
+        let new_address = self
+            .wallet
+            .unified_spend_auth()
```
zancas, line 1033: A `unified_spend_auth` is created in this fluent expression.   Shouldn't it be in charge exposing `new_addresses` via another call?

We definitely have the case of wanting to generate new addresses, and then not use them right away, why not use that same method of exposing previously created addresses, that's done in those cases?   Instead of returning the new UA here?

lib/src/wallet/keys/unified.rs
```diff
@@ -0,0 +1,406 @@
...
+    next_sapling_diversifier_index: DiversifierIndex,
+
+    // Note that unified spend authority encryption is not yet implemented,
+    // These are placeholder fields, and are currenly always false
+    pub(crate) encrypted: bool,
+    pub(crate) unlocked: bool,
+}
+
+#[derive(Debug, Clone, Copy)]
+#[non_exhaustive]
```
zancas, line 38: According to [the release notes](https://blog.rust-lang.org/2019/12/19/Rust-1.40.0.html#non_exhaustive-structs-enums-and-variants) the rationale for the [`#[non_exhaustive] attribute`](https://doc.rust-lang.org/reference/attributes.html)

lib/src/wallet/keys/unified.rs
```diff
@@ -0,0 +1,406 @@
...
+        reader.read_exact(&mut orchard_key_bytes)?;
+        let orchard_key: orchard::keys::SpendingKey =
+            Option::from(orchard::keys::SpendingKey::from_bytes(orchard_key_bytes)).ok_or(
+                io::Error::new(io::ErrorKind::InvalidData, "Invalid orchard spending key"),
+            )?;
+        let sapling_key = zcash_primitives::zip32::ExtendedSpendingKey::read(&mut reader)?;
+        let transparent_parent_key =
+            super::extended_transparent::ExtendedPrivKey::read(&mut reader, ())?;
+        let receivers_selected =
+            Vector::read(&mut reader, |mut r| ReceiverSelection::read(&mut r, ()))?;
```
zancas, line 269: I don't understand why a `UnifiedSpendAuthority` would need more than one `ReceiverSelection`.

zancas: Is this to support more than one seed_or_key from which the wallet has derived addresses?

lib/src/wallet/keys/unified.rs
```diff
@@ -0,0 +1,406 @@
...
+use zingoconfig::ZingoConfig;
+
+use crate::wallet::traits::ReadableWriteable;
+
+use super::{extended_transparent::KeyIndex, get_zaddr_from_bip39seed, ToBase58Check};
+
+#[derive(Clone, Debug)]
+pub struct UnifiedSpendAuthority {
+    orchard_key: orchard::keys::SpendingKey,
+    sapling_key: zcash_primitives::zip32::ExtendedSpendingKey,
```
zancas, line 22: Why aren't these fields `Option<..>`?

lib/src/wallet/keys/unified.rs
AloeareV: We could. I thought it would be convenient to return the address of the newly generated UA.

lib/src/wallet/keys/unified.rs
AloeareV: Yep! It's a drop in the bucket when it comes to the amount of breaking changes a new pool would likely entail for our public interface, but this means consumers of a `ReceiverSelection` would need to have a catch-all match for future receiver types.

lib/src/wallet/keys/unified.rs
AloeareV: This is to support diversified addresses.

lib/src/wallet/keys/unified.rs
AloeareV: Because by default we generate a single UA with all three pools... So we need all three keys. We don't currently have a use-case for _not_ including all key types, and the logic is simpler when we can rely on their existence.

lib/src/lightclient.rs
AloeareV: Returning the new UA here is intended purely for convenience. I don't mind removing it.

lib/src/wallet/keys/unified.rs
zancas: It _is_ convenient, but there's a clarity cost.   Your call.

lib/src/blaze/trial_decryptions.rs
```diff
@@ -289,10 +289,11 @@ impl TrialDecryptions {
...
 
                 workers.push(tokio::spawn(async move {
-                    let keys = keys.read().await;
-                    let wallet_key = &D::Key::get_keys(&*keys)[ivk_num];
-                    let fvk = wallet_key.fvk().unwrap();
-                    let have_spending_key = wallet_key.spend_key().is_some();
+                    let usa = usa.read().await;
+                    let fvk = D::Key::usc_to_fvk(&*usa);
+
+                    // We don't have fvk import, all our keys are spending
```
zancas, line 295: How about we have an issue that tracks the state of viewing key imports?   We're not certain that it's worth supporting.

lib/src/wallet/keys.rs
```diff
@@ -113,899 +85,47 @@ impl FromBase58Check for str {
...
-    // viewing keys and imported spending keys.
-    pub(crate) zkeys: Vec<SaplingKey>,
-
-    // Transparent keys. If the wallet is locked, then the secret keys will be encrypted,
-    // but the addresses will be present. This Vec contains both wallet and imported tkeys
-    pub(crate) tkeys: Vec<TransparentKey>,
-
-    // Orchard keys
-    pub(crate) okeys: Vec<OrchardKey>,
+pub fn get_zaddr_from_bip39seed(
```
zancas, line 88: Add unittest

lib/src/wallet/keys/transparent.rs
```diff
@@ -44,23 +41,7 @@ impl TransparentKey {
...
-            )
-            .unwrap()
-            .derive_private_key(KeyIndex::hardened_from_normalize_index(0).unwrap())
-            .unwrap()
-            .derive_private_key(KeyIndex::Normal(0))
-            .unwrap()
-            .derive_private_key(KeyIndex::Normal(pos))
-            .unwrap()
-            .private_key
+        ExtendedPrivKey::get_ext_taddr_from_bip39seed(config, bip39_seed, pos).private_key
```
zancas, line 44: `pos`?   ProofOfStake?!

zancas (approved):
We need to expose this logic to the mobile platform to test in that context.

PR #98, opened 2022-09-11
Refactor cli

zancas:
This only needs one reviewer.   It's a code refactor that should make subsequent enhancement of `zingo-cli` easier.


cli/src/main.rs
```diff
@@ -15,41 +15,36 @@ pub fn main() {
...
     let maybe_birthday = matches.value_of("birthday");
-
     if seed.is_some() && maybe_birthday.is_none() {
         eprintln!("ERROR!");
         eprintln!(
             "Please specify the wallet birthday (eg. '--birthday 600000') to restore from seed."
         );
-        eprintln!("This should be the block height where the wallet was created. If you don't remember the block height, you can pass '--birthday 0' to scan from the start of the blockchain.");
-        return;
+        panic!("This should be the block height where the wallet was created. If you don't remember the block height, you can pass '--birthday 0' to scan from the start of the blockchain.");
```
AloeareV: Could we return with a non-success exit code? I'm of the opinion that a panic indicates something has gone wrong internally, we shouldn't be panicking on invalid user input.

cli/src/main.rs
```diff
@@ -15,41 +15,36 @@ pub fn main() {
...
     }
-
     let birthday = match maybe_birthday.unwrap_or("0").parse::<u64>() {
         Ok(b) => b,
         Err(e) => {
-            eprintln!(
+            panic!(
```
AloeareV: As above.

AloeareV (changes requested):
Other than the change to panic, this looks good!

cli/src/main.rs
AloeareV: I believe returning `Err` from `main` accomplishes this.

PR #99, opened 2022-09-11
Short circuit help

zancas:
Only one reviewer is needed.    `help` and `help $arg` now run without launching unnecessary infrastructure.

This short-circuit pattern should be easy to extend to other commands that should be short-circuited.


lib/src/commands.rs
```diff
@@ -287,11 +287,11 @@ impl ShortCircuitedCommand for HelpCommand {
                 responses.sort();
                 responses.join("\n")
             }
-            1 => match get_commands().get(args[0]) {
+            1 => match get_commands().get(&args[0]) {
                 Some(cmd) => cmd.help(),
                 None => format!("Command {} not found", args[0]),
             },
-            _ => self.help(),
+            _ => panic!("Unexpected number of parameters."),
```
AloeareV, line 294: Similar to previous PR. A panic means something unexpected has gone wrong.. We shouldn't be panicking on invalid user input. A non-zero exit code would be the ideal solution.

AloeareV (approved):
Looks good! Works locally! Merging, once CI passes!

PR #116, opened 2022-09-30
Reduce some numbers to make more sense with batch size 500_000 -> 1_000 change

AloeareV:


AloeareV:
Numbers are all still fairly arbitrary.

lib/src/blaze/block_witness_data.rs
```diff
@@ -73,7 +73,7 @@ impl BlockAndWitnessData {
             blocks_in_current_batch: Arc::new(RwLock::new(vec![])),
```
zancas, line 73: @AloeareV I propose we chage this struct name to:   `BatchProcessor`

zancas (approved):

zancas:
We don't yet support integration tests in CI.

PR #117, opened 2022-09-30
Move orchard anchor selection to not block other things

AloeareV:


lib/src/blaze/block_witness_data.rs
```diff
@@ -372,6 +372,42 @@ impl BlockAndWitnessData {
...
+    pub async fn update_orchard_anchors_process(
+        &self,
+        start_block: u64,
+        end_block: u64,
+        uri: Arc<std::sync::RwLock<Uri>>,
+    ) -> JoinHandle<Result<(), String>> {
+        let orchard_anchors = self.orchard_anchors.clone();
+        tokio::spawn(async move {
+            let mut workers = FuturesUnordered::new();
+            for block_height in end_block..=start_block {
```
zancas, line 384: Is it possible that this will end up consuming more threads of execution than is optimal?

Should we have some kind of cap on the number of workers, or is the interface to `workers.push` somehow managing this in a safe manner?

zancas: I guess I see the `workers.push` inside loops in plenty of other places...

lib/src/lightclient.rs
```diff
@@ -1715,6 +1715,25 @@ impl LightClient {
                 .await
         });
 
+        // Update anchors only on the last batch
+
+        let update_orchard_anchors_handle = {
```
zancas, line 1720: I like this.  Quite cute.

zancas (dismissed):

lib/src/blaze/sync_status.rs
```diff
@@ -74,14 +76,15 @@ impl fmt::Display for SyncStatus {
         if self.blocks_total > 0 && self.in_progress {
             write!(
                 f,
-                "id: {}, blocks: {}% ({}), decryptions: {}% ({}), tx_scan: {}% ({})",
+                "id: {}, blocks: {}% ({}), decryptions: {}% ({}), tx_scan: {}% ({}), anchors: {}",
```
zancas, line 79: Should we be explicit that the these are `Orchard`?

PR #125, opened 2022-10-04
Update clear_old_witnesses to clear orchard witnesses too

### LINKED ISSUE #123: clear_old_witnesses
Needs to be generalized for orchard as well as sapling

AloeareV:
Fixes #123


zancas (changes requested):
This looks great!  How about a test that asserts that there are old witnesses for both domains, then calls the domain clearing method, and asserts that the witnesses are cleared.

It fails for orchard prior to this code landing.

zancas (dismissed):

zancas (approved):

PR #130, opened 2022-10-10
Refactor test utils

zancas:
My goal with this work is to push test framework implementation details into a lower layer of abstraction, than test details.

Specifically, a test author should not need to write sleep or Duration or a polling loop in their integration test to ensure that the environment is correctly configured to test the properties of interest. That logic (e.g. syncing lightwalletd with the regtest blockchain, and syncing zingocli-client with lightwalletd, should be handled by standard reused utility functions UNLESS those things are being tested.


zancas:
My goal with this work is to push test framework implementation details into a lower layer of abstraction, than test details.

Specifically, a test author should not need to write `sleep` or `duration` or a polling loop in their integration test to ensure that the environment is correctly configured to test the properties of interest.   That logic (e.g. syncing lightwalletd with the regtest blockchain, and syncing `zingocli-client` with `lightwalletd`, should be handled by standard reused utility functions _UNLESS_ those things are being tested.

cli/tests/integration_tests.rs
```diff
@@ -110,26 +108,25 @@ fn send_mined_sapling_to_orchard() {
...
-        sleep(Duration::from_secs(2)).await;
-
-        client.do_sync(true).await.unwrap();
+        utils::increase_height_and_sync_client(&regtest_manager, &client, 2).await;
         let balance = client.do_balance().await;
         assert_eq!(balance["verified_orchard_balance"], 5000);
     });
 }
-
+/// This implements similar behavior to 'two_clients_a_coinbase_backed', but with the
```
AloeareV, line 116: Stale comment

cli/tests/utils/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+use zingo_cli::regtest::RegtestManager;
+use zingolib::lightclient::LightClient;
+
+pub async fn increase_height_and_sync_client(
+    manager: &RegtestManager,
+    client: &LightClient,
+    n: u32,
+) {
+    let start_height = client
+        .do_wallet_last_scanned_height()
```
AloeareV, line 13: If the client (or lightwalletd) isn't fully synced when this is called, then `start_height` will lead to an improperly determined target, which I think will depending on the numbers involved either early-return or hang forever.

cli/tests/integration_tests.rs
zancas: Fixed.

cli/tests/utils/mod.rs
zancas: I believe this is fixed by running `do_sync` immediately before this line.

zancas: Though..  this fix assumes that `do_sync` fulfills its contract reliably.

AloeareV (approved):
Looks great!

PR #136, opened 2022-10-15
Update Serialization Version Naming Scheme

zancas:
I assume this means that @juanky201271 does not use `Keys::read` directly from `zingo-mobile`, and that this is therefore a safe update.


AloeareV (approved):
Good clarity here!

PR #146, opened 2022-10-26
tweak to make test slightly faster without loss of insight

zancas:


AloeareV (approved):
Yeah, this should be fine.

PR #153, opened 2022-10-28
match network strings to info.chain_name

zancas:


juanky201271 (changes requested):
What about changing the name of the field too???? chain_name vs network...

AloeareV (approved):

lib/src/commands.rs
```diff
@@ -78,18 +78,18 @@ impl Command for ParseCommand {
...
-                    zingoconfig::BlockChain::Testnet,
-                    zingoconfig::BlockChain::Regtest,
+                    zingoconfig::ChainType::Mainnet,
+                    zingoconfig::ChainType::Testnet,
+                    zingoconfig::ChainType::Regtest,
                 ]
                 .iter()
                 .find_map(|chain| RecipientAddress::decode(chain, &args[0]).zip(Some(chain)))
                 .map(|(recipient_address, chain_name)| {
                     let chain_name_string = match chain_name {
```
juanky201271, line 88: I don't know much of rust, but I think here is still the field named `chain_name`

lib/src/lib.rs
```diff
@@ -46,10 +46,10 @@ pub fn create_zingoconf_with_datadir(
         let config = ZingoConfig {
             server_uri: Arc::new(RwLock::new(server)),
             chain: match info.chain_name.as_str() {
```
juanky201271, line 48: and I think here as well...

juanky201271 (changes requested):
I did a couple of comments...

lib/src/commands.rs
zancas: Well...  it's a local variable.   Do you prefer another name?  I thought we wanted to name it `chain_name`?

lib/src/lib.rs
zancas: Right...  I thought that you wanted us to match the interface of `info`?

juanky201271: yes, match the interface of `info` (chain_name) and with the new feature `parse` (network)... both with the same name and the same values....

lib/src/commands.rs
juanky201271: If this a local var... sorry. My bad...

lib/src/lib.rs
zancas: It looks like the `InfoCommand` will return an object with:

https://github.com/zancas/zingolib/blob/90996dc2862249d81af6bf3013a1bc6ce4278bdb/lib/src/lightclient.rs#L585

`"chain_name" => i.chain_name,`

Meanwhile will return an object with a key `chain_name` like this:

https://github.com/zancas/zingolib/blob/90996dc2862249d81af6bf3013a1bc6ce4278bdb/lib/src/commands.rs#L98

"chain_name" => chain_name_string,


So the interface to each should see a key "chain_name"..  right?

juanky201271: yes, sounds perfect.

PR #164, opened 2022-11-04
Android seed

AloeareV:


lib/src/blaze/trial_decryptions.rs
```diff
@@ -169,7 +169,7 @@ impl TrialDecryptions {
 
         for compact_block in compact_blocks {
             let height = BlockHeight::from_u32(compact_block.height as u32);
-            info!("trial decrypting block {}", height);
+            //info!("trial decrypting block {}", height);
```
zancas, line 172: Instead of commenting this out...   should we change its "loudness"?

lib/src/wallet.rs
```diff
@@ -210,7 +210,10 @@ impl LightWallet {
             system_rng.fill(&mut seed_bytes);
             Mnemonic::from_entropy(seed_bytes)
         } else {
-            Mnemonic::from_phrase(seed_phrase.unwrap().as_str())
+            let mnemonic = Mnemonic::from_phrase(seed_phrase.unwrap().as_str());
+
+            // This should be a no-op, but seems to be needed on android for some reason
```
zancas, line 215: Let's feature gate?

e.g.

`#[cfg(target_os = "android")]`

lib/src/blaze/trial_decryptions.rs
AloeareV: Honestly I don't think we need it.. Happy to just cut it entirely

zancas (approved):

zancas (approved):

PR #166, opened 2022-11-07
Use GATs in definition instead of where-clause in consumer

AloeareV:


zancas (approved):
Yay for stabilized GATs!

PR #168, opened 2022-11-08
I was having difficulty understanding the structure of CLI launch

zancas:
This structure was largely my own construction, so I thought it would be reasonable to factor it into a more coherent flow.

The struct that is set up as a function of the command line arguments (as parsed by clap) is now called a ConfigTemplate.  Its constructor is now named "fill".  The 'filled' config template is now passed between the operational functions of the library.

This decomposes configuration from other tasks (like launch).  I hope this is more understandable.


AloeareV (approved):
Looks easier to follow! Works locally.

PR #173, opened 2022-11-08
remove excessive logging and unused recover option

zancas:


cli/src/lib.rs
```diff
@@ -380,7 +374,7 @@ pub fn startup(
     filled_template: &ConfigTemplate,
 ) -> std::io::Result<(Sender<(String, Vec<String>)>, Receiver<String>)> {
     // Initialize logging
-    LightClient::init_logging()?;
```
AloeareV, line 383: I'd prefer to tune our logging, instead of turning it off... But that's non-blocking.

PR #174, opened 2022-11-09
update toolchain to GAT stabilization release

zancas:


juanky201271 (approved):
Well, I think it's correct.

PR #175, opened 2022-11-09
try stable per Juanky's suggestion for workflow rust toolchain

zancas:
@juanky201271 how's this look?


juanky201271 (approved):
sounds good if the syntax is correct in this context.

PR #176, opened 2022-11-09
make sync status align

zancas:
Before we can land this we need to confirm that @juanky201271 can get the runtime logs in Android Studio.

I tested that on the command line setting the RUST_LOG env variable as follows:

`RUST_LOG=zingolib=debug`

gave me the same information that I was getting via `info!` before this change.

I don't know exactly how the logs are being accessed in AS so, I don't know how/where to configure that variable.

### DIFF
--- a/lib/src/blaze/sync_status.rs
+++ b/lib/src/blaze/sync_status.rs
@@ -59,13 +59,8 @@ impl fmt::Display for BatchSyncStatus {
         if self.blocks_total > 0 && self.in_progress {
             write!(
                 f,
-                "id: {}, blocks_done/blocks_total: {:4}/{:4}, decryptions: {:4}, tx_scan: {:4}, anchors: {}",
-                self.sync_id,
-                self.blocks_done,
-                self.blocks_total,
-                self.trial_dec_done,
-                self.txn_scan_done,
-                self.orchard_anchors_done
+                "**Batch** Current: {:4} Total: {:4}\n   Blocks Loaded: {:4} TrialDecrypted: {:4}, Total: {:4}",
+                self.batch_num, self.batch_total, self.blocks_done, self.trial_dec_done, self.blocks_total,
             )
         } else {
             write!(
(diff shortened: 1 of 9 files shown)


zancas:
@juanky201271 and I looked at this a bit last night.

We confirmed that we can tune the logging sensitivity by changing the code in this function:

https://github.com/juanky201271/zingo-mobile/blob/c76093c150c79a73282d9dbaf64aabb814322658/rust/android/src/lib.rs#L15

lib/src/blaze/sync_status.rs
```diff
@@ -53,36 +52,18 @@ impl SyncStatus {
...
-    }
 }
 
 impl fmt::Display for SyncStatus {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         if self.blocks_total > 0 && self.in_progress {
             write!(
                 f,
-                "id: {}, blocks: {}% ({}), decryptions: {}% ({}), tx_scan: {}% ({}), anchors: {}",
+                "id: {}, blocks_done/blocks_total: {:4}/{:4}, decryptions: {:4}, tx_scan: {:4}, anchors: {}",
```
AloeareV, line 62: This display used to show the number of blocks done in the batch, as well as the total percentage of blocks done. (same with decryptions). Now, with the removal of `perct`, we see the number of blocks done in the batch, and the number of blocks in the batch, but we no longer see any obvious progress from batch to batch.

zancas: This is incoherent.   "Total percentage of _WHICH_ blocks done?"

All blocks on the chain?
Blocks anticipated for this particular sync?
Blocks in this batch?

Is it's intended to that the number of blocks scheduled for this sync should be displayed then I will make that a mut field of sync_status (if it's not already) and report that number (which increases on batch completion).

If someone wants to subsequently throw out the denominator to report a less informative % later in life, they can do so, since all of the relevant information will be available and clearly labeled.

Notice that they could _NOT_ go in the opposite direction if I just reported the %.

zancas: `sync_id` `blocks_done` `trial_dec_done` `blocks_total` `txn_scan_done` `orchard_anchors_done`


Do we need to add more?

AloeareV (approved):

PR #178, opened 2022-11-10
Save on sync

zancas:
This feature, saves per-batch processed.   I am not entirely clear on _when_ this is a good idea!


lib/src/lightclient.rs
```diff
@@ -1537,6 +1541,9 @@ impl LightClient {
             *self.wallet.verified_tree.write().await = highest_tree;
         }
 
+        debug!("About to run save after syncing {}th batch!", batch_num);
```
AloeareV, line 1544: Ah, the illustrious 2th batch!

AloeareV (approved):

PR #180, opened 2022-11-14
add compilation cfg canary

zancas:


zancas:
Deprecated in favor of compile_error based approach:

https://doc.rust-lang.org/std/macro.compile_error.html

PR #182, opened 2022-11-16
We now fetch orchard anchors only on send

AloeareV:


lib/src/wallet.rs
```diff
@@ -199,7 +198,7 @@ pub struct LightWallet {
 use crate::wallet::traits::{Diversifiable as _, ReadableWriteable};
 impl LightWallet {
     pub fn serialized_version() -> u64 {
-        return 25;
+        return 26;
```
zancas, line 201: Does this mean that existing `Zingo!` wallets will have to be reloaded?

AloeareV: Nope! That's the whole point of incrementing the version number.. When we read a v25 wallet, we know it had anchors saved, so we read them and ignore them.

AloeareV:
> Why not just make it on first send?

Well.. It's on any send where we don't have spendable orchard notes.

zancas (approved):

PR #183, opened 2022-11-17
add tests for client sending while not in sync

dannasessha:
Nothing like a `catch_unwind` yet, but tests for the issue I encountered, set to `#[should_panic]`...

`"Error: SendResponse { error_code: -26, error_message: \"tx-expiring-soon: expiryheight is 27 but should be at least 30 to avoid transaction expiring soon\" }"`


zancas:
This is fixed by this test:

https://github.com/zingolabs/zingolib/blob/0288aca2d861e97fc108a08f6f14ab8247ef0daa/cli/tests/integration_tests.rs#L30

PR #184, opened 2022-11-17
Interrupt sync

zancas:
Let's get this implemented next!

### DIFF
--- a/lib/src/lightclient.rs
+++ b/lib/src/lightclient.rs
@@ -312,6 +312,10 @@ impl LightClient {
         Ok(())
     }
 
+    pub async fn interrupt_sync_after_batch(&mut self) {
+        self.continue_sync = false;
+    }
+
     pub async fn get_initial_state(&self, height: u64) -> Option<(u64, String, String)> {
         if height <= self.config.sapling_activation_height() {
             return None;
--- a/lib/src/lightclient.rs
+++ b/lib/src/lightclient.rs
@@ -82,7 +82,7 @@ pub struct LightClient {
     sync_lock: Mutex<()>,
 
     bsync_data: Arc<RwLock<BlazeSyncData>>,
-    continue_sync: bool,
+    continue_sync: Arc<RwLock<bool>>,
 }
 
 use serde_json::Value;
@@ -176,7 +176,7 @@ impl LightClient {
             mempool_monitor: std::sync::RwLock::new(None),
             bsync_data: Arc::new(RwLock::new(BlazeSyncData::new(&config))),
             sync_lock: Mutex::new(()),
-            continue_sync: true,
+            continue_sync: Arc::new(RwLock::new(true)),
         })
     }
     pub fn set_server(&self, server: http::Uri) {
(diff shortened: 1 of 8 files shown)


lib/src/lightclient.rs
```diff
@@ -140,6 +141,7 @@ impl LightClient {
         &self,
         addrs: Vec<(&str, u64, Option<String>)>,
     ) -> Result<String, String> {
+        let transaction_submission_height = self.latest_served_blockheight().await;
```
AloeareV, line 144: Submission target should be 1 above latest served blockheight, we're aiming for the next block, not the one already mined.

lib/src/lightclient.rs
```diff
@@ -1517,6 +1540,7 @@ impl LightClient {
     }
 
     pub async fn do_shield(&self, address: Option<String>) -> Result<String, String> {
+        let transaction_submission_height = self.latest_served_blockheight().await;
```
AloeareV, line 1543: Same as above

lib/src/lightclient.rs
```diff
@@ -1559,8 +1584,20 @@ impl LightClient {
         result.map(|(transaction_id, _)| transaction_id)
     }
 
+    async fn latest_served_blockheight(&self) -> BlockHeight {
+        let latest_height = GrpcConnector::get_latest_block(self.config.get_server_uri())
+            .await
+            .unwrap()
+            .height;
+        BlockHeight::from_u32((latest_height - self.config.reorg_buffer_offset as u64) as u32)
```
AloeareV, line 1592: This function's name mislead me into think this function returns the latest server blockheight, not the latest served height minus our intended anchor offset.

AloeareV: I think the function is named what it wants to do, and this `-` bit should be removed.

lib/src/wallet.rs
```diff
@@ -632,7 +632,7 @@ impl LightWallet {
             .unwrap_or_default()
     }
 
-    async fn get_target_height(&self) -> Option<u32> {
+    async fn get_wallet_latest_height(&self) -> Option<u32> {
```
AloeareV, line 635: This function returns 1 + latest height, so calling it `wallet_latest_height` is incorrect.

lib/src/wallet.rs
```diff
@@ -1186,19 +1190,17 @@ impl LightWallet {
...
         println!("Selected notes worth {}", u64::from(selected_value));
 
-        let orchard_anchor = self
-            .get_orchard_anchor(&orchard_notes, target_height)
-            .await?;
+        let orchard_anchor = self.get_orchard_anchor(&orchard_notes, last_height).await?;
         let mut builder = Builder::with_orchard_anchor(
             self.transaction_context.config.chain,
-            target_height,
+            submission_height,
```
AloeareV, line 1203: Right now, this is `reorg_buffer_offset + 1` less than what the submission height should be, which is one above the highest block known to the server.

AloeareV (changes requested):

lib/src/wallet.rs
zancas: Wait...   why does this function return 1 + last_synced_height?

zancas: Yes.

AloeareV (approved):
Great!

PR #189, opened 2022-11-23
Orchard change with uas memo

AloeareV:


lib/src/blaze/fetch_full_transaction.rs
```diff
@@ -404,6 +406,50 @@ impl TransactionContext {
...
                 .unwrap_or(Memo::Future(memo_bytes));
+            if let Memo::Arbitrary(ref wallet_internal_data) = memo {
+                let mut uas_bytes = vec![0; 511];
+                wallet_internal_data
+                    .as_slice()
+                    .read(&mut uas_bytes)
+                    .unwrap();
+                let uas_string = String::from_utf8(uas_bytes);
+                if let Ok(uas_string) = uas_string {
+                    for ua_string in uas_string.split('!') {
```
zancas, line 417: Use HRP! to delimit to save space:  https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#bech32

lib/src/blaze/fetch_full_transaction.rs
```diff
@@ -163,6 +169,52 @@ impl TransactionContext {
...
+        &self,
+        arbitrary_memos_with_txids: Vec<([u8; 511], TxId)>,
+    ) {
+        for (wallet_internal_data, txid) in arbitrary_memos_with_txids {
+            let mut uas_bytes = vec![0; 511];
+            wallet_internal_data
+                .as_slice()
+                .read(&mut uas_bytes)
+                .unwrap();
+            let uas_string = String::from_utf8(uas_bytes);
```
zancas, line 182: We should be using this encoding:  https://zips.z.cash/zip-0316#encoding-of-unified-addresses

zancas: Unless there's a compression of that?

lib/src/blaze/fetch_full_transaction.rs
```diff
@@ -161,6 +169,52 @@ impl TransactionContext {
...
+                        RecipientAddress::decode(&self.config.chain, ua_string)
+                    {
+                        if let Some(transaction) = self
+                            .transaction_metadata_set
+                            .write()
+                            .await
+                            .current
+                            .get_mut(&txid)
+                        {
+                            let outgoing_potential_recievers = [
```
zancas, line 195: `i` before `e` except after `c`.

lib/src/blaze/fetch_full_transaction.rs
```diff
@@ -464,14 +534,23 @@ impl TransactionContext {
...
+                                ) {
+                                    if let Memo::Text(_) = memo {
+                                        Some(OutgoingTxMetadata {
+                                            address,
+                                            value: D::WalletNote::value_from_note(&note),
+                                            memo,
+                                            ua: None,
+                                        })
+                                    } else {
+                                        None
```
zancas, line 546: We need to consider this case carefully...

PR #197, opened 2022-12-07
reverse order of bytes in zcash-cli output

zancas:
I think we should just update this test expectation, because the `zcash-cli` output is only exposed to us as part of our test-framework.   In production `zingolib` talks to `lightwalletd`.


AloeareV (commented):
Agreed. Looks good!

AloeareV (approved):

PR #201, opened 2022-12-08
Description text

zancas:


app_description.txt
```diff
@@ -0,0 +1,49 @@
...
+seed-phrase, will be fully recovered even if you lose your device. 
+
+We're currently implementing another use of these on-chain memos, that will allow you to keep track (even if 
+you lose your device) of which addresses you previously produced to receive funds at. That is, you'll always
+have a full list of all the addresses you asked people to use to send you zcash.
+
+Taken together these two instrospective mechanisms provide you with insight into who you've sent
+funds to, and who you've asked for funds from.  This is just the beginning!  We'll soon implement
+mechanisms to allow users to share their insights in an anonymous, and economically viable way that
+will allow communities to aggregate around marketplaces of interesting insights!
```
AloeareV, line 30: I'm concerned about this assertion. I agree that this is an idea we're interested in pursuing, but "We'll soon implement" is stronger of a promise than I would like to make.

app_description.txt
```diff
@@ -0,0 +1,49 @@
...
+Only you have the power to share your information! Your perspective matters. You matter.
+
+When you send Zcash you can write a memo containing anything you like. When you receive
+Zcash you also receive the memo that the Sender wrote. As with most cryptocurrencies, wallets
+routinely send transactions to themselves to recover "change" from the transaction.  In this case, where
+the application is both the Sender and the Receiver, Zingo uses the memo system to write
+notes to itself about the transactions it builds for the User.
+
+Zingo extracts that data to provide insights exclusively to you.  In our initial release we
+already use this introspective mechanism to "remember" which addresses you've previously sent Zcash.
```
AloeareV, line 18: This already exists by-default for the sapling and transparent pools, so I'm worried that this sounds like we're making a mountain out of something that someone not familiar with the UA problem would think "My wallet already does that". Maybe "which full unified addresses you've previously sent Zcash" or something along those lines?

app_description.txt
```diff
@@ -0,0 +1,49 @@
...
+you lose your device) of which addresses you previously produced to receive funds at. That is, you'll always
+have a full list of all the addresses you asked people to use to send you zcash.
+
+Taken together these two instrospective mechanisms provide you with insight into who you've sent
+funds to, and who you've asked for funds from.  This is just the beginning!  We'll soon implement
+mechanisms to allow users to share their insights in an anonymous, and economically viable way that
+will allow communities to aggregate around marketplaces of interesting insights!
+
+Your data is never exposed to any third party. By providing you with private insight into
+your own economic behavior, we are enabling everyone to make healthier financial decisions.
```
AloeareV, line 33: This could maybe use some clarification about what information the lightwalletd is privy to.

app_description.txt
```diff
@@ -0,0 +1,49 @@
+Zingo! A Zcash Application
+
+Use zero knowledge cryptography to manage your transaction history in a private and mindful way.
```
AloeareV, line 3: I don't want to bury the lead on the fact that Zingo has full wallet functionality. Maybe something like:
```suggestion
Use zero knowledge cryptography to transact, and to manage your transaction history, in a private and mindful way.
```

AloeareV (changes requested):
Looks pretty good! A few specific concerns:

app_description.txt
zancas: Switched to "we intend to"

app_description.txt
zancas: How about `transaction data`?

AloeareV (approved):
Alright! My feedback has been addressed, to my satistfaction. Description looks good. I'm leaving this open, as I'd like to give the rest of the team a chance to make suggestions/proof-read. The normal policy for code is one additional reviewer, and I don't know that documentation needs to be more strenuous than that, so this is probably good to merge, but I don't want to make that call without consulting at least you.

juanky201271 (approved):
looks good to me.

Oscar-Pepper (approved):

PR #202, opened 2022-12-09
Fix syncstatus display of update_witnesses

AloeareV:
I believe this solves #191

### DIFF
--- a/lib/src/blaze/block_witness_data.rs
+++ b/lib/src/blaze/block_witness_data.rs
@@ -804,7 +804,7 @@ impl BlockAndWitnessData {
         // Replace the last witness in the vector with the newly computed one.
         WitnessCache::new(vec![w], height)
     }
-    pub(crate) async fn update_witness_after_reciept<D: DomainWalletExt<zingoconfig::ChainType>>(
+    pub(crate) async fn update_witness_after_receipt<D: DomainWalletExt<zingoconfig::ChainType>>(
         &self,
         height: &BlockHeight,
         transaction_id: &TxId,
--- a/lib/src/blaze/block_witness_data.rs
+++ b/lib/src/blaze/block_witness_data.rs
@@ -741,13 +741,12 @@ impl BlockAndWitnessData {
         return WitnessCache::new(fsb.into_vec(), top_block);
     }
 
-    pub(crate) async fn update_witness_after_pos<D: DomainWalletExt<zingoconfig::ChainType>>(
+    async fn update_witness_after_reciept_to_block_end<D: DomainWalletExt<zingoconfig::ChainType>>(
         &self,
         height: &BlockHeight,
         transaction_id: &TxId,
         output_num: u32,
         witnesses: WitnessCache<<D::WalletNote as ReceivedNoteAndMetadata>::Node>,
-        nullifier: <D::WalletNote as ReceivedNoteAndMetadata>::Nullifier,
     ) -> WitnessCache<<D::WalletNote as ReceivedNoteAndMetadata>::Node>
     where
         D::Recipient: crate::wallet::traits::Recipient,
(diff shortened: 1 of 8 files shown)


lib/src/blaze/block_witness_data.rs
```diff
@@ -725,7 +727,12 @@ impl BlockAndWitnessData {
                     yield_now().await;
                     blocks = self.blocks_in_current_batch.read().await;
                 }
-                self.sync_status.write().await.witnesses_updated = top_block - bottom_block - i;
+                self.sync_status.write().await.witnesses_updated.insert(
+                    <D::Bundle as crate::wallet::traits::Bundle<D, ChainType>>::Spend::wallet_nullifier(
```
zancas, line 731: `Spend::wallet_nullifier` superficially looks like an AssType instead of a method.   Can we rename it to `get_wallet_nullfier`?

zancas: Let's revisit the name `WitnessCache::top_height`.

zancas: Maybe for another time but "MAX_REORG" and "REORG_BUFFER_OFFSET" are confusingly similar concepts.

zancas: Why convert into and out of a 'fixed_size_buffer' if the eventual structure is a vec?

https://github.com/zingolabs/zingolib/pull/202/files#diff-fe24446938f05b862f0ff1d05aa869ec72d727a008b5c335db2cb868768b8a68R708

zancas: Yikes!   `yield_now`!  

https://github.com/zingolabs/zingolib/pull/202/files#diff-fe24446938f05b862f0ff1d05aa869ec72d727a008b5c335db2cb868768b8a68R693

Inside `wait_for_first_block`...  is it  likely that this is a bottleneck since it busy-waits in the case of 0-ready-threads?

https://doc.rust-lang.org/std/thread/fn.yield_now.html

AloeareV: > `Spend::wallet_nullifier` superficially looks like an AssType instead of a method. Can we rename it to `get_wallet_nullfier`?

I think the snake_case name makes this clear enough.. But.. This doesn't need to be an associated fn, I think a `from` impl is a better solution anyway. Will make a pr for that, and see if I can clean up some other traits as well.

AloeareV: > Maybe for another time but "MAX_REORG" and "REORG_BUFFER_OFFSET" are confusingly similar concepts.

Good point. This is worth thinking about, and perhaps consolidating.

AloeareV: > Why convert into and out of a 'fixed_size_buffer' if the eventual structure is a vec?
> 
> https://github.com/zingolabs/zingolib/pull/202/files#diff-fe24446938f05b862f0ff1d05aa869ec72d727a008b5c335db2cb868768b8a68R708

I suspect the fixed_size_buffer stuff is to avoid heap_allocation/resizing during sync for efficiency purposes, but it's a holdover from zecwallet that I haven't investigated much.

lib/src/blaze/block_witness_data.rs
```diff
@@ -797,7 +805,9 @@ impl BlockAndWitnessData {
         // Replace the last witness in the vector with the newly computed one.
         let witnesses = WitnessCache::new(vec![w], height);
 
-        return self.update_witness_after_block::<D>(witnesses).await;
+        return self
```
zancas, line 808: Why `return FOO;`?   Why not `FOO`?

lib/src/blaze/sync_status.rs
```diff
@@ -13,7 +16,7 @@ pub struct BatchSyncStatus {
     pub trial_dec_done: u64,
     pub txn_scan_done: u64,
 
-    pub witnesses_updated: u64,
+    pub witnesses_updated: HashMap<ChannelNullifier, u64>,
```
zancas, line 19: Let's update the name `ChannelNullfier` now that it's also used for `HashMap` keys.

lib/src/blaze/update_notes.rs
```diff
@@ -96,7 +96,13 @@ impl UpdateNotes {
                     .read()
                     .await
                     .block_data
-                    .update_witness_after_pos::<D>(&created_height, &txid, output_num, witnesses)
+                    .update_witness_after_pos::<D>(
```
zancas, line 99: Can we intelligently update `pos` to a really meaningful ident component?

zancas (approved):

zancas (approved):

PR #206, opened 2022-12-12
Set reorg buffer to 1

AloeareV:
Setting the buffer to 0 causes us an attempt to fetch orchard anchors from above the chain. It seems this number is 1-indexed. I'm not sure what's happening with that, exactly. Needs to be investigated.

### DIFF
--- a/cli/tests/integration_tests.rs
+++ b/cli/tests/integration_tests.rs
@@ -140,13 +140,6 @@ fn send_mined_sapling_to_orchard() {
         let balance = client.do_balance().await;
         // We send change to orchard now, so we should have the full value of the note
         // we spent, minus the transaction fee
-        assert_eq!(
-            balance["unverified_orchard_balance"],
-            625_000_000 - u64::from(DEFAULT_FEE)
-        );
-        assert_eq!(balance["verified_orchard_balance"], 0);
-        utils::increase_height_and_sync_client(&regtest_manager, &client, 1).await;
-        let balance = client.do_balance().await;
         assert_eq!(balance["unverified_orchard_balance"], 0);
         assert_eq!(
             balance["verified_orchard_balance"],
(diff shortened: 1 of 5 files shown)


AloeareV:
Got it. `last_height` was being used in several places where `submission_height` should have been used, and was 1 higher than it should have been to accommodate them.

cli/tests/integration_tests.rs
```diff
@@ -140,13 +140,6 @@ fn send_mined_sapling_to_orchard() {
         let balance = client.do_balance().await;
         // We send change to orchard now, so we should have the full value of the note
         // we spent, minus the transaction fee
-        assert_eq!(
-            balance["unverified_orchard_balance"],
-            625_000_000 - u64::from(DEFAULT_FEE)
-        );
-        assert_eq!(balance["verified_orchard_balance"], 0);
-        utils::increase_height_and_sync_client(&regtest_manager, &client, 1).await;
-        let balance = client.do_balance().await;
```
zancas, line 149: Why are these `assert_eq!` removed?

AloeareV: We were asserting that the funds where unverified, then syncing higher, then asserting that they're verified. Now they're verified immediately, so there's no need for the second step.

zancas: Hmmm...    they're not really "verified"...   we're just not waiting for verification to send?   I mean...   isn't there still an idea of something that's more certain?

config/src/lib.rs
```diff
@@ -72,7 +72,7 @@ impl ZingoConfig {
             server_uri: Arc::new(RwLock::new(http::Uri::default())),
             chain,
             monitor_mempool: false,
-            reorg_buffer_offset: 4,
+            reorg_buffer_offset: REORG_BUFFER_OFFSET,
```
zancas, line 75: Nice catch.

lib/src/wallet.rs
```diff
@@ -1472,7 +1474,8 @@ impl LightWallet {
                     .iter_mut()
                     .find(|u| utxo.txid == u.txid && utxo.output_index == u.output_index)
                     .unwrap();
-                spent_utxo.unconfirmed_spent = Some((transaction.txid(), u32::from(last_height)));
+                spent_utxo.unconfirmed_spent =
+                    Some((transaction.txid(), u32::from(submission_height)));
```
zancas, line 1478: Are we correct in cases where the submitted note is pushed out of the mempool without making it into a block?

zancas (approved):

zancas (approved):

PR #207, opened 2022-12-12
Assorted cleanup

AloeareV:


lib/src/wallet/traits.rs
```diff
@@ -718,7 +716,7 @@ impl ReceivedNoteAndMetadata for ReceivedOrchardNoteAndMetadata {
     }
 }
 
-pub trait DomainWalletExt<P: Parameters>: Domain + BatchDomain
+pub trait DomainWalletExt: Domain + BatchDomain
```
zancas, line 719: Why are we able to do this?   What suggested this change?  `clippy`?

zancas (approved):

lib/src/wallet/traits.rs
AloeareV: I remember trying and failing to do it in the past. I decided I wanted to try it again, and I don't know what I did wrong the first time, because it worked this time.

PR #213, opened 2022-12-19
Ci update

Oscar-Pepper:


Oscar-Pepper:
the Dockerfile is still setup for testing our zingo-cli-in-container issue but it is cleaned up.
only added debug to source code where it's getting stuck.

docker-ci/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# Parent image for zingolibs CI
+FROM parent-ci:latest
+
+# RUN git clone https://github.com/zingolabs/zingolib.git
+RUN git clone https://github.com/Oscar-Pepper/zingolib.git # test only
+
+# Copy zcash and lightwallet binaries into zingolib regtest directory
+WORKDIR zingolib
+RUN git checkout ci-update # test only
+RUN cp /usr/bin/lightwalletd /usr/bin/zcashd /usr/bin/zcash-cli ./regtest/bin/
```
zancas, line 10: Can we use `ln -s` instead of `cp`?

zancas: I suspect `ln -s` will  be more efficient/fast.

docker-ci/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+# Create CARGO_HOME directory for package caching
+# RUN mkdir ./cargo_home
+
+# ------------test only ----------------------
+RUN cargo build --release
```
zancas, line 16: Should we add the `cross_version` feature here?

docker-ci/parent/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    apt install sudo -y && \
+    echo $USERNAME ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/$USERNAME && \
+    chmod 0440 /etc/sudoers.d/$USERNAME
+USER $USERNAME
+
+# Install packages
+RUN sudo apt install curl build-essential gcc protobuf-compiler libssl-dev pkg-config git apt-transport-https wget gnupg2 unzip -y
+    
+# Install Rust
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | bash -s -- -y
```
zancas, line 20: Did we decide that the `-sSf` was unnecessary?

zancas: Nice `bash -s`

docker-ci/parent/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+ENV PATH="/home/${USERNAME}/.cargo/bin:${PATH}"
+
+# Install Zcash and fetch zero-knowledge parameters
+RUN wget -qO - https://apt.z.cash/zcash.asc | gpg --import && \
```
zancas, line 24: I think `wget` and `curl` are redundant..   it's likely you can do this with `curl` which means we don't need `wget` installed.

docker-ci/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# Parent image for zingolibs CI
+FROM parent-ci:latest
+
+# RUN git clone https://github.com/zingolabs/zingolib.git
```
zancas, line 4: I get why you're doing this now, and I think it's the right way, but once the image is in CI, I think the version we want to test against will be available through the github CI interface.

Oscar-Pepper:
work now tracked in new PR #219

docker-ci/Dockerfile
Oscar-Pepper: added in #221

docker-ci/Dockerfile
Oscar-Pepper: added to cargo test entrypoint in #221

docker-ci/Dockerfile
Oscar-Pepper: ok yes we are moving onto integrating to github now so will hook the parent image up to the CI workflow

docker-ci/parent/Dockerfile
Oscar-Pepper: i think its fine to keep -sSf, it silences except for errors and fail fast with no output on server fails returning an error code

docker-ci/parent/Dockerfile
Oscar-Pepper: yes nice idea, removed wget

PR #215, opened 2022-12-25
Error handle update price

zancas:


zancas:
This adds proper error handling to price fetch.   All `unwrap` statements are removed, so the app won't crash on a failed price fetch!   Instead it will (hopefully) produce a meaningful error string!

lib/src/lightclient.rs
```diff
@@ -87,28 +87,105 @@ pub struct LightClient {
...
-            .await?
-            .json::<Value>()
-            .await?
-            .as_array()
-            .unwrap()
-            .into_iter()
-            .map(repr_price_as_f64)
-            .collect();
-    trades.sort_by(|a, b| a.partial_cmp(b).unwrap());
+async fn get_recent_median_price_from_gemini() -> Result<f64, PriceFetchError> {
```
Oscar-Pepper, line 150: maybe this is more readable but could we drop all the 'let' by nesting the options/results e.g.

   150 async fn get_recent_median_price_from_gemini() -> Result<f64, PriceFetchError> {
   151     match reqwest::get("https://api.gemini.com/v1/trades/zecusd?limit_trades=11").await {
   152         Ok(httpget) => match httpget.json::<Value>().await {
   153             Ok(serialized) => {
   154                 if let Some(elements) = serialized.as_array() {
●  155                     match elements
   156                         .into_iter()
   157                         .map(repr_price_as_f64)
●  158                         .collect::<Vec<f64>>()
   159                     {
●  160                         Ok(trades) => {
   161                             if trades.iter().any(|x| x.is_nan()) {
   162                                 return Err(PriceFetchError::NanValue);
   163                             }
   164                             // NOTE:  This code will panic if a value is received that:
   165                             // 1. was parsed from a string to an f64
   166                             // 2. is not a NaN
   167                             // 3. cannot be compared to an f64
   168                             // TODO:  Show that this is impossible, or write code to handle
   169                             // that case.
   170                             trades.sort_by(|a, b| {
   171                                 a.partial_cmp(b).expect(
   172                                     "a and b are non-nan f64, I think that makes them comparable",
   173                                 )
   174                             });
   175                             Ok(trades[5])
   176                         }
●  177                         Err(e) => {
   178                             return Err(PriceFetchError::PriceReprError(e));
   179                         }
   180                     }
   181                 } else {
   182                     return Err(PriceFetchError::NoElements);
   183                 }
   184             }
   185             Err(_) => {
   186                 return Err(PriceFetchError::NotJson);
   187             }
   188         },
   189         Err(e) => {
   190             return Err(PriceFetchError::ReqwestError(e.to_string()));
   191         }
   192     }
   193 }

although this is not correct due to issues with the turbofish format on collect()

zancas:
I just force-pushed an update that separates the #215 work into its own PR.

lib/src/lightclient.rs
```diff
@@ -87,28 +87,105 @@ pub struct LightClient {
...
+}
+impl std::fmt::Display for PriceReprError {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        fmt.write_str(self.to_string().as_str())
+    }
+}
+fn repr_price_as_f64(from_gemini: &Value) -> Result<f64, PriceReprError> {
+    if let Some(value) = from_gemini.get("price") {
+        if let Some(stringable) = value.as_str() {
+            if let Ok(parsed) = stringable.parse::<f64>() {
```
Oscar-Pepper, line 137: match statement for the Result is better here?

lib/src/lightclient.rs
zancas: Could do..   maybe just push your version as a subsequent PR?

zancas:
@AloeareV what would you think about landing this before the test is finished?

AloeareV (approved):
This looks like a pretty straightforward upgrade. Merging, and this will inspire me to hurry up getting the test finished.

PR #222, opened 2022-12-28
CI optimisation

Oscar-Pepper:


zancas (dismissed):

zancas:
Still draft?

.github/workflows/main.yml
```diff
@@ -13,20 +13,19 @@ jobs:
     env:
       RUSTFLAGS:                -D warnings
     container:
-      image:                    zingodevops/ci-build:latest
+      image:                    zingodevops/ci-build:stable
```
zancas, line 16: What about the policy of selecting a constant version?  Is that unavailable in Debian?

Oscar-Pepper: sorry i don't understand, regarding the docker image? ```stable``` is the tag i gave our current "in-use" CI docker image, my thoughts were latest (the default tag) could be overwritten with something unstable and it break our CI

Oscar-Pepper: we could start a version system if you prefer i.e. ```zingodevops/ci-build:0```

Oscar-Pepper: i just realised this may be confusing as this image is under development and shares the same tag as our actual ```stable``` image in dockerhub which we are using for CI. This is due to me working on this PR before the CI-update which took some small improvements from this PR. when i continue working on this i will change the name from ```stable``` to ```latest``` or maybe ```test``` and upload to dockerhub so there is a distinction between images we are using and developping

PR #223, opened 2022-12-29
reduce batch size by 10x

zancas:
We need to discuss this before letting it integrate into mobile.

### DIFF
--- a/zingolib/src/lightclient.rs
+++ b/zingolib/src/lightclient.rs
@@ -1265,7 +1265,7 @@ impl LightClient {
 
         // Re-read the last scanned height
         let last_scanned_height = self.wallet.last_synced_height().await;
-        let batch_size = 1_000;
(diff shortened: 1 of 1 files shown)


zancas:
How can we safely test this?

AloeareV:
> How can we safely test this?

Right now, we have no tests that involve large numbers of blocks... Generating blocks is computationally expensive, so I don't think we want to tell regtest to generate thousands of blocks during a test run. I think we want to generate a large blockchain (ideally several thousand blocks), with some sends back and forth, and save it as a test vector somewhere. That way, we can spend a long time generating it once, and then we can have tests that rescan it.

zancas:
Great answer, but in the short term I was thinking of a manual test!

juanky201271 (approved):
Well, I think this change It's pretty straightforward... and we have to test it ASAP.

PR #225, opened 2023-01-02
updated github actions to v3 and updated Dockerfile

Oscar-Pepper:


Oscar-Pepper:
added a "stable" tag to the docker image and added some changes to match format and some of the flags etc. used in zingo-mobile Dockerfile. changed parent image to bullseye-slim and also stopped rustup toolchain from being downloaded twice with a symlink

zancas (approved):
LGTM, and CI passes.   Hooray fot v3!

PR #227, opened 2023-01-05
Test multiple workflows

Oscar-Pepper:


Oscar-Pepper:
test revealed that zingo-mobile will build HEAD of feature branch and not the specific commit that trigger the test. improvements can be made to build specific commits instead of using github ref

PR #230, opened 2023-01-22
Compilation targets test

Oscar-Pepper:


Oscar-Pepper:
tested that builds fail both on pc and mobile builds when compile error flags are reversed. this PR separates all code for each build target and also adds compile errors for safety

zingoconfig/src/lib.rs
```diff
@@ -28,12 +28,12 @@ pub const MAX_REORG: usize = 100;
...
-    0
-} else {
-    5
-};
+
+#[cfg(any(target_os = "ios", target_os = "android"))]
+pub const GAP_RULE_UNUSED_ADDRESSES: usize = 0;
+
+#[cfg(not(any(target_os = "ios", target_os = "android")))]
+pub const GAP_RULE_UNUSED_ADDRESSES: usize = 5;
```
zancas, line 36: Is this the code you said was dead code?

Oscar-Pepper: not in zingolib but pub so will leave for now

zancas (approved):

PR #232, opened 2023-01-26
cleaning mobile build

Oscar-Pepper:
not compiling unused commands in zingo-mobile


zancas:
What's going on with this..  does it contain relevant work?  Or is it obsolete?

macintoshhelper:
In case this PR is still active/isn’t obsolete: I feel it would be best to not remove `encryptmessage` and `decryptmessage` from mobile compilation, as there are use cases where they could be exposed via mobile wallet UI.

PR #233, opened 2023-02-01
Publish interrupt

zancas:
This exposes interruption so that consumers can toggle it.


juanky201271:
Looks good, but maybe is good that @AloeareV take a look to this new `command`.

AloeareV (approved):

PR #236, opened 2023-02-02
cratify zingo-memo-utils

AloeareV:


zancas:
I propose the name `zingomemo` or `zingo_memo` or `zingo-memo`, such that when a developer uses the public interface they express things like:

`zingomemo::unpack`
`zingomemo::pack`
`zingomemo::authenticate_nym`
`zingomemo::record_nym`
`zingomemo::distrust_nym`
`zingomemo::sign_message`

Cargo.toml
```diff
@@ -2,7 +2,8 @@
 members = [
     "zingolib",
     "zingocli",
-    "zingoconfig"
+    "zingoconfig",
+    "zingo-memo-utils"
```
zancas, line 6: Shall we rename this to "zingomemo"?

zingo-memo-utils/src/lib.rs
```diff
@@ -4,16 +4,23 @@ use zcash_address::unified::{Address, Container, Encoding, Receiver};
...
+/// corresponding to the key we sent with.
 #[non_exhaustive]
+#[derive(Debug)]
 pub enum ParsedMemo {
     Version0 { uas: Vec<UnifiedAddress> },
 }
 
+/// Packs a list of UAs into a memo. The UA only memo is version 0 of the protocol
+/// Note that a UA's raw representation is 1 byte for length, +21 for a T-receiver,
+/// +44 for a Sapling receiver, and +44 for an Orchard reciever. This totals a maximum
```
zancas, line 19: `i` before `e` _EXCEPT_ after "c".

zingo-memo-utils/src/lib.rs
```diff
@@ -4,16 +4,23 @@ use zcash_address::unified::{Address, Container, Encoding, Receiver};
...
+/// of 110 bytes per UA, and attemtping to write more than 510 bytes will cause an error.
```
zancas, line 20: `attemtping` --> `attempting`

Cargo.toml
zancas: Once we make that name change I am +1 on landing this.

Oscar-Pepper: "zingomemo" matches the format of our other crates but i think "zingo-memo" is more readable which is my preference

PR #244, opened 2023-02-16
Verify pool upgrade paths

zancas:
I think this verifies that all the 7 possible clearances of lower-protocol pools into orchard for self-spends achieves the correct balances.

### DIFF
--- a/zingocli/tests/integration_tests.rs
+++ b/zingocli/tests/integration_tests.rs
@@ -42,7 +42,7 @@ fn verify_old_wallet_uses_server_height_in_send() {
 
         // Without sync push server forward 100 blocks
         utils::increase_server_height(&regtest_manager, 100).await;
-        let orchard_receiver = get_address_string!(client_receiving, "orchard");
+        let orchard_receiver = get_base_address!(client_receiving, "orchard");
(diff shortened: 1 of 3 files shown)


zancas:
Hi @AloeareV this will land the #242 so if it passes your review we can just merge this to take care of both.

zingocli/tests/integration_tests.rs
```diff
@@ -756,22 +1017,18 @@ fn send_to_ua_saves_full_ua_in_wallet() {
         scenarios::two_clients_one_saplingcoinbase_backed();
     tokio::runtime::Runtime::new().unwrap().block_on(async {
         utils::increase_height_and_sync_client(&regtest_manager, &sender, 5).await;
-        let recipient_address = recipient.do_addresses().await[0]["address"].take();
+        let recipient_orchard_address = get_base_address!(recipient, "unified");
```
AloeareV, line 1020: It's a UA, not an 'orchard address'.

AloeareV (changes requested):
A terminology quibble, other than that it looks great!

AloeareV (approved):

PR #254, opened 2023-02-23
change type names to be more human readable

zancas:
@jarys

This naming scheme seems easier to parse and understand to me.   I certainly had a hand in the previous names, but after more experience I feel like they're unwieldy and not really more explicit.
For the moment I left the local variables with 'metadata' containing idents.  I'm not sure that's a bad thing.   Obviously this PR might be irritating due to conflicts, so I will leave it draft.


jarys:
Ok. Yes, new names sounds better and they are more explicit.

AloeareV:
Delegating this one to @Oscar-Pepper, once it's ready

zancas:
Abandoning, we can reconsider names later.

PR #255, opened 2023-02-23
refactor implementation and rename

zancas:
Refactor of transaction returning method.

This method is called 5 times.

Four out 5 of those times it has an identical arg list, and is the first call in a public wrapper.

Maybe we can add a new wrapper that calls this method and passes its result to the 5 wrappers?

### DIFF
--- a/zingolib/src/wallet/transactions.rs
+++ b/zingolib/src/wallet/transactions.rs
@@ -510,7 +510,8 @@ impl TransactionMetadataSet {
         });
     }
 
-    fn get_or_create_transaction_metadata(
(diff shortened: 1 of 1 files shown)


zancas:
Once the call to this internal method [get_wallet_transaction](https://github.com/zingolabs/zingolib/pull/255/files#diff-428904fdafe177d7b87505c16a515aaf0a39c7cb69beaed93c2ec4fba243a8baR514) is factored out of the 5 `add_$NOTE` wrapper `fn` the parameter list for `get_wallet_transaction` can be removed from the `add_` wrapper, this is because all of the parameters are exposed as `pub` members of the `TransactionMetadata`.

PR #259, opened 2023-02-27
let people know where the wallet is, on collision

zancas:

PR #258, opened 2023-02-26
Minor cli updates

zancas:

PR #260, opened 2023-02-27
Change sync message

zancas:
Print out the block height during sync.

Needs to be tuned in the case of sync completion.


fluidvanadium:
these are improvements, however what is happening at zingolib/src/lightclient/tests.rs? surely that's a generated file so you didn't repeat yourself?

zancas:
Superceded by #258 and #259

PR #263, opened 2023-02-28
Recover test upgrade

zancas:
I clipped out the unspent_note test and test upgrades when I force pushed to dev.


AloeareV (approved):
A bit scary that we lost this stuff. Good to get it back. Tests pass, looks good.

PR #268, opened 2023-03-03
[#267] debug output leaking into CLI non-interactive calls

### LINKED ISSUE #267: Debug and Informational output leaking into CLI non-interactive calls
When calling the CLI 

````
[zingolib/src/grpc_connector.rs:361] uri.clone() = https://mainnet.lightwalletd.com:9067/
[zingolib/src/grpc_connector.rs:368] &client = CompactTxStreamerClient {
    inner: Grpc {
        inner: BoxCloneService,
    },
}
Lightclient connecting to https://mainnet.lightwalletd.com:9067/
```` is printed regardless of the log policy that the user selected.

### PR DESCRIPTION
**Pull request by pacu**
This checks that the CLI has been invoked in interactive mode before printing informational strings on startup.

Non-Interactive invocations will have not initial output.

Also removes dbg! call dangling when creating a GRPC connection and removes dbg call from GrpcConnecto constructor
Closes #267

### DIFF
--- a/zingoconfig/src/lib.rs
+++ b/zingoconfig/src/lib.rs
@@ -5,7 +5,7 @@ use std::{
     sync::{Arc, RwLock},
 };
 
-use log::{info, LevelFilter};
+use log::LevelFilter;
 use log4rs::{
     append::rolling_file::{
         policy::compound::{
@@ -150,7 +150,7 @@ impl ZingoConfig {
                 #[cfg(not(any(target_os = "macos", target_os = "windows")))]
                 {
                     if dirs::home_dir().is_none() {
-                        info!("Couldn't determine home dir!");
+                        log::info!("Couldn't determine home dir!");
                     }
                     zcash_data_location =
                         dirs::home_dir().expect("Couldn't determine home directory!");
--- a/zingocli/src/lib.rs
+++ b/zingocli/src/lib.rs
@@ -429,12 +429,14 @@ pub fn startup(
         }
     };
 
-    // Print startup Messages
-    info!(""); // Blank line
-    info!("Starting Zingo-CLI");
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by zancas — APPROVED** [review]

**Comment by zancas** [comment]
Very nice!  Thanks!

File: zingoconfig/src/lib.rs [hunk 1]
```diff
@@ -5,7 +5,7 @@ use std::{
     sync::{Arc, RwLock},
 };
 
-use log::{info, LevelFilter};
```
zancas, line 8: This import is needed, to provide access to the `info!` macro called later in the file.   Other than that this PR look good.

zancas (changes requested):
See the line note in zingoconfig/src/lib.rs

zingoconfig/src/lib.rs
pacu: can this be added conditionally with that macro as well so it's obvious that the import it's needed in such macros?

pacu: I added a change to address this based on this answer in stackoverflow

https://stackoverflow.com/questions/58167542/how-to-group-imports-in-rust-with-conditional-compilation

zancas (approved):
I re-added the deleted `info`.   After the checks pass this can be merged.

zingoconfig/src/lib.rs
zancas: Ahh..   just read this.

Shall we explicitly qualify instead of using `use` at all?

zancas (dismissed):
Prefer explicit qualification over a `use` statement, especially in a case where the qualified name is only referred to once.

zancas (approved):
I rebased to add my signature to the commit.

PR #272, opened 2023-03-07
Port view tests to integration

zancas:


zingocli/tests/integration_tests.rs
```diff
@@ -9,13 +9,272 @@ use json::JsonValue;
...
+        );
+
+        // Extract viewing keys
+        let wc = original_recipient
+            .extract_unified_capability()
+            .read()
+            .await
+            .clone();
+        use zcash_primitives::sapling::keys::DiversifiableFullViewingKey as SaplingFvk;
+        //let extfvk: SaplingFvk = (&wc).try_into().unwrap();
```
zancas, line 79: What's the deal with this?   It seems to be a different version of:

`let s_fvk = Fvk::Sapling(SaplingFvk::try_from(&wc).unwrap().to_bytes());`

a few lines later.  I'd like a reviewer to sanity check me on that before I remove the line.

zancas: I removed it.

AloeareV (approved):

PR #273, opened 2023-03-08
Async client

zancas:


zingolib/src/lightclient.rs
```diff
@@ -260,6 +260,45 @@ impl LightClient {
     }
 }
 impl LightClient {
+    /// The wallet this fn associates with the lightclient is specifically derived from
+    /// a spend authority.
+    pub async fn new_from_wallet_base_async(
```
AloeareV, line 265: Given that we're adding an `async_client` feature, why is this not `#[cfg(async_client)]`?

zingolib/src/lib.rs
```diff
@@ -21,6 +21,48 @@ use std::{
 use tokio::runtime::Runtime;
 use zingoconfig::{ChainType, ZingoConfig};
 
+#[cfg(feature = "async_client")]
```
AloeareV, line 24: Why do we need to feature-gate this?

zingolib/src/lib.rs
```diff
@@ -21,6 +21,48 @@ use std::{
...
+pub async fn create_zingoconfdir_async(
+    server: http::Uri,
```
AloeareV, line 26: This is the same code as the sync version, expect without the runtime. Can we dry this? Make them both call a shared helper, or remove the feature flag so that `create_zingoconf_with_datadir` can directly call this function?

zingolib/src/lightclient.rs
```diff
@@ -260,6 +260,45 @@ impl LightClient {
...
+        wallet_base: WalletBase,
+        config: &ZingoConfig,
+        birthday: u64,
+        overwrite: bool,
+    ) -> io::Result<Self> {
+        #[cfg(not(any(target_os = "ios", target_os = "android")))]
+        {
+            if !overwrite && config.wallet_exists() {
+                return Err(Error::new(
```
AloeareV, line 274: As with `create_zingoconfdir_async`, this should be straightforward to dry.

AloeareV (changes requested):
Removing the runtime creation from every test in favor of using `tokio::test` is _so much_ nicer, I have some quibbles with the feature gate and wet code.

zancas:
Nice comments, I intend to clean up `zingolib` as soon as I finish porting `cross_compat`.

zancas:
The reason zingolib has crude wet hacks like these is because I wanted to make sure the test port would work, before I put effort into making something clean in `zingolib`.

zingolib/src/lib.rs
zancas: This code isn't currently used by any consumer, _except_ the tests, I thought it made sense to gate it..   at least until when/if more consumers use it.

zingolib/src/lib.rs
zancas: DRYed.

zingolib/src/lightclient.rs
zancas: DRYed

zancas:
OK, now this is review ready.

zancas:
Consider #275 which is the same changes _except_ to Cargo.lock..

Which makes the PR a net reduction in lines of code!  ;)

AloeareV (approved):

PR #280, opened 2023-03-14
New memo version

AloeareV:


zingo-memo/src/lib.rs
```diff
@@ -11,159 +11,44 @@ use zcash_encoding::{CompactSize, Vector};
 #[non_exhaustive]
 #[derive(Debug)]
 pub enum ParsedMemo {
-    Version0 { uas: Vec<UnifiedAddress> },
```
zancas, line 14: Isn't the above doc-comment out of date?   Don't we need a brief description of a `txindex-memo` (or whatever we want to call it)?

zingo-memo/src/lib.rs
```diff
@@ -11,159 +11,44 @@ use zcash_encoding::{CompactSize, Vector};
...
-    } else {
-        uas_bytes[..uas_bytes_vec.len()].copy_from_slice(uas_bytes_vec.as_slice());
-        Ok(uas_bytes)
-    }
+    Version0 {
+        uas: Vec<UnifiedAddress>,
+    },
+    Version1 {
+        uas: Vec<UnifiedAddress>,
+        transaction_heights_and_indexes: Vec<(BlockHeight, usize)>,
```
zancas, line 19: Is there a case for a format that doesn't include a `ua` vector?    Since we know that we're send to an internal address?

AloeareV: The ua vector is in there for the same reason it already was, marking the full UA we sent to, as we're piggybacking off of change memos at the moment. That Ua vector can also always be empty if it needs to be, so adding a new format version to occasionally save a byte doesn't seem very important even if we start doing these memos in note cultivation.

zingo-memo/src/lib.rs
AloeareV: Oh, yes, it is!

zingo-memo/src/memo_serde.rs
```diff
@@ -0,0 +1,156 @@
...
+    ParsedMemo,
+};
+use zcash_client_backend::address::UnifiedAddress;
+use zcash_encoding::{CompactSize, Vector};
+use zcash_primitives::consensus::BlockHeight;
+
+/// Packs a list of UAs into a memo. The UA only memo is version 0 of the protocol
+/// Note that a UA's raw representation is 1 byte for length, +21 for a T-receiver,
+/// +44 for a Sapling receiver, and +44 for an Orchard receiver. This totals a maximum
+/// of 110 bytes per UA, and attempting to write more than 510 bytes will cause an error.
```
zancas, line 18: Why 510?  ` max(44X + 21Y+1)  <  511` where:  X,Y --> [0, 1, 2,...] ?

zingo-memo/src/memo_serde.rs
```diff
@@ -0,0 +1,156 @@
...
+pub fn create_memo_v0(uas: impl AsRef<[UnifiedAddress]>) -> io::Result<[u8; 511]> {
```
zancas, line 19: Would "`where T: AsRef<[UnifiedAddress]>`" enhance readability?

zingo-memo/src/memo_serde.rs
```diff
@@ -0,0 +1,156 @@
...
+        ))
+    } else {
+        uas_bytes[..uas_bytes_vec.len()].copy_from_slice(uas_bytes_vec.as_slice());
+        Ok(uas_bytes)
+    }
+}
+
+pub fn create_memo_v1(
+    uas: impl AsRef<[UnifiedAddress]>,
+    mut transaction_heights_and_indexes: Vec<(BlockHeight, usize)>,
```
zancas, line 39: Can't we force this to be a u16 or u32..   what's the biggest txindex?

zingo-memo/src/memo_serde.rs
```diff
@@ -0,0 +1,156 @@
...
+        write_unified_address_to_raw_encoding(&ua, &mut w)
+    })?;
+    let mut last_noted_height = None;
+    transaction_heights_and_indexes.sort_unstable();
+    transaction_heights_and_indexes.reverse();
+    let heights_indexes_and_target_heights =
+        transaction_heights_and_indexes
+            .iter()
+            .fold(Vec::new(), |mut acc, (height, index)| {
+                acc.push((*height, *index, last_noted_height));
```
zancas, line 53: Why flip the order of `height` and `index` in this transform?

zingo-memo/src/memo_serde.rs
AloeareV: 512, -1 byte for arbitrary-memo, -1 byte for zingo-memo-version.

zingo-memo/src/memo_serde.rs
AloeareV: `CompactSize::write` takes a `usize` as its argument, so we either take a usize ourselves, or take something else and cast/convert.

zingo-memo/src/memo_serde.rs
AloeareV: Done

PR #281, opened 2023-03-15
Add txindex

AloeareV:
We now store a transaction's position-in-block, so we can use it for memo-sync.


zingocli/tests/integration_tests.rs
```diff
@@ -1085,6 +1085,23 @@ async fn send_to_ua_saves_full_ua_in_wallet() {
...
+    let recipient_unified_address = get_base_address!(recipient, "unified");
+    let sent_value = 50_000;
+    faucet
+        .do_send(vec![(recipient_unified_address.as_str(), sent_value, None)])
+        .await
+        .unwrap();
+    utils::increase_height_and_sync_client(&regtest_manager, &faucet, 1).await;
+    let list = faucet.do_list_transactions(false).await;
+    for transaction in list.members() {
+        assert_ne!(transaction["txindex"], JsonValue::Null);
```
zancas, line 1101: How about a more positive test that matches a specific txid?   It'd be more sensitive.

PR #300, opened 2023-03-29
DRY buggy interface, reveal test bug that permitted back-incompatible…

zancas:
… wallet deser

How 'bout them apples?


AloeareV (commented):
Unfortunately, there's a good (depending on how you define good) reason for this duplication. In most (non-mobile) cases, storing the wallet path in the config is the most sensible way to do things. The `read_wallet_from_disk` function already exists, and does basically what you've defined `read_wallet_from_buffer` to do. 

The reason we have a `read_wallet_from_buffer` public function which takes the wallet file as an argument (despite its location being defined in the config) is because on mobile, `File::read` will fail. We need to read the file from native kotlin, and pass it to rust as a buffer.

PR #308, opened 2023-04-17
Fix wallet version parse bug, move fvk source to WalletCapability

zancas:


zingolib/src/lightclient.rs
```diff
@@ -1662,7 +1662,12 @@ impl LightClient {
     pub async fn do_shield(&self, address: Option<String>) -> Result<String, String> {
         let transaction_submission_height = self.get_submission_height().await?;
         let fee = u64::from(DEFAULT_FEE);
-        let tbal = self.wallet.tbalance(None).await;
+        let tbal = self
+            .wallet
+            .tbalance(None)
+            .await
+            .as_u64()
+            .expect("To represent Json as u64");
```
AloeareV, line 1670: Currently, I think this will cause a taddr-less view_only wallet to panic if `do_shield` is called on it. This isn't likely to come up, but it's also probably easier to `?` it instead of expecting _anyway_.

Edit: `tbalance` is still infallible.. I don't know that that's correct, but the worst-case scenario is someone in view-only mode encounters slightly confusing behavior that's the current behaviour anyway, so I'm not too worried about it.

AloeareV (changes requested):

zingolib/src/lightclient.rs
zancas: There're a couple of issues here, that we should discuss.  In particular there're two different mechanisms to filter results for taddresses:

*  `by address` and `by key`

If you don't have a view capability for taddresses, should you be prevented from checking their balances?  What if you have a specific address (itself a capability) and you are interested in "its" balance..  and you don't have the key for it?

AloeareV: The mechanism we expose for 'being interested in a taddr's balance' _is_ having a view capability for that address.

AloeareV (approved):

AloeareV (approved):

PR #313, opened 2023-04-20
updated CI to sync with zingomobile

Oscar-Pepper:


.github/workflows/ci.yaml
```diff
@@ -41,60 +41,134 @@ jobs:
...
+          repository: zingolabs/zingo-mobile
+      
+      - name: Check if native rust cache exists
+        id: check-build-cache
+        uses: actions/cache@v3
+        with:
+          path: android/app/src/main/jniLibs/${{ matrix.abi }}
+          key: native-rust-${{ matrix.abi }}-${{ env.CACHE-KEY }}
+          lookup-only: true
+          fail-on-cache-miss: true
```
AloeareV, line 105: This seems to be causing the integration test jobs to also fail due to the cache miss...is this intended?

Oscar-Pepper:
Yes it is designed to fail on cache miss which triggers the build job ☺️

.github/workflows/ci.yaml
Oscar-Pepper: Are you sure this is failing integration tests? The cache is designed to fail when the rust has changed which triggers the build job

Oscar-Pepper: Looking at the logs it is the build that is failing. cI is working as intended but there is errors in the rust build

AloeareV (approved):
Awesome!

PR #316, opened 2023-04-21
Improve note selection

AloeareV:
I haven't made any of the functional changes I'd like to...including adding some more user-specification for what privacy leaks are allowed (or manual pool specification, even)...but this is certainly tidier, and will be easier to work with.

### DIFF
--- a/zingolib/src/lightclient.rs
+++ b/zingolib/src/lightclient.rs
@@ -73,6 +73,12 @@ impl WalletStatus {
     }
 }
 
+impl Default for WalletStatus {
+    fn default() -> Self {
+        WalletStatus::new()
+    }
+}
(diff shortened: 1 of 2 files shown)


zancas (approved):

zancas:
Why not mark this as review ready and merge it?  It seems like a clear improvement to me.

I am getting a CI failure, but the test pass locally.

zancas:
I vote we land this before building on...  possibly with clippification of the two changed files each in their own commit.

AloeareV:
> I vote we land this before building on top of it... possibly with clippification of the two changed files each in their own commit.
 Deal.

zancas (approved):

PR #318, opened 2023-04-23
Reject trailing whitespace in CI; add utility; fix trailing whitespace

nathan-at-least:
This PR adds `./utils/trailing-whitespace.sh` to either detect-and-reject trailing whitespace or to just fix it. It only runs these checks on "well-known" file formats based on extension: `.rs`, `.md`, `.toml`, and `.yaml`

It then integrates trailing whitespace rejection into CI.

⚠ I am not confident the CI change is correct. I assume the current directory for a step is the repo root.

Finally, it uses the tool to fix existing trailing whitespace.

### Why?

- Trailing whitespace is almost never intentional
- Rejecting all trailing whitespace keeps files consistent across devs and environments making commit diffs only specific to semantic changes instead of whitespace changes. (For example, my editor always removes trailing whitespace on any save, which means when I make small code changes to some zingo files, I inadvertently change whitespace elsewhere in the file which makes the commit diffs noisier / confusing.)
- Although I'm not certain this matters, it looks like `cargo fmt` may have one or more bugs where it gets confused by trailing whitespace; see [this issue search](https://github.com/rust-lang/rustfmt/issues?q=is%3Aissue+is%3Aopen+trailing+whitespace)

### DIFF
--- a/zingocli/tests/integration_tests.rs
+++ b/zingocli/tests/integration_tests.rs
@@ -86,7 +86,6 @@ async fn load_wallet(dir: PathBuf) -> zingolib::wallet::LightWallet {
 }
 
 #[tokio::test]
-#[traced_test]
 async fn load_and_parse_different_wallet_versions() {
     let (_sap_wallet, _sap_path, sap_dir) = get_wallet_nym("sap_only").unwrap();
     let _loaded_wallet = load_wallet(sap_dir).await;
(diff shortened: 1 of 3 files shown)


utils/trailing-whitespace.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+set -efuo pipefail
```
zancas, line 2: It seems like `-e` is redundant with `fuo`?

zancas (approved):
@Oscar-Pepper please check this BASH for backdoors when you have time.

PR #321, opened 2023-04-24
Clippy continuous integration

nathan-at-least:
⚠ This PR extends #320; if #320 is rejected, this needs to be rebased.

This PR fixes all outstanding clippy errors, then attempts to integrate `cargo clippy` into continuous integration.

The `cargo clippy` tool checks many ergonomic and style issues, and sometimes potential bugs. Benefits:

- Ensure the style in zingo matches what standard rust devs are familiar with.
- Teaches zingo devs newer to rust new language and library idioms, improving their rust skill as they go.
- Potentially catch some bugs.

Note that specific clippy checks can be bypassed by using explicit attributes on specific items. This PR used this approach where I thought that fixing the clippy errors would require refactors that seemed too invasive.

Wherever possible I removed code rather than fixing the clippy error when I determined the code was dead.

Every commit used `./util/git-hook-pre-commit.sh` thus passes all of those checks.

⚠ I never ran the fuller integration checks.

### DIFF
--- a/utils/git-hook-pre-commit.sh
+++ b/utils/git-hook-pre-commit.sh
@@ -21,3 +21,4 @@ cargo fmt -- --check
 cargo check
 cargo test --bins --lib
 #cargo test --doc # No doc tests yet.
+cargo clippy
--- a/utils/git-hook-pre-commit.sh
+++ b/utils/git-hook-pre-commit.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+#
+# zingolib pre-commit hook
+#
+# zingolib devs can ensure consistent commits by using this pre-commit
+# hook. Install with:
+#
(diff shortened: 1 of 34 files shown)


utils/git-hook-pre-commit.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+#
+# zingolib pre-commit hook
+#
+# zingolib devs can ensure consistent commits by using this pre-commit
+# hook. Install with:
+#
+# $ ln -s ../../utils/git-hook-pre-commit.sh .git/hooks/pre-commit
+
+set -efuo pipefail
```
zancas, line 10: `-e` is always true when either of `fu` is true, so why set all three?

zancas (dismissed):

zancas (dismissed):

PR #328, opened 2023-05-01
Add cargo cache to CI

Oscar-Pepper:


.github/workflows/ci.yaml
```diff
@@ -18,19 +18,26 @@ jobs:
...
         run:  rustup show
+
       - name: Check formatting of code
         run: cargo +stable fmt --check
-      - name: Clippy Checks
+
+      - name: Clippy checks
         run: cargo +stable clippy
 
   test:
```
fluidvanadium, line 43: name: Test
can this be more specific?

fluidvanadium (changes requested):

.github/workflows/ci.yaml
Oscar-Pepper: hmm what do you suggest? unit-and-integration-test maybe?

Oscar-Pepper: or cargo-test ?

Oscar-Pepper: all our test jobs are not plural so `Cargo test` fits the format and also is the actual command so i think thats better

fluidvanadium (approved):

PR #334, opened 2023-05-05
enable fail fast for integ tests in ci

Oscar-Pepper:


Oscar-Pepper:
> Hi @Oscar-Pepper is now the time to revisit this?

this didnt come up on my github notifications sorry for late reply... yes this should have merge a long time ago but its out of date now we can close it. the fail fast is enabled in zingo mobile now and zingolib uses zingomobile's CI

PR #336, opened 2023-05-08
We no longer panic when failing to connect to server on blockheight request

AloeareV:
This needs one change in zingo-mobile to adapt to the new type, which I will open as a zingo-mobile PR

### DIFF
--- a/zingocli/src/lib.rs
+++ b/zingocli/src/lib.rs
@@ -427,7 +427,10 @@ pub fn startup(
                 let server_uri = config.get_lightwalletd_uri();
                 let block_height = zingolib::get_latest_block_height(server_uri);
                 // Create a wallet with height - 100, to protect against reorgs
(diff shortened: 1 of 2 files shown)


zingocli/src/lib.rs
```diff
@@ -427,7 +427,10 @@ pub fn startup(
                 let server_uri = config.get_lightwalletd_uri();
                 let block_height = zingolib::get_latest_block_height(server_uri);
                 // Create a wallet with height - 100, to protect against reorgs
-                Arc::new(LightClient::new(&config, block_height.saturating_sub(100))?)
+                Arc::new(LightClient::new(
+                    &config,
+                    block_height?.saturating_sub(100),
```
zancas, line 432: I find this syntax confusing.

zancas (approved):

PR #351, opened 2023-05-25
retool shield command to shield sapling and/or transparent pools with argument

AloeareV:
@juanky201271 when this lands, the command `shield` will now need an argument of 'transparent', 'sapling', or 'all'. I think we want 'all' to be what the shield button does...once we have basic/advanced mode, that's something we'll want to allow to be configurable in advanced mode.

### DIFF
--- a/zingolib/src/commands.rs
+++ b/zingolib/src/commands.rs
@@ -1,5 +1,5 @@
 use crate::wallet::keys::is_shielded_address;
-use crate::wallet::MemoDownloadOption;
+use crate::wallet::{MemoDownloadOption, Pool};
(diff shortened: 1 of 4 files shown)


zingolib/src/commands.rs
```diff
@@ -489,30 +489,39 @@ struct ShieldCommand {}
 impl Command for ShieldCommand {
     fn help(&self) -> &'static str {
         indoc! {r#"
-            Shield all your transparent funds
+            Shield all your transparent and/or orchard funds
```
zancas, line 492: I think you mean `and/or sapling funds`.

zancas (approved):

PR #358, opened 2023-05-31
Add total data to address interface

zancas:


juanky201271 (approved):
the test are passed, my rust level is low, but I trust in @zancas

PR #362, opened 2023-06-02
changed initial min_height from 0 to block height of first tx in summary

Oscar-Pepper:


juanky201271 (dismissed):
I think this is going to work.

juanky201271 (approved):
looks good.

PR #367, opened 2023-06-05
Add birthday to exportufvk, and align different methods of getting birthday

AloeareV:
exportufvk now returns an object including birthday, and we once again always calculate the birthday the same way.

### DIFF
--- a/zingolib/src/commands.rs
+++ b/zingolib/src/commands.rs
@@ -77,7 +77,7 @@ impl Command for GetBirthdayCommand {
     }
 
     fn exec(&self, _args: &[&str], lightclient: &LightClient) -> String {
-        RT.block_on(async move { lightclient.do_get_birthday().await.to_string() })
+        RT.block_on(async move { lightclient.wallet.get_birthday().await.to_string() })
     }
 }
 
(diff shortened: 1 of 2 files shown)


zingolib/src/commands.rs
```diff
@@ -600,7 +600,11 @@ impl Command for ExportUfvkCommand {
         match ufvk_res {
             Ok(ufvk) => {
                 use zcash_address::unified::Encoding as _;
-                ufvk.encode(&lightclient.config().chain.to_zcash_address_network())
+                object! {
+                    "ufvk" => ufvk.encode(&lightclient.config().chain.to_zcash_address_network()),
+                    "birthday" => RT.block_on(lightclient.wallet.get_birthday())
```
zancas, line 605: Why pack this into this API when there's a purpose specific `get_birthday` already?

zingolib/src/lightclient.rs
```diff
@@ -787,19 +787,6 @@ impl LightClient {
         }
         finsight::TotalMemoBytesToAddress(memobytes_by_address)
     }
-    pub async fn do_get_birthday(&self) -> u64 {
```
zancas, line 790: Nice.

juanky201271 (commented):
IMO: In my experience when you load a new wallet (seed or ufvk) you send to the server `the user birthday (correct or incorrect)` and after that when you run the `sync` & `syncstatus` commands, you obtain the info related to that `birthday`, which apparently is stored in the wallet session... and when you run `seed` to obtain both values (seed & user birthday)... This behavior for me is coherent... I need the same behavior when you `load from ufvk`... because now I can't get the user birthday through `exportufvk`... and the value from `get_birthday (first block with tx)` is useless to me, because it's nothing to do with the birthday from the user...

zingolib/src/commands.rs
AloeareV: As with `seed`, the birthday is something you almost always want in the context where this function is called....and if you forget or don't know you need to export your birthday when you export your key/seed, it can be really inconvenient. So we return the birthday to make it obvious that it's something you probably want if you're asking for the seed/key.

AloeareV: It also makes a more convenient interface for zingo-mobile/zingo-pc.

zancas (approved):

PR #368, opened 2023-06-06
updated readme and added birthday flag to examples

Oscar-Pepper:
is this still true now we have re-implemented the manual shielding button?

https://github.com/zingolabs/zingolib/blob/dev/README.md?plain=1#L19

i think this was a security risk maybe? where an attacker could dust a non-orchard pool and watch the note management automatically shield when the user sent funds?

"releases are currently only provisional" what does this mean? is this correct?
https://github.com/zingolabs/zingolib/blob/dev/README.md?plain=1#L6

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,9 @@ Zingo-CLI does automatic note and utxo management, which means it doesn't allow 
 * Defaults to sending shielded transactions, even if you're sending to a transparent address
 * Sapling funds need at least 5 confirmations before they can be spent
 * Can select funds from multiple shielded addresses in the same transaction
-* Will automatically shield your transparent funds at the first opportunity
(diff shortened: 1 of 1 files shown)


AloeareV:
> is this still true now we have re-implemented the manual shielding button?
> 
> https://github.com/zingolabs/zingolib/blob/dev/README.md?plain=1#L19

It's still true for sapling, I think! It's no longer true for transparent. 

> i think this was a security risk maybe? where an attacker could dust a non-orchard pool and watch the note management automatically shield when the user sent funds?

You can still attempt to dust sapling and learn which transactions someone is sending if they don't shield first. It's no where near as leaky as it was when you could do the same with the transparent pool, but it's still a viable attack

> "releases are currently only provisional" what does this mean? is this correct? https://github.com/zingolabs/zingolib/blob/dev/README.md?plain=1#L6

AloeareV (approved):
Good improvements!

PR #395, opened 2023-06-21
Count outputs to be trial decrypted

zancas:


zingolib/src/blaze/block_witness_data.rs
```diff
@@ -63,13 +63,14 @@ pub struct BlockAndWitnessData {
     orchard_activation_height: u64,
 }
 
+pub const BATCHSIZE: u32 = 25;
```
fluidvanadium, line 66: yes

AloeareV (dismissed):
This is pretty great! It's atop #399, so the diff includes a lot of stuff not actually in the scope of this PR...so I locally observed the diff between this and 399. I've pushed a commit fixing a couple things, @zancas if you're happy with the changes I've added, feel free to merge once 399 lands.

AloeareV (approved):

PR #397, opened 2023-06-26
Viewkey

fluidvanadium:


zingolib/src/commands.rs
```diff
@@ -97,13 +97,13 @@ impl Command for WalletKindCommand {
 
     fn exec(&self, _args: &[&str], lightclient: &LightClient) -> String {
         RT.block_on(async move {
-            if lightclient.do_seed_phrase().await.is_ok() {
+            if lightclient.get_wallet_kind().await.is_ok() {
```
zancas, line 100: This doesn't seem like a well-defined fn for publishing whether a wallet is view-key only.

zingolib/src/lightclient.rs
```diff
@@ -1063,6 +1063,7 @@ impl LightClient {
             .read()
             .await
             .sync_id;
+        dbg!();
```
zancas, line 1066: Let's chop this out?

zingolib/src/lightclient.rs
```diff
@@ -1162,6 +1162,16 @@ impl LightClient {
         *self.interrupt_sync.read().await
     }
 
+    pub async fn get_wallet_kind(&self) -> Result<JsonValue, &str> {
```
zancas, line 1165: If I read only the name of the fn in calling code..  then I will get really confused about the return types.

How about we define a two-variant enum:

```Rust
enum WalletKind {
       Spend,
       View,
}
```

And we have `get_wallet_kind` return a `WalletKind`?

zingolib/src/commands.rs
zancas: I am +1 on the logical flow of this exec implementation.

zancas (changes requested):
I posted some line notes.   I would like to migrate the `get_wallet_kind` implementation to use an enum...    we could pack data into the variants..   but I prefer a simpler fn that just provides a concise way to control the flow in the caller, and does nothing else.

PR #406, opened 2023-07-05
Time rming instrumentation

zancas:


AloeareV (approved):
Local benchmarks indicate that this about doubles the speed of sync compared to dev....I'm so glad we made these benchmarks. Would not have expected the added logging to interfere with sync to the level it did. Good that we caught it before release.

PR #411, opened 2023-07-11
Add viewkey only benchmark

zancas:
* reorganize LightClient impl blocks into constructors and other
* alphabetize each impl block
* remove `integration_tests` feature gate
* unify test utils
* export and use fvks for a client
* rename ClientManager to ClientBuilder


fluidvanadium (dismissed):
ok

fluidvanadium (approved):
this is good

PR #421, opened 2023-07-26
run ci outside docker container

Oscar-Pepper:


Oscar-Pepper:
did not yield any benefits

PR #426, opened 2023-08-08
simplified cli keyloading

fluidvanadium:


zingocli/src/lib.rs
```diff
@@ -299,14 +297,13 @@ impl ConfigTemplate {
...
-        let viewing_key = matches.value_of("viewing-key").map(|s| s.to_string());
+        let from = matches.value_of("from").map(|s| s.to_string());
         let maybe_birthday = matches.value_of("birthday");
-        if seed.is_some() && maybe_birthday.is_none() {
+        if from.is_some() && maybe_birthday.is_none() {
             eprintln!("ERROR!");
             eprintln!(
-            "Please specify the wallet birthday (eg. '--birthday 600000') to restore from seed."
-        );
+                "Please specify the wallet birthday (eg. '--birthday 600000') to restore a wallet. (If you want to load the entire blockchain instead, you can use birthday 0. /this would require extensive time and computational resources)"
```
AloeareV, line 305:
```suggestion
                "Please specify the wallet birthday (eg. '--birthday 600000') to restore a wallet. (If you want to load the entire blockchain instead, you can use birthday 0. This would require extensive time and computational resources)"
```

AloeareV (dismissed):
Great!!

fluidvanadium:
> Great!!

wait wait wait
can i do `
impl WalletBase {
     fn FromString`
instead? wouldnt that be cleaner?

AloeareV:
> > Great!!
> 
> wait wait wait can i do ` impl WalletBase { fn FromString` instead? wouldnt that be cleaner?

You could do this, yes! It would be mildy cleaner, but yes, it would be cleaner!

AloeareV:
Hang on...you could `impl Parse for WalletBase`, I think that would be even better

zancas (approved):

PR #440, opened 2023-08-18
Drop obsolete paragraph from README

AArnott:
The `patch.crates-io` section from Cargo.toml no longer exists.


zancas (approved):
Nice catch!  Thanks.

PR #442, opened 2023-08-19
Make persisted file names configurable

AArnott:
This allows the zingolib to be used by other applications without them having to share a wallet file.

### DIFF
--- a/zingoconfig/src/lib.rs
+++ b/zingoconfig/src/lib.rs
@@ -25,8 +25,8 @@ use zcash_primitives::{
 
 pub const DEFAULT_LIGHTWALLETD_SERVER: &str = "https://mainnet.lightwalletd.com:9067";
 pub const MAX_REORG: usize = 100;
-pub const WALLET_NAME: &str = "zingo-wallet.dat";
-pub const LOGFILE_NAME: &str = "zingo-wallet.debug.log";
+pub const DEFAULT_WALLET_NAME: &str = "zingo-wallet.dat";
+pub const DEFAULT_LOGFILE_NAME: &str = "zingo-wallet.debug.log";
 pub const REORG_BUFFER_OFFSET: u32 = 0;
 
 #[cfg(any(target_os = "ios", target_os = "android"))]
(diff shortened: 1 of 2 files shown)


zingolib/src/lib.rs
```diff
@@ -52,6 +52,8 @@ pub fn load_clientconfig(
         monitor_mempool,
         reorg_buffer_offset: zingoconfig::REORG_BUFFER_OFFSET,
         zingo_wallet_dir: data_dir,
+        wallet_name: DEFAULT_LOGFILE_NAME.into(),
```
zancas, line 55: I think you meant `DEFAULT_WALLET_NAME.into()` here.

AArnott:
I'm not sure why an audit of dependencies failed with my change, which doesn't change dependencies. Is there any work _I_ need to do here?

zancas:
We'll @juanky201271 's approval once we're sure our `zingo-pc` and `zingo-mobile` dependencies are compatible.

zingoconfig/src/lib.rs
```diff
@@ -74,6 +79,8 @@ impl ZingoConfig {
             monitor_mempool: false,
             reorg_buffer_offset: REORG_BUFFER_OFFSET,
             zingo_wallet_dir: dir,
```
juanky201271, line 81: Maybe it's a good idea to change `zingo_wallet_dir` to `wallet_dir`, I don't know if this can be so much trouble...

juanky201271 (changes requested):

juanky201271:
@AArnott can you update your fork because it is 123 commit behind, and I get weird errors when I try to build it in zingo-mobile?

AArnott:
I've rebased and force pushed, @juanky201271

zingoconfig/src/lib.rs
AArnott: Done.

juanky201271 (dismissed):

juanky201271:
I built the rust libs for zingo-mobile & zingo-pc and both were working fine without any change in the code. Great!!! @AArnott

juanky201271:
zingo-pc is working well.

But zingo-mobile is not, when I run the `command` `sync` I got this: `task 4754 panicked`... and I don't know if this is related with this PR or maybe with the last major PR about `shardtree`...

@AArnott If you build zingo-cli in this branch and you run `sync` you obtain the same `error`?

AArnott:
sync works for me on this branch under zingo-cli.

juanky201271 (approved):
looks great.

PR #443, opened 2023-08-19
Add support for targeting win-arm64

AArnott:
The `ring` crate doesn't natively support win-arm64 yet. But a fork adds that support [as described here][workaround].

- [x] Add support for building the `aarch64-pc-windows-msvc`.
- [ ] Verify that building `x86_64-pc-windows-msvc` hasn't been broken.
- [ ] Add a job to the GitHub workflow that runs in CI/PR to build Windows x64 and arm64 to demonstrate and keep it working.
 
[workaround]: https://github.com/briansmith/ring/issues/1514#issuecomment-1258562375

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1473,7 +1473,7 @@ version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e2abad23fbc42b3700f2f279844dc832adb2b2eb069b2df918f455c4e18cc646"
 dependencies = [
- "spin",
+ "spin 0.5.2",
 ]
 
 [[package]]
@@ -2465,15 +2465,13 @@ dependencies = [
 [[package]]
 name = "ring"
 version = "0.16.20"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3053cf52e236a3ed746dfc745aa9cacf1b791d846bdaf412f60a8d7d6e17c8fc"
+source = "git+https://github.com/awakecoding/ring?branch=0.16.20_alpha#ec78aff5b5c4734957522f092590e0b58a3161cf"
 dependencies = [
  "cc",
  "libc",
  "once_cell",
- "spin",
+ "spin 0.9.8",
  "untrusted",
- "web-sys",
  "winapi",
 ]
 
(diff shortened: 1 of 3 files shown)


AArnott:
Unfortunately, building `cargo build --target x86_64-pc-windows-msvc` with this change fails during linking with such as this:

> libring-f96eea544d4a0e80.rlib(ring-f96eea544d4a0e80.ring.e6b10a1c-cgu.2.rcgu.o) : error LNK2001: unresolved external symbol ring_core_0_16_20_sha256_block_data_order

Ironically, the arm64 build is fine. But we can't merge this while it breaks win-x64 builds.

AloeareV:
@AArnott is this still draft? It looks like a complete 1-line change that I'd be happy to merge

AArnott:
In my testing, this fixes `win-arm64` builds but breaks `win-x64` builds. So I'm surprised the PR checks didn't see that. We should confirm manually that it doesn't break local dev box builds, at least.

AArnott:
Also, it would make the PR more useful if it actually added a win-arm64 build to the CI to prove it works.

AArnott:
Oh ya... the CI doesn't even test building win-x64 right now. That's why it doesn't catch the problem with this PR as-is.
I'll update the PR description with a checklist.

juanky201271:
@AArnott I see the `0.17.5`, I'm trying to patch the ring crate in `zingo-pc` to this version to build in `aarch64`, but I realized that in zingolib there are several different versions of ring and I got an error in the patching... dis you try to build with the last version of ring?

AArnott:
The challenge with the latest version of `ring` with its win-arm64 support is that the version number changed in a more significant way that cargo _patches_ are not allowed to cross. So we have to get every transitive path to `ring` to upgrade. I and some others have already started chasing these dependencies down. I haven't checked back on their progress for a few weeks though, so I guess I'll resume that.

AArnott:
`cargo tree -i ring@0.16.20` is useful to find what needs to be upgraded.

https://github.com/jjyr/hdwallet/pull/21 is one such upgrade that we're still waiting on.

AArnott:
I'm going to close this PR in favor of a new ones that takes the non-hacky approach (just updating dependencies as they are available to get to ring 0.17.)

PR #447, opened 2023-08-22
Produce strongly typed struct as sync result

AArnott:
Prior to this, the rather predictable data was returned as a JSON object, which made programmatic extraction uncertain to succeed and runtime and largely undocumented. With this, a caller gets a struct with predictable fields.


AArnott:
What is this format check that's failing? It shows diffs, but it doesn't call out what's wrong.

skyl (approved):
utACK

fluidvanadium (approved):
much better!! thank you

zancas (approved):
LGTM!  Thanks @AArnott !

PR #448, opened 2023-08-22
Add .editorconfig to repo

AArnott:
This helps ensure all editors will handle the code files the same way.
For example, my editor was inserting tabs into .rs file which clearly use spaces. By adding this file, the editor always uses spaces for this repo, regardless of my editor's default.s


fluidvanadium (approved):
this is an asset

PR #452, opened 2023-08-27
[#451] Detach --regtest from local zcashd environment setup

### LINKED ISSUE #451: Detach `--regtest` from local zcashd environment setup
Since `RegtestManager` was introduced, Zingolib attempts to start up the cache <> zcashd <> lightwalletd  setup. 

This breaks the darksidewalletd tests on regtest since `RegtestManager` kicks in when the `--regtest` flag is specified.

In order to be able to point to a `lightwalletd` server in `--darkside-very-insecure` mode this has to be modified so callers 
can tell Zingolib that they won't be needing the `RegtestManager` at all.

pacu:
This breaks the darksidewalletd tests on regtest since RegtestManager kicks in when the --regtest flag is specified.

In order to be able to point to a lightwalletd server in --darkside-very-insecure mode this has to be modified so callers can tell Zingolib that they won't be needing the RegtestManager at all.

closes #451

### DIFF
--- a/zingocli/src/lib.rs
+++ b/zingocli/src/lib.rs
@@ -63,6 +63,11 @@ pub fn build_clap_app() -> clap::App<'static> {
                 .value_name("regtest")
                 .help("Regtest mode")
                 .takes_value(false))
+            .arg(Arg::new("darkside-test")
+                .long("darkside-test")
+                .value_name("darkside-test")
+                .help("Uses darksidewalletd. Assumes that the lightwalletd server specified is launched in darkside mode.")
+                .takes_value(false))
             .arg(Arg::new("no-clean")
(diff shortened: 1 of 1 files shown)


AloeareV:
I don't think this is necessary. The `--regtest` flag is a convenience for the specific scenario of zingo managing all of the binaries for you, but since the darkside tests don't want that, I think `--chain regtest` will do what you want here: connect to a regtest server but otherwise act like normal zingo.

AloeareV:
@pacu let me know if you still need this, and I'll merge it

pacu:
@AloeareV can you reopen this? I've updated it to the latest dev, don't merge because I need to test it again with the new code

PR #454, opened 2023-08-28
Adding a new command to delete the wallet file -> for zingo-pc

juanky201271:
I tested from zingo-pc (windows) and it works, it removed the wallet file making easy to the user to change to another wallet.
This new command do `nothing`  for IOS or Android, like the command `save`.
And for zingo-cli, I'm not sure, in my testing it did nothing at all even in non-interactive mode.

@zancas @AloeareV Please, take a look carefully, because as you know `rust` is not my thing.


fluidvanadium (approved):
this is correct

PR #456, opened 2023-08-31
stop ignoring promotion test

zancas:


fluidvanadium (approved):
sure? if ci works with it

PR #459, opened 2023-09-01
Initialize wallet with unified spending key

AArnott:
This unlocks the ability for importing a wallet across FFI that has spending keys, but for which the mnemonic is not known, or perhaps the account index is not the 0 presumed by `WalletBase` when given a mnemonic.

### DIFF
--- a/zingolib/src/wallet/keys/unified.rs
+++ b/zingolib/src/wallet/keys/unified.rs
@@ -10,8 +10,10 @@ use bip0039::Mnemonic;
 use byteorder::{ReadBytesExt, WriteBytesExt};
 use orchard::keys::Scope;
 
+use secp256k1::SecretKey;
 use zcash_address::unified::{Container, Encoding, Fvk, Ufvk};
 use zcash_client_backend::address::UnifiedAddress;
+use zcash_client_backend::keys::{Era, UnifiedSpendingKey};
 use zcash_encoding::Vector;
 use zcash_primitives::{
     legacy::TransparentAddress, sapling::note_encryption::PreparedIncomingViewingKey,
@@ -22,7 +24,7 @@ use zingoconfig::ZingoConfig;
 use crate::wallet::traits::ReadableWriteable;
 
 use super::{
-    extended_transparent::{ExtendedPubKey, KeyIndex},
+    extended_transparent::{ExtendedPrivKey, ExtendedPubKey, KeyIndex},
     get_zaddr_from_bip39seed, ToBase58Check,
 };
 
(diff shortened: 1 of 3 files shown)


zingolib/src/wallet/keys/unified.rs
```diff
@@ -352,6 +353,19 @@ impl WalletCapability {
...
 
+    /// Creates a new `WalletCapability` from a unified spending key.
+    pub fn new_from_usk(usk: &[u8]) -> Result<Self, String> {
+        // Decode unified spending key
+        let usk = UnifiedSpendingKey::from_bytes(Era::Orchard, usk)
+            .map_err(|_| "Error decoding unified spending key.")?;
+        Ok(Self {
+            orchard: Capability::Spend(usk.orchard().to_owned()),
+            sapling: Capability::Spend(usk.sapling().to_owned()),
+            // transparent: Capability::Spend(usk.transparent()),
```
AArnott, line 364: Does anyone know how to translate an `legacy::AccountPrivKey` to an `ExtendedPrivKey`?

The private key that we have here is the `m/44'/133'/account'` derivation. I'm not certain whether `WalletCapability` wants that, or a `m/44'/133'/account'/0/0` derivation, but I suspect the former, so that it can read 'change' UTXOs.

AArnott: `AccountPrivKey` apparently has exactly one field, which is an `ExtendedPrivKey`. So no conversion necessary. Just access to that field, which is _not_ public, is required. Ironically, `zcash_primitives` doesn't appear to expose that field for reading at all(!). So I may be forced to call its `to_bytes()` method to serialize it out, and then deserialize it. What a pain.

AArnott: https://github.com/zcash/librustzcash/issues/929

zingolib/src/wallet/keys/unified.rs
```diff
@@ -352,6 +354,25 @@ impl WalletCapability {
         Ok(Self::new_from_seed(config, &bip39_seed, position))
     }
 
+    /// Creates a new `WalletCapability` from a unified spending key.
+    pub fn new_from_usk(usk: &[u8]) -> Result<Self, String> {
+        // Decode unified spending key
+        let usk = UnifiedSpendingKey::from_bytes(Era::Orchard, usk)
+            .map_err(|_| "Error decoding unified spending key.")?;
+
+        // Workaround https://github.com/zcash/librustzcash/issues/929 by serializing and deserializing the transparent key.
```
AloeareV, line 363: I did a bit of digging, and it turns out the `TransparentPrivKey` type being used here is actually a zingolib type, albeit a redefinition/copy-paste of the `hdwallet` type with a bunch of additional methods....I think it would make more sense to use the `hdwallet` type (either sacrifice a little bit of method call syntax and define local helpers instead of being method calls, or just define a local trait with the additional functionality in order to keep the method syntax) to avoid the surprising amount of duplicated code I just discovered, but that's not a blocking concern and definitely beyond the scope of this PR.

AloeareV (approved):
Thanks! Looks great.

zingolib/src/wallet/keys/unified.rs
zancas: Where's the `hdwallet` type defined?   Why do we have to copy paste?

AArnott: IIRC it was because the struct we might have reused carried private fields, and exposed them only in a limited way.
But as I think @AloeareV was alluding to, I may have even been looking at a similarly named, but incompatible type anyway, so maybe this was all necessary. I'd have to go back and check.

PR #468, opened 2023-09-07
Return struct from do_balance and do_send_progress instead of JSON

AArnott:


AArnott:
@juanky201271 to take a particular look at this.

juanky201271:
It seems perfect

> @juanky201271 to take a particular look at this.

I tested `balance` & `sendprogress` from your branch `nerdcash:stronglyTypedSyncResult2` and the result is perfectly aligned with what is expecting `zingo-mobile`. Thanks.

I'm not the person for reviewing the `rust` stuff... sorry.

fluidvanadium (approved):
another great step toward saner data handling

AArnott:
Ah, I authored that commit on my ubuntu VM which doesn't sign. I've pushed a fix.

PR #469, opened 2023-09-07
Take raw memo bytes when sending transactions

AArnott:
This allows clients that have already encoded the memo to simply supply it. 
Those clients that have text can do what commands.rs now does in utilizing the existing API to do the encoding.

As it was, the core library had been translating text into binary just because the text happened to start with `0x`. This seems more appropriate to be an app-level decision, particularly because the app can better handle an error case where the memo is not in fact hex.


Oscar-Pepper:
looks great to me! i have requested zancas / aloeareV to give it a second look as they are more familiar with this part of the code than me

AloeareV (approved):
This makes a lot more sense as part of the send command, nice going!

PR #473, opened 2023-09-08
Add .gitattributes file

AArnott:
Every repo should have this. It ensures that no matter how git itself or text editors are configured on each developer's machine, line endings will always be checked in with LF line endings.
Without this file, CRLF endings can be checked in, leading to huge diffs of nothing but line endings and senseless merge conflicts.


fluidvanadium (approved):
ok

PR #474, opened 2023-09-10
Support non-zero account indexes

AArnott:
This allows for wallets to be created with a mnemonic and an explicit (non-zero) account index.

### DIFF
--- a/zingo-testutils/src/lib.rs
+++ b/zingo-testutils/src/lib.rs
@@ -400,7 +400,7 @@ pub mod scenarios {
                 //! A "faucet" is a lightclient that receives mining rewards
                 let zingo_config = self.make_unique_data_dir_and_load_config();
                 LightClient::create_from_wallet_base_async(
-                    WalletBase::MnemonicPhrase(self.seed.clone(), 0),
+                    WalletBase::MnemonicPhrase(self.seed.clone()),
                     &zingo_config,
                     birthday,
                     overwrite,
@@ -416,7 +416,7 @@ pub mod scenarios {
             ) -> LightClient {
                 let zingo_config = self.make_unique_data_dir_and_load_config();
                 LightClient::create_from_wallet_base_async(
-                    WalletBase::MnemonicPhrase(mnemonic_phrase, 0),
+                    WalletBase::MnemonicPhrase(mnemonic_phrase),
                     &zingo_config,
                     birthday,
                     overwrite,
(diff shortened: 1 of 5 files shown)


AArnott:
We should probably save the account index with the mnemonic in the wallet and file. I'll work toward that. But depending on how resilient the data file is to adding new data, this may be a wallet-file breaking change.

AArnott:
I figured out the file versioning thing. Please pay extra attention to the conditions around which we read and write the extra data to verify I did it right.

fluidvanadium:
the existing struct initializers should stay the same, with a default index of 0.

AArnott:
Thanks for reviewing, @fluidvanadium. I'm still learning rust, so if you'll pardon the question: are you asking me to move the account_index field elsewhere in the struct so that it *may* be initialized but isn't required?
I'm not sure what a "struct initializer" is.

zingo-testutils/src/lib.rs
```diff
@@ -400,7 +400,7 @@ pub mod scenarios {
                 //! A "faucet" is a lightclient that receives mining rewards
                 let zingo_config = self.make_unique_data_dir_and_load_config();
                 LightClient::create_from_wallet_base_async(
-                    WalletBase::MnemonicPhrase(self.seed.clone()),
```
fluidvanadium, line 403: so that this doesnt need to change

zingolib/src/wallet.rs
```diff
@@ -190,9 +190,9 @@ impl WalletOptions {
 /// Data used to initialize new instance of LightWallet
 pub enum WalletBase {
     FreshEntropy,
-    SeedBytes([u8; 32]),
-    MnemonicPhrase(String),
-    Mnemonic(Mnemonic),
+    SeedBytes([u8; 32], u32),
+    MnemonicPhrase(String, u32),
```
fluidvanadium, line 194: please leave this enum variant as is. instead you could create a new variant

zingo-testutils/src/lib.rs
AArnott: Done

AArnott:
The validation failure is because [an agent ran out of disk space](https://github.com/zingolabs/zingolib/actions/runs/6192323714/job/16812205236?pr=474#step:7:8477).

AloeareV (approved):
This looks great! Sorry to leave it in limbo for two weeks

AloeareV:
We've improved our CI since you made this PR, and it seems to be using the old CI flow. I'm testing this locally instead.

AloeareV:
All tests pass after merge into dev, locally. Bypassing required checks, as the failure is in the outdated CI trying to run, and not in the code.

PR #475, opened 2023-09-11
Stop writing to STDOUT

AArnott:
A library should never write to STDOUT of its own accord. It doesn't know what app it's running in and how STDOUT is being used.

We have the zingo-cli and a .NET CLI that I'm writing where this logging was polluting the UX with debugging info.


fluidvanadium (approved):
appreciated

PR #478, opened 2023-09-12
Darkside foray

fluidvanadium:
transaction proven to exist on-chain, zingo recipient isn't detecting it for some reason..this doesn't seem to happen in happy-path integration tests, something about the darkside environment is causing the recipient to not notice incoming transparent funds even though sender successfully detects the send during rescan

docs updated


fluidvanadium:
why is this unverified? this is the result of
git log --show-signature
gpg: Signature made Tue 12 Sep 2023 03:57:41 PM UTC
gpg:                using RSA key 628940176302182DF3342B8BD25FF64056B88C25
gpg: Good signature from " (533) <fluidvanadium@zingolabs.org>" [ultimate]
Author: fluidvanadium <fluidvanadium@zingolabs.org>
Date:   Tue Sep 12 15:57:41 2023 +0000

zancas:
@pacu do you understand what the issue is with darkside transparent receipts?

zancas:
Should we keep this PR open?

PR #485, opened 2023-09-14
add an error message to docs/TEST_TROUBLESHOOTING

fluidvanadium:


docs/TEST-SETUP-NOTES.txt
```diff
@@ -0,0 +1,21 @@
+cargo install cargo-nextest
```
Oscar-Pepper, line 1: add --locked as per cargo nextest installing from source instructions
https://nexte.st/book/installing-from-source.html

PR #491, opened 2023-09-19
assert pre-sync transactions are a specific set

zancas:


zancas:
This minor test upgrade constrains the conditions under which the test passes to match specific transaction lists.   In particular it makes assertions about the the pre-"reorged" and post-"reorged" lists being different.

AloeareV (approved):

PR #493, opened 2023-09-19
move read and then write lock to just get write lock in the first place

AloeareV:


zancas:
Hi @AloeareV I converted this back to a draft, since the CI tests are all failing.

I will try to run locally..

PR #492, opened 2023-09-19
Mark txid nf spent error

AloeareV:
Note that three tests are now failing at an explicit todo! macro invocation. I believe this to be the pre-existing error, that's now failing early instead of silently doing the wrong thing.

PR #496, opened 2023-09-20
removed FromBytes unused import

Oscar-Pepper:
fix build warnings introduced in #492


Oscar-Pepper:
reverting #492 instead

PR #501, opened 2023-09-21
define empty_trees fn to make start_sync readable, and more correct

zancas:
This makes the start_sync function more readable and restricts the definition of a "fresh" wallet to exclude wallets that previously had sapling value commitments.

Here're questions:

Do we have a buggy edge case more generally where a wallet is pre-orchard, but has sapling spends?

Do we have a test of a wallet that was initialized pre-orchard and had funds, and is then synced post-orchard?

### DIFF
--- a/zingolib/src/lightclient.rs
+++ b/zingolib/src/lightclient.rs
@@ -1462,6 +1462,37 @@ impl LightClient {
         *lc.mempool_monitor.write().unwrap() = Some(h);
     }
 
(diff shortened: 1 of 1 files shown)


AloeareV (approved):
Great!

AloeareV:
Not reverting this, as it's an improvement and not a regression....but I think this is insufficient. We want to check if _either_ tree is empty, not only if _both_ are.

PR #502, opened 2023-09-21
Don't unwrap a tree that might not be there

AloeareV:


fluidvanadium (changes requested):
we should do this sanity check for both orchard AND sapling

AloeareV:
Closing in favor of a PR I can push to

PR #506, opened 2023-09-21
Refactor send to address

zancas:
@Oscar-Pepper this is a good start into `zingolib` it's just a refactor that makes `send_to_adresses_inner` less spaghetti-code..

Chopping the spaghetti, I call it.

A critical bit that may be questionable is the new scope (smaller) that the read lock is taken and dropped in.


zancas:
@AloeareV @fluidvanadium I think one of you will need to review-and-land this, Oscar is busy elsewhere.

fluidvanadium (approved):
i really like all these comments! all these exciting non-urgent tos to do

PR #507, opened 2023-09-22
Refactor add outputs to builder

zancas:
This should merge after #506


AloeareV (approved):
Great!! This function is now starting to approach a manageable size, with all these guts split into helpers

PR #509, opened 2023-09-22
Refine tx builder helpers

zancas:
This PR, if landed after the ones I opened earlier is _just_ a refactor that properly separates conversion code out of the `send_to_addresses` code (failing before that fn is ever called in the error case).


AloeareV (approved):
Nice! Keep 'em coming!

fluidvanadium (approved):
skimmed and merged

PR #514, opened 2023-09-22
Break build out

zancas:
This PR sums up my work thus far on `refactoring send_to_addresses_inner`.   Merging it should close any preceding PRs.


fluidvanadium (approved):
basically seems like we mortared zingoconfig into builder as txbuilder

PR #516, opened 2023-09-24
Point at our rc tag

zancas:
This changes to point at our own fork of `librustzcash` and removes the patch requirement.

We'll need to update our fork to pick up features and bug-fixes from `librustzcash` but...

We:

* won't have to patch
* can easily add our own code the fork
* won't be surprised by movement in librustzcash we didn't explicitly pull in

Interestingly the `zcash_note_encryption` crate *cannot* be referenced via this mechanism...   this may be a bug in up-stream.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -766,7 +766,7 @@ dependencies = [
 [[package]]
 name = "equihash"
 version = "0.2.0"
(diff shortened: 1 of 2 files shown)


juanky201271 (approved):
nice, I'm going to do the same in zingo-mobile...

Issue #520, opened 2023-09-26
buffer


fluidvanadium:
8fbcdae2
    2
    3
    4  $cargo run -- --from "orphan diesel giggle rigid mutual inside provide meat heavy laugh grab spray color maple else weekend start small giraffe force sign tiny random maid" --birthday 1660000
    5  warning: unused import: `FromBytes`
    6    --> zingolib/src/wallet/transactions.rs:27:37
    7     |
    8  27 | ...DomainWalletExt, FromBytes, Nullifier, Rece...
    9     |                     ^^^^^^^^^
   10     |
   11     = note: `#[warn(unused_imports)]` on by default
   12
   13  warning: `zingolib` (lib) generated 1 warning
   14      Finished dev [unoptimized + debuginfo] target(s) in 0.16s
   15       Running `target/debug/zingo-cli --from 'orphan diesel giggle rigid mutual inside provide meat heavy laugh grab spray color maple else weekend start small giraffe force sign tiny random maid' --birthday 1660000`
   16  2023-09-21T16:28:20.900069Z  INFO zingo_cli: data_dir: wallets
   17  2023-09-21T16:28:21.604065Z  INFO zingo_cli:
   18  2023-09-21T16:28:21.604138Z  INFO zingo_cli: Starting Zingo-CLI
   19  2023-09-21T16:28:21.604157Z  INFO zingo_cli: Light Client config ZingoConfig { lightwalletd_uri: RwLock { data: https://mainnet.lightwalletd.com:9067/, poisoned: false, .. }, chain: Mainnet, reorg_buffer_offset: 0, monitor_mempool: true, wallet_dir: Some("wallets"), wallet_
   20  name: "zingo-wallet.dat", logfile_name: "zingo-wallet.debug.log" }
   21  2023-09-21T16:28:21.604205Z  INFO zingo_cli: Lightclient connecting to https://mainnet.lightwalletd.com:9067/
   22  [zingolib/src/lightclient.rs:1475] self.wallet.last_synced_height().await = 1660000
   23  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: UnexpectedEof, message: "failed to fill whole buffer" }', zingolib/src/wallet.rs:262:14
   24  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
    ~

PR #521, opened 2023-09-26
fix error #520. fix error #498. remove sync status. AND rename blaze to shard

fluidvanadium:
i merged some stuff that got blocked this week. this is the aggregate branch


zingolib/src/lightclient.rs
```diff
@@ -1543,7 +1545,8 @@ impl LightClient {
         let lightclient_exclusion_lock = self.sync_lock.lock().await;
 
         // The top of the wallet
-        let last_synced_height = dbg!(self.wallet.last_synced_height().await);
+        // let last_synced_height = dbg!(self.wallet.last_synced_height().await);
```
zancas, line 1548: Let's just delete the `dbg`, no reason to keep it as a comment.

zingolib/src/shard/fetch_full_transaction.rs
```diff
@@ -414,19 +413,18 @@ impl TransactionContext {
...
-            <<D as DomainWalletExt>
-              ::WalletNote as zingo_traits::ReceivedNoteAndMetadata>
-                ::Nullifier::get_nullifiers_of_unspent_notes_from_transaction_set(
-                &*self.transaction_metadata_set.read().await,
-            );
+            let unspent_nullifiers = self
+                .transaction_metadata_set
+                .read()
+                .await
+                .get_nullifier_value_txid_outputindex_of_unspent_notes::<D>();
```
zancas, line 420: Should this be:   `get_nullifier_values_and_txid_outputindexes_of_unspent_notes`?

zingolib/src/wallet/transactions.rs
```diff
@@ -306,7 +307,12 @@ impl TransactionMetadataSet {
                         {
                             Some((
                                 *txid,
-                                PoolNullifier::Sapling(sapling_note_description.nullifier),
+                                PoolNullifier::Sapling(
+                                    sapling_note_description.nullifier.unwrap_or_else(|| {
+                                        todo!("Do something about note even with missing nullifier")
```
zancas, line 312: Why might the nullifier be missing, what does that imply?

zingolib/src/wallet/transactions.rs
```diff
@@ -403,25 +402,29 @@ impl TransactionMetadataSet {
     pub fn mark_txid_nf_spent(
```
zancas, line 402: What does this `fn` name **MEAN**?

zancas (changes requested):
This looks great!   There's a lot going on.  Would it better to break it into multiple PRs?  Can we do the rename as a separate concern?  There're still a few todos! in the code.

I would like @AloeareV to look over this PR.   My preference is to break the separate concerns into separate PRs.   Overall I think these are great changes.

PR #524, opened 2023-09-26
Cargo update

zancas:


zancas:
This is just a `cargo update` it should be immediately landable once CI completes.

juanky201271 (approved):
good.

PR #527, opened 2023-09-27
Funded orchard with 3 txs mobileclient scenario

Oscar-Pepper:


juanky201271 (approved):
really nice work.

zingo-testutils/src/lib.rs
```diff
@@ -681,6 +681,69 @@ pub mod scenarios {
         )
     }
 
+    pub async fn funded_orchard_with_3_txs_mobileclient(
```
zancas, line 684: Why do we include the work `mobileclient` in this `ident`?  Is there anything about this code that's specific to a `mobileclient`?

zingo-testutils/src/lib.rs
```diff
@@ -681,6 +681,69 @@ pub mod scenarios {
...
+            .unwrap();
+        // send to a faucet
+        recipient
+            .do_send(vec![(
+                &get_base_address!(faucet, "unified"),
+                value.checked_div(10).unwrap(),
+                None,
+            )])
+            .await
+            .unwrap();
```
zancas, line 724: Why is the following `increase_height_and_sync_client` called?   Test time is important, faster tests are better.

zingo-testutils/src/lib.rs
```diff
@@ -681,6 +681,69 @@ pub mod scenarios {
...
+        // send to self sapling
+        recipient
+            .do_send(vec![(
+                &get_base_address!(recipient, "sapling"),
+                value.checked_div(10).unwrap(),
+                None,
+            )])
+            .await
+            .unwrap();
+        scenario_builder
```
zancas, line 737: I recommend to not use this method *unless* you *want* the block-chain to be ahead of the clients.   I recommend increase_height_and_sync...   instead.

zingo-testutils/src/lib.rs
Oscar-Pepper: one of your comments was a block is needed to be generated to confirm the previous send right? and the recipient needs to sync for next send

zingo-testutils/src/lib.rs
Oscar-Pepper: syncing would only increase the test time as we dont need the lightclients to be fully synced for mobile, we only care about setting up the blockchain

zingo-testutils/src/lib.rs
zancas: No.   The recipient can GOSSIP any number of sends that it knows it has funding for.

In this case the two separate receivers can be updated as the 1st and 2nd members of the `vec![` in a single `do_send`..   OR `do_send` can be called twice, and then the sync can happen after.

The critical point to understand is that there's multistep process:

1.  transaction construction (can be completely offlinne)
2.  transaction GOSSIP, zcashd nodes learn about the transaction, but NO BLOCKS HAVE CONFIRMED
3. confirmed transaction, the transaction is *in* a block

The `LightClient.do_send` method does steps 1 and 2, but (of course) it can't do step three which is up to the miners.   Clarifying this mental model is really important for `zingolib` development and use.

PR #535, opened 2023-09-28
Reorder send to addresses inner

zancas:
Finally we move the output addition early, it will not be caught up in the zip317 fee calculation loop.  (Hopefully.)

If its comfortable, this PR can simply be landed encapsulating the simpler "prereq" PRs.


zancas:
WHAT..  just happened...

PR #537, opened 2023-09-28
use earmark in value quantifying ident

zancas:


AloeareV (approved):
Make the transaction construction separate from the send....I like this.

Issue #483, opened 2023-09-13
Sapling balance incorrect - Fully synced


juanky201271:
this is before `shardtree` new feature: (you can see the command `version` doesn't exists)

![image](https://github.com/zingolabs/zingolib/assets/43755989/71814148-ec15-43c4-bb93-2621cf0ff142)

this is after `shardtree` new feature: (you can see the version: `mob-release-1.2.5_126-264-g51a2c63b`)

![image](https://github.com/zingolabs/zingolib/assets/43755989/3f82d48b-24a5-4692-b8c4-f8e6ea643091)

juanky201271:
the `550000` is incorrect, this amount was `upgrade` to `orchard` in the past.

fluidvanadium:
cargo run -- --from "future winner tumble lab salon social dream hair hollow guard combine custom cannon knock stem bracket message vacuum gold property squirrel talent quick predict" --birthday 2138530

fluidvanadium:
2023-09-28T21:33:49.343930Z  WARN zingolib::lightclient: Mempool monitor returned Err("status: Unknown, message: \"error reading a body from connection: stream error received: unexpected internal error encountered\", details: [], metadata: MetadataMap { headers: {} }"), will restart listening
2023-09-28T21:34:56.789178Z  INFO zingolib::blaze::fetch_full_transaction: Spent: utxo from 9387d08a40614b8f145c6ecc95239a4414aecb93fcd6c3468da41d7c9dcde870 was spent in 3e07122a6585fd0c9d72e6d080af2450dbfda26f20f41a4db28cae1237e8aae6

AloeareV:
> 2023-09-28T21:33:49.343930Z WARN zingolib::lightclient: Mempool monitor returned Err("status: Unknown, message: "error reading a body from connection: stream error received: unexpected internal error encountered", details: [], metadata: MetadataMap { headers: {} }"), will restart listening 

This is a fairly common message to see..it just means that there was a network hiccough and the mempool monitor lost connection.

fluidvanadium:
(main) Block:2243205 (type 'help') >> list
2023-09-28T22:07:17.911223Z ERROR zingolib::lightclient: position: 1 does not match position: 0
2023-09-28T22:07:17.911376Z ERROR zingolib::lightclient: position: 1 does not match position: 0
[
  {

Issue #484, opened 2023-09-13
txsummaries sometimes showing absurd fees after shield all cli call (probably integer underflow?)


AloeareV:
This may be related to some value still showing up in the sapling pool after being migrated to orchard via do_shield (making the balance show as incorrect)

juanky201271:
the same case here: https://github.com/zingolabs/zingolib/issues/483 really similar.

zancas:
If you have a PR for this, can you link to it?

zancas:
@AloeareV how's this bug progressing?

zancas:
What we ( @juanky201271 and I ) to know is..  is this a bug in `txsummaries`?

If yes, then we need to fix it before we can deprecate `do_list_transactions`.

juanky201271:
@zancas 
I don't see this astronomic fee for a while... I have to say that maybe this is fixed.

zancas:
Since this is evidently (at least) intermittent (if not fixed), lets move it out of the milestone.

juanky201271:
This is definitely fixed. closing...

PR #539, opened 2023-09-28
Fix transaction sandblast filter preventing trees from being updated

AloeareV:


AloeareV:
Fixes #483 and #484

zingolib/src/blaze/trial_decryptions.rs
```diff
@@ -167,26 +169,59 @@ impl TrialDecryptions {
...
+                        )
+                    })
+                    .collect();
+
                 if let Some(filter) = transaction_size_filter {
                     if compact_transaction.outputs.len() + compact_transaction.actions.len()
                         > filter as usize
                     {
-                        break;
+                        sapling_notes_to_mark_position_in_block
```
fluidvanadium, line 207: we use .extend_from_slice here... and again on line 237??

fluidvanadium (commented):
why do we add orchard_notes_to_mark_position_in_tx to orchard_notes_to_mark_position twice?
is mark position the same as add witness mark?

zingolib/src/blaze/trial_decryptions.rs
AloeareV: We call `continue` on the next line, which skips to the next iteration of the loop

fluidvanadium: aha.

fluidvanadium: so im reading if the transaction is bigger than the filter...

fluidvanadium: then we SKIP trial decryption, BUT we STILL mark_position of every note in it as ephemeral?

fluidvanadium (approved):
alright she know what it do

PR #540, opened 2023-09-28
removed extra zingocli infosudo apt install tree

fluidvanadium:
wow, i guess i shouldnt use !s in a commit message, yikes


fluidvanadium:
but perhaps they should be changed to log::debug?

AloeareV:
> wow, i guess i shouldnt use !s in a commit message, yikes

It's not git's fault, bash is catching the `!!`s before they make it to git. If you don't use `-m` and go to the interactive commit message screen, you can use as many `!`s as you want!

AloeareV (dismissed):

AloeareV (dismissed):

AloeareV:
I merged this to fix a merge conflict, and then didn't format after. @fluidvanadium could you run a cargo format and push?

zancas (dismissed):

PR #559, opened 2023-10-04
Implement zip317

zancas:
This PR finishes off the prep work.   To drop in zip317 there's one one line change on line 1260 of `wallet.rs`..  but it will require test work, and we need to think about note selection policy.


fluidvanadium:
oops, i see now i fucked it up. reverting

fluidvanadium:
i mean, rethinking

fluidvanadium:
oh right, the order the match statement has been written on has no bearing on the order the for loop executes the &policy

fluidvanadium (changes requested):
seems like the sum does not take into account the action cost of adding a small note,

this opens up an attack vector where a wallet with notes smaller than the transaction fee will get confused trying to add them to the transaction

fluidvanadium:
ah yes, note selection policy

fluidvanadium:
i thought that we tried to spent smaller notes first? but looking at select_notes_and_utxos i dont see any such sorting

fluidvanadium (approved):

PR #562, opened 2023-10-04
[#451] Detach --regtest from local zcashd environment setup

### LINKED ISSUE #451: Detach `--regtest` from local zcashd environment setup

pacu:
This breaks the darksidewalletd tests on regtest since RegtestManager kicks in when the --regtest flag is specified.

In order to be able to point to a lightwalletd server in --darkside-very-insecure mode this has to be modified so callers can tell Zingolib that they won't be needing the RegtestManager at all.

closes #451

### DIFF
--- a/zingocli/src/lib.rs
+++ b/zingocli/src/lib.rs
@@ -63,6 +63,11 @@ pub fn build_clap_app() -> clap::App<'static> {
                 .value_name("regtest")
                 .help("Regtest mode")
                 .takes_value(false))
+            .arg(Arg::new("darkside-test")
+                .long("darkside-test")
+                .value_name("darkside-test")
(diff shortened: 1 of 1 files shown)


pacu:
Closing because this is not needed:

using `--server localhost:9067 --chain regtest` has the same effect

PR #564, opened 2023-10-04
Adding a new scenario in zingo-testutils for sapling balance

juanky201271:
A new scenario for testing sapling balance added in `zingo-testutils`

### DIFF
--- a/zingo-testutils/src/lib.rs
+++ b/zingo-testutils/src/lib.rs
@@ -767,7 +767,7 @@ pub mod scenarios {
             )])
             .await
             .unwrap();
-        increase_height_and_wait_for_client(&scenario_builder.regtest_manager, &recipient, 1)
+        increase_height_and_wait_for_client(&scenario_builder.regtest_manager, &faucet, 1)
             .await
             .unwrap();
         // received from a faucet to sapling
@@ -779,7 +779,7 @@ pub mod scenarios {
             )])
             .await
             .unwrap();
-        increase_height_and_wait_for_client(&scenario_builder.regtest_manager, &recipient, 1)
+        increase_height_and_wait_for_client(&scenario_builder.regtest_manager, &faucet, 1)
             .await
             .unwrap();
         // received from a faucet to transparent
@@ -791,7 +791,7 @@ pub mod scenarios {
             )])
             .await
             .unwrap();
-        increase_height_and_wait_for_client(&scenario_builder.regtest_manager, &recipient, 1)
+        increase_height_and_wait_for_client(&scenario_builder.regtest_manager, &faucet, 1)
             .await
             .unwrap();
         // send to a faucet
@@ -855,7 +855,7 @@ pub mod scenarios {
         // end
         scenario_builder
             .regtest_manager
-            .generate_n_blocks(4)
+            .generate_n_blocks(1)
             .expect("Failed to generate blocks.");
         (
             scenario_builder.regtest_manager,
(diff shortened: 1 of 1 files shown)


zingo-testutils/src/lib.rs
```diff
@@ -737,6 +738,131 @@ pub mod scenarios {
...
+            .unwrap();
+        increase_height_and_wait_for_client(&scenario_builder.regtest_manager, &recipient, 1)
+            .await
+            .unwrap();
+        // upgrade sapling
+        recipient.do_shield(&[Pool::Sapling], None).await.unwrap();
+        // end
+        scenario_builder
+            .regtest_manager
+            .generate_n_blocks(4)
```
Oscar-Pepper, line 858: this can just generate 1 block, the 4 blocks is because i wanted to use same scenario for sync and send tests. we can update that now and create a 10block scenario with no sends for the sync test.

zingo-testutils/src/lib.rs
```diff
@@ -737,6 +738,131 @@ pub mod scenarios {
...
+            .await
+            .unwrap();
+        increase_height_and_wait_for_client(&scenario_builder.regtest_manager, &recipient, 1)
+            .await
+            .unwrap();
+        // shield transparent
+        recipient
+            .do_shield(&[Pool::Transparent], None)
+            .await
+            .unwrap();
```
Oscar-Pepper, line 849: would it be good to include some of these actions in the kotlin test itself?
so then we can test that android is shielding correctly?
maybe you intend this scenario to just test balance on its own and other tests will test shielding... its up to you

zingo-testutils/src/lib.rs
```diff
@@ -737,6 +738,131 @@ pub mod scenarios {
         )
     }
 
+    pub async fn funded_orchard_sapling_transparent_shielded_mobileclient(
+        value: u64,
```
Oscar-Pepper, line 742: i think it would be more useful to have a separate scenario called `funded_mobileclient` with three inputs, `orchard_funds`, `sapling_funds` and `transparent funds`. this can replace `funded_orchard_mobileclient` and also be nested into this scenario. I am happy to approve this PR without this work done now but maybe i the future we can refactor?

Oscar-Pepper (changes requested):

zingo-testutils/src/lib.rs
juanky201271: oh, sounds really smart, but I don't know if my rust skills are good enough to do that. But I'm happy to do this in some near future... as I said, I'm planning to create a few more tests & I think I'll do this with your help for those new tests... How sounds that?

Oscar-Pepper: great! yeah lets work together on this upgrade after this PR

zingo-testutils/src/lib.rs
```diff
@@ -737,6 +738,131 @@ pub mod scenarios {
...
+        // received from a faucet to transparent
+        faucet
+            .do_send(vec![(
+                &get_base_address!(recipient, "transparent"),
+                value.checked_div(4).unwrap(),
+                None,
+            )])
+            .await
+            .unwrap();
+        increase_height_and_wait_for_client(&scenario_builder.regtest_manager, &recipient, 1)
```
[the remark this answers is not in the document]
juanky201271: I reverted this... thanks @Oscar-Pepper

PR #568, opened 2023-10-05
Regtest network upgrade activation heights

### LINKED ISSUE #135: The real world has a blockchain where Sapling is Activated below Orchard
We need a test scenario that reflects this.  Once such a scenario has been implemented we need to test that our application functions in this space.

Oscar-Pepper:
follows from #548 to expand activation heights to all network upgrades instead of only orchard


Oscar-Pepper:
the changes in this PR require the test author to create a RegtestNetwork as a requirement for all scenarios. This regtest network has helper functions for creating a regtest network with all network upgrades active or to specify an orchard activation height whih are the two most common use cases. however, the regtest network can be configured to activate any network upgrade at any height. the regtest network is then fed into the zcash conf file construction and also to zingoconfig where it can be called with `activation height` (and similar helper functions). The motivation for creating a regtest network as an outer structure to hold ActivationHeights is to lay the foundation to further simplify zingoconfig in a future PR. i also feel it is more intuitive and is good to have all regtest network parameters in one place.

zingolib/src/lightclient.rs
```diff
@@ -2212,5 +2213,4 @@ mod tests {
             );
         });
     }
-    pub const TEST_SEED: &str = "chimney better bulb horror rebuild whisper improve intact letter giraffe brave rib appear bulk aim burst snap salt hill sad merge tennis phrase raise";
```
AloeareV, line 2215: Nice movement!

zingolib/src/blaze/trial_decryptions.rs
```diff
@@ -304,72 +304,83 @@ impl TrialDecryptions {
...
+                        return Ok::<_, String>(());
+                    };
+
+                    //TODO: Wrong. We don't have fvk import, all our keys are spending
+                    let have_spending_key = true;
+                    let uri = bsync_data.read().await.uri().clone();
+
+                    // Get network upgrade activation height
+                    let activation_height: BlockHeight = match config.chain {
+                        ChainType::Regtest => config
```
AloeareV, line 331: Could this bit go into the `ChainType::activation_height` impl?

zingoconfig/src/lib.rs
```diff
@@ -350,66 +376,198 @@ impl std::fmt::Display for ChainType {
...
-        &self,
-        nu: NetworkUpgrade,
-    ) -> Option<zcash_primitives::consensus::BlockHeight> {
+    fn activation_height(&self, nu: NetworkUpgrade) -> Option<BlockHeight> {
         match self {
             Mainnet => MAIN_NETWORK.activation_height(nu),
             Testnet => TEST_NETWORK.activation_height(nu),
-            FakeMainnet | Regtest => Some(BlockHeight::from_u32(1)),
+            // This function should never be called in regtest but fails to send in integ tests if None??
+            Regtest => Some(BlockHeight::from_u32(1)),
```
AloeareV, line 384: Could you move the `ActivationHeights` (or even the entire `RegtestNetwork`, if that doesn't break how it's currently used too much) inside of the the `ChainType::Regtest` enum variant so that we can use this method in regtest mode? transaction construction calls this function in order to determine what transaction format to create

AloeareV (changes requested):

zingoconfig/src/lib.rs
Oscar-Pepper: yeah great idea

AloeareV (approved):
Fantastic!

PR #571, opened 2023-10-06
Add dependabot.yml to keep dependencies current

AArnott:


fluidvanadium:
we used to have one of these... i know because i resolved an ancient issue opened by dependabot...
why not bring it back?

fluidvanadium (approved):
this is a great tool.

PR #575, opened 2023-10-06
put position in the right place

AloeareV:
We weren't compliant with bip32 transparent keygen for non-zero account indexes. This shouldn't break anything not belonging to @AArnott, as support for non-zero account indexes landed very recently, hasn't hit release, and I think isn't exposed anywhere other than the library.

### DIFF
--- a/zingolib/src/wallet/keys/extended_transparent.rs
+++ b/zingolib/src/wallet/keys/extended_transparent.rs
@@ -102,9 +102,9 @@ impl ExtendedPrivKey {
                 KeyIndex::hardened_from_normalize_index(config.get_coin_type()).unwrap(),
             )
             .unwrap()
-            .derive_private_key(KeyIndex::hardened_from_normalize_index(0).unwrap())
+            .derive_private_key(KeyIndex::hardened_from_normalize_index(position).unwrap())
             .unwrap()
-            .derive_private_key(KeyIndex::Normal(position))
+            .derive_private_key(KeyIndex::Normal(0))
             .unwrap()
     }
 


AArnott (approved):
Thanks.

zingolib/src/wallet/keys/extended_transparent.rs
```diff
@@ -102,9 +102,9 @@ impl ExtendedPrivKey {
                 KeyIndex::hardened_from_normalize_index(config.get_coin_type()).unwrap(),
             )
             .unwrap()
-            .derive_private_key(KeyIndex::hardened_from_normalize_index(0).unwrap())
+            .derive_private_key(KeyIndex::hardened_from_normalize_index(position).unwrap())
             .unwrap()
-            .derive_private_key(KeyIndex::Normal(position))
+            .derive_private_key(KeyIndex::Normal(0))
             .unwrap()
```
AArnott, line 108: The BIP-44 rules (which Zcash espouses for transparent addresses) prescribe a derivation path of
`m / purpose' / coin_type' / account' / change / address_index`
After this fix, this derivation gets this far:
`m / purpose' / coin_type' / account' / change`

Where does the `address_index` derivation happen? It needs to happen somewhere (even if the address_index is always 0) or the derived address won't match what other wallets derive.

AArnott: It looks like the `new_address` function in zingolib does indeed derive another level, but I'm not certain. If so, then what's here is standards-compliant to the best of my knowledge.
Further, it looks like `new_address` creates a unique transparent address every time it's called. That's cool, so long as the wallet file remembers that all these addresses have been created so that they are all individually checked for UTXOs forever after. Technically, it violates the [address gap limit rule](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#user-content-Address_gap_limit) in BIP-44 because it could generate more than 20 unique transparent addresses without any of them having actually received any funds, which makes restoring a wallet from seed likely to never discover the funds at an address beyond 20 addresses that have not been used.

AloeareV: > The BIP-44 rules (which Zcash espouses for transparent addresses) prescribe a derivation path of `m / purpose' / coin_type' / account' / change / address_index` After this fix, this derivation gets this far: `m / purpose' / coin_type' / account' / change`

Before the fix, it was incorrectly deriving `m / purpose' / coin_type' / change' / account`, which doesn't break anything for our use-case because `account` and `change` are both always zero.
> Where does the `address_index` derivation happen? It needs to happen somewhere (even if the address_index is always 0) or the derived address won't match what other wallets derive.

The `address index` derives a new level, yes.

fluidvanadium:
idk, but if AArnott approves it is good

AArnott:
I certainly do approve. The only concern I have is around the address gap limit rule, but that's unrelated to this PR and the regression it fixes.

fluidvanadium (approved):
:)

PR #581, opened 2023-10-10
Low-Hanging Cargo and Dependabot updates

fluidvanadium:


fluidvanadium:
hey, @Oscar-Pepper i messed with the zingocli/regtest/bin directory because it worked better with my flow (i have been symlinking that whole directory)

if this is not compatible with your flow, i am not attached. i can just rearrange. what do you think?

PR #585, opened 2023-10-12
Test send pre orchard funds

Oscar-Pepper:


Oscar-Pepper:
- added tests to send and shield funds since heartwood network upgrade
- added temporary mine to transparent scenario (will implement custom mine to address next PR)
- added two failing tests (ignored for now) which demonstrate failures to sync pre-sapling and shield funds mined to transparent pool
- cleared dbgs as console was cluttered
- reset faucet scenario to birthday of 0 so it is aware of all its funds and added sync so faucet is synced to tip by default

AloeareV (approved):

PR #587, opened 2023-10-12
add not-very-much-better clippy fix

zancas:


AloeareV (approved):
Eh. I agree that it's not very much better.

PR #589, opened 2023-10-12
Check if positions match before deduplicating

AloeareV:
Draft, as it still needs tests


juanky201271:
> How long until we can _completely_ excise `do_list_transactions` from the codebase?

zingo-mobile is not using it anymore.
zingo-pc is still using it, I need to migrate to summaries.
and I don't know if @AArnott is using it in his project...

zancas:
I want to amend my suggestion.    I think we should keep it around as a `test-util` only...   at least for a while.   But we should move it into a test crate.

zancas (approved):

AArnott:
I never used `do_list_transactions`. I use `lightclient.wallet.transactions()`.

PR #592, opened 2023-10-13
scenarios refactor before adding mine to orchard

Oscar-Pepper:


zingo-testutils/src/lib.rs
```diff
@@ -576,49 +585,12 @@ pub mod scenarios {
...
-            recipient,
-            txid,
-        )
+    pub async fn faucet_default() -> (RegtestManager, ChildProcessHandler, LightClient) {
+        let regtest_network = zingoconfig::RegtestNetwork::all_upgrades_active();
+        faucet(regtest_network).await
     }
 
-    pub async fn two_wallet_one_miner_fund(
+    pub async fn faucet_recipient(
```
zancas, line 593: Can we make this something like `faucet_recipient_network`?

zingo-testutils/src/lib.rs
```diff
@@ -653,7 +625,121 @@ pub mod scenarios {
         )
     }
 
-    pub async fn two_wallet_one_miner_fund_transparent(
+    pub async fn faucet_recipient_default() -> (
```
zancas, line 628: And then change this to just be `faucet_recipient`..  that way when I am writing my test I am (by default) just writing `faucet_recipient`..  and only typing extra characters in special cases where I need to specify the network upgrades...

zingo-testutils/src/lib.rs
Oscar-Pepper: as you will see in the next PR (mining to orchard) these scenarios will not be limited by parameters. regtestnetwork is just the first of many parameters we think would be useful for setting up scenarios. i think default calling a general version makes more sense than using the general name by default and adding for example `_custom` to the general version. also custom_clients would become custom_clients_custom hehe

zingo-testutils/src/lib.rs
Oscar-Pepper: see my other comment

zancas (approved):
This is great!

PR #593, opened 2023-10-13
Advanced ReOrg Test: ReOrg changes tx height

pacu:
Closes https://github.com/pacu/zwcd/issues/42 

This implements an advanced Re Org test that uses this dataset https://github.com/zingolabs/darksidewalletd-datasets/pull/13


zingocli/Cargo.toml
```diff
@@ -33,7 +33,7 @@ zcash_primitives = { workspace = true }
 tempdir = { workspace = true }
 portpicker = { workspace = true }
 env_logger = "0.10.0"
-zingolib = { path = "../zingolib/" }
+zingolib = { path = "../zingolib/", features = ["lightclient-deprecated", "darkside_tests"] }
```
zancas, line 36: If you want to catch `deprecations` more generally going forward, maybe use the `deprecations` gate?

zancas: Actually I am not sure we want this hard coded in the manifest..   would it make sense to require users to explicitly compile in deprecated functionality?  I guess the problem is that the integration_tests.rs are part of `zingocli` instead of their own crate...  maybe it's time to revisit that...

zancas (commented):
Looks pretty good so far!   I'd like @Oscar-Pepper to review it before landing.

zingocli/Cargo.toml
pacu: True.

Is there a way of handling this with more granularity? I'm not a Cargo expert so, I did what it worked but it is very possible that it's not the best solution.

fluidvanadium (approved):
yes!

PR #597, opened 2023-10-15
Remove do list transactions

zancas:
This demotes `do_list_transactions` and some other `LightClient` methods to only be compiled when the `integration-tests` feature is used.

This moves ~ 500 lines of code out of `LightClient`... which significantly reduces the space bug can live in.


fluidvanadium:
seems like the guys arent calling it

zancas:
@juanky201271 @AArnott once this PR lands `do_list_transactions` will only be available if `zingolib` is compiled with the `integration-tests` feature.

fluidvanadium (dismissed):
this is a valuable refactor and reduces ballast

zingolib/CHANGELOG.md
```diff
@@ -16,4 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Deprecated
 
+### Test Only
+- do_list_notes  --> integration_helpers.rs
+- do_list_transactions  --> integrations_helpers.rs
```
Oscar-Pepper, line 21: spelling mistake is rust filename - remove 's' from integrations

zingolib/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
+integration-tests = []
```
Oscar-Pepper, line 10: what is the benefit of having a integration test feature over using cfg(test) ? if we use cfg(test) we wouldnt need to send a feature flag when running tests

zingolib/src/lightclient.rs
```diff
@@ -163,6 +162,8 @@ pub struct LightClient {
     bsync_data: Arc<RwLock<BlazeSyncData>>,
     interrupt_sync: Arc<RwLock<bool>>,
 }
+
+/// We used to have an impl LightClient that was gated on test code, should we reintroduce that?
```
Oscar-Pepper, line 166: this PR does implement this change right? do we need the comment now?

zingolib/src/commands.rs
```diff
@@ -1379,48 +1353,6 @@ impl Command for NewAddressCommand {
     }
 }
 
-struct NotesCommand {}
```
Oscar-Pepper, line 1382: there is still code in zingo-mobile thats calls this command here:
https://github.com/zingolabs/zingo-mobile/blob/dev/app/rpc/RPC.ts#L1255
if you havent discussed with juanky might be good before this PR lands

Oscar-Pepper (changes requested):
added some comments and also failing on clippy

zingolib/src/commands.rs
juanky201271: Yes, @Oscar-Pepper is right, I thought this was for removing `list`, if you are removing `notes` as well, I need to do same changes in zingo-mobile, and I need some other command or tool instead.

AArnott:
I don't use `do_list_transactions` anyway. I call directly through `lightclient.wallet.transactions()`, which evidently remains after this change.

zingolib/Cargo.toml
zancas: Good question.   The reason we can't use this attribute is that the *kind* of test that uses these methods, is an "integration" test, that means that the code-under-test is a compiled version of the lib, built in `release` mode.

Integration tests apply scenarios to a compiled artifact produced by a different crate to the integration test crate.

The `#[cfg(test)]` attribute selectively compiles code when tests are run *within* a particular crate.

zingolib/src/lightclient.rs
zancas: Good catch!

zancas: Fixed.

zingolib/src/commands.rs
zancas: Good catch @Oscar-Pepper .

zancas:
OK, folks here's a revamped approach to `deprecation` that is applied to `do_list_transactions` as the first example.   Let me know what you think.

zingolib/src/commands.rs
```diff
@@ -1031,32 +1031,6 @@ impl Command for SeedCommand {
     }
 }
 
-struct TransactionsCommand {}
```
AloeareV, line 1034: Could this also be deprecated, instead of being fully removed? While I can use this in the integration tests (assuming that's still the plan here?) without the command interface, being able to do it in a manual test as well would be useful.

zancas: OK, I will put it back in.

zingoconfig/src/lib.rs
```diff
@@ -239,7 +239,7 @@ impl ZingoConfig {
     pub fn wallet_path_exists(&self) -> bool {
         self.get_wallet_path().exists()
     }
-    #[deprecated(since = "1.3.2", note = "this function was renamed for clarity")]
+    #[deprecated(note = "this method was renamed 'wallet_path_exists' for clarity")]
```
juanky201271, line 242: this is beautiful, thanks.

AloeareV (approved):

PR #601, opened 2023-10-18
divide tests into fast and slow

zancas:
This hack moves all integrations into a fast or slow `mod`.

The old test invocations will have the same results..  e.g.:

`cargo nextest run --features lightclient-deprecated,darkside_tests`

To run all the non slow tests:

` cargo nextest run --features lightclient-deprecated,darkside_tests --no-fail-fast -E 'not test(slow)'`

To only run the `fast` integration tests:

` cargo nextest run --features lightclient-deprecated,darkside_tests --no-fail-fast fast`

To only run the `slow` integration tests:

 `cargo nextest run --features lightclient-deprecated,darkside_tests --no-fail-fast slow`

PR #602, opened 2023-10-18
Propose fast pr ci

zancas:
This PR builds on #601 .

I think we're getting sufficient signal for merge into `dev` from this subset of the tests (provided we're gating merge from `dev` into release more strictly, with e.g. the full 'nightly' test suite).


juanky201271 (approved):
I continue believing in you

PR #603, opened 2023-10-18
added tests to reusable workflow and added all test to nightly ci

Oscar-Pepper:


juanky201271 (approved):
nice work.

zancas (approved):
Nice!

zancas:
I wonder which the partitions are taking so long?  They should be fast.

zancas (dismissed):
This is blocked on required jobs never returning statuses.    Can we move artifact construction/caching into nightly?

Oscar-Pepper:
> This is blocked on required jobs never returning statuses. Can we move artifact construction/caching into nightly?

both ci and ci-nightly call test.yaml which includes the artifact construction , caching and testing. they take cargo nextest flags as an argument so we can filter tests differently when calling the test.yaml workflow. there is nothing wrong with the PR, we just need to update branch protection. i cant do that right now as i am still in process of replacing my phone so cant authorise it

PR #607, opened 2023-10-18
Artifact cache to nightly

zancas:


Oscar-Pepper:
#603 already has no issues. it already has artifact cache in the test.yaml called by ci-nightly

PR #611, opened 2023-10-19
test that shows a big error in either do_summaries or do_save!!!!!

fluidvanadium:


zancas:
What's the status of this work?   Does the fix you recently merged have relevance?

PR #615, opened 2023-10-21
reorg expires incoming tx

### LINKED ISSUE #614: Implement a reorg that removes an incoming transaction
## ReOrg Expires Inbound Transaction
In this case the wallet receives a transaction and a reorg removes it
from the chain. This can happen if the "best chain" does not contain
that transaction, and the subsequent blocks don't include it either.

- Setup w/ default dataset with a received transaction at received_tx_height
- sync up to received_tx_height minus some blocks.
- capture balance
- sync up to received_tx_height or higher and capture txid of the received transaction
- trigger the reorg that removes the received transaction from the chain
- sync to chaintip (the wallet should recover from the reorg)7
- verify that the balance is equal to step 3 and that the txid found on step 4
- is not present anymore.

uses dataset:
https://github.com/zingolabs/darksidewalletd-datasets/pull/14

closes https://github.com/pacu/zwcd/issues/45

pacu:
[#614] Implement a reorg that removes an incoming transaction

Closes #614

In this case the wallet receives a transaction and a reorg removes it
from the chain. This can happen if the "best chain" does not contain
that transaction, and the subsequent blocks don't include it either.

1. Setup w/ default dataset with a received transaction at `received_tx_height`
2. sync up to `received_tx_height` minus some blocks.
3. capture balance
4. sync up to `received_tx_height` or higher and capture txid of the received transaction
5. trigger the reorg that removes the received transaction from the chain
6. sync to chaintip (the wallet should recover from the reorg)7
7. verify that the balance is equal to step 3 and that the txid found on step 4
is not present anymore.

uses dataset: https://github.com/zingolabs/darksidewalletd-datasets/pull/14
Closes https://github.com/pacu/zwcd/issues/45


pacu:
check instead https://github.com/zingolabs/zingolib/pull/712

PR #641, opened 2023-10-29
upgrade cli syncing diagnostic

fluidvanadium:
this improves cli experience, especially around finding transaction heights while syncking


AloeareV (approved):
Purely a change to cli sync display output...and mostly removal of outdated metrics that don't make sense anymore!

PR #647, opened 2023-10-31
simplify names now that we understand note receipt better

zancas:


zingolib/src/wallet/traits.rs
```diff
@@ -392,7 +391,7 @@ impl Nullifier for orchard::note::Nullifier {
 /// "Received" because this data is eventually chain-retrievable
 /// "Note" because this contains the Note itself
 /// "AndMetaData" everything that the client will eventually know about the note
-pub trait ReceivedNoteAndMetadata: Sized {
+pub trait NoteInterface: Sized {
```
Oscar-Pepper, line 394: the comments should also be updated on the above lines

zingolib/src/wallet/traits.rs
```diff
@@ -450,7 +449,7 @@ pub trait ReceivedNoteAndMetadata: Sized {
     fn witnessed_position_mut(&mut self) -> &mut Option<Position>;
 }
 
-impl ReceivedNoteAndMetadata for ReceivedSaplingNoteAndMetadata {
+impl NoteInterface for SaplingNote {
```
Oscar-Pepper, line 452: although a sapling note is probably technically `zcash_primitives::sapling::Note` i do agree with this name. in terms of our wallet i think it makes sense that the metadata is part of our note struct

Oscar-Pepper (dismissed):
great! just left behind some old comments that need updating

Oscar-Pepper (approved):
good!

PR #648, opened 2023-10-31
Only add outgoing metadata when empty to begin with

AloeareV:


fluidvanadium:
this is a simple elegant solution.
its a step toward 'single source of truth'
and most importantly, it turns cargo nextest into a pass

fluidvanadium (approved):

fluidvanadium:
fixes #630

PR #655, opened 2023-11-03
the save update

fluidvanadium:
consumer warning @AArnott @juanky201271 : export_save_buffer_runtime is now the only pub fn to export a LightClient save buffer

ZingoCli: save command removed.

LIghtClient: Keeps an internal save buffer in memory corresponding to the last certified complete state. it created at wallet creation time, updated at the end of each batch, and saved to disk if on an appropriate architecture

Error handling improved. ZingoLibError enum now includes strings and transmutes into other error types when necessary, thanks @AloeareV 

Test Framework: Must now be run with --features test.

### DIFF
--- a/zingolib/src/lightclient.rs
+++ b/zingolib/src/lightclient.rs
@@ -2050,3 +2050,5 @@ impl LightClient {
 }
 #[cfg(feature = "lightclient-deprecated")]
 mod deprecated;
+#[cfg(feature = "test-features")]
+mod test_features;
--- a/zingolib/src/lightclient.rs
+++ b/zingolib/src/lightclient.rs
@@ -143,6 +143,12 @@ pub struct AccountBackupInfo {
     pub account_index: u32,
 }
 
+#[derive(Default)]
+pub struct ZingoSaveBuffer {
+    buffer: Vec<u8>,
+    written: bool,
+}
+
 /// The LightClient provides a unified interface to the separate concerns that the zingolib library manages.
 ///  1. initialization of stored state
 ///      * from seed
@@ -161,6 +167,8 @@ pub struct LightClient {
 
     bsync_data: Arc<RwLock<BlazeSyncData>>,
     interrupt_sync: Arc<RwLock<bool>>,
+
+    save_buffer: ZingoSaveBuffer,
 }
 
 ///  This is the omnibus interface to the library, we are currently in the process of refining this types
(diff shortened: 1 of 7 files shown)


juanky201271 (commented):
Looks beautiful... but I'm not the guy for reviewing this... sorry.

zingolib/src/lightclient.rs
```diff
@@ -271,10 +279,75 @@ impl LightClient {
...
+
+    //        SAVE METHODS
+
+    async fn save_internal_rust(&self) -> Result<bool, ZingoLibError> {
+        self.save_internal_buffer().await?;
+        self.rust_write_save_buffer_to_file().await
+    }
+
+    async fn save_internal_buffer(&self) -> Result<(), ZingoLibError> {
+        let mut buffer: Vec<u8> = vec![];
```
fluidvanadium, line 291: this line is actually kinda weird. i dont see why we dont just write buffer  = self.wallet.write... but i cant tell right now

fluidvanadium:
passed local `cargo nextest run --features test`

AArnott:
This is great! It should go a long way to improving the consumption experience from .NET.

AArnott:
Will we be able to suppress zingolib from writing the save file to disk itself?
Will there be notifications to let the host know when new content for the file is available so we can write it within the app?

fluidvanadium:
> Will we be able to suppress zingolib from writing the save file to disk itself?
currently its only suppressed by target_os = ios or android. when i revamp the LightClient struct this is a goal of mine

> Will there be notifications to let the host know when new content for the file is available so we can write it within the app?
this is an area i have not figured out yet.
i considering using an atomic bool to mark when new content is available, but i eventually skipped the idea for expediency. if youre consuming zingolib within a rust runtime, a worker and channel solution may be appropriate

fluidvanadium:
> Will we be able to suppress zingolib from writing the save file to disk itself? Will there be notifications to let the host know when new content for the file is available so we can write it within the app?

currently its only suppressed by target_os = ios or android. when i revamp the LightClient struct this is a goal of mine
i considering using an atomic bool to mark when new content is available, but i eventually skipped the idea for expediency. if youre consuming zingolib within a rust runtime, a worker and channel solution may be appropriate

AloeareV (approved):
Reviewed via shared session

PR #658, opened 2023-11-06
added test flag. works in cargo nextest

fluidvanadium:


AloeareV (approved):

docs/integration_tests.txt
```diff
@@ -4,5 +4,5 @@
 
 cargo nextest list
 
-cargo nextest run {TESTNAME} --nocapture
+cargo nextest run --features test {TESTNAME} --nocapture
```
Oscar-Pepper, line 7: add deprecations to features list?
also i think --nocapture will clutter the console as a default?

zingolib/Cargo.toml
```diff
@@ -8,7 +8,9 @@ edition = "2021"
 
 [features]
 deprecations = ["lightclient-deprecated"]
+test = ["lightclient-deprecated", "test-features"]
```
Oscar-Pepper, line 11: or add deprecations here instead of lightclient-deprecations?

docs/integration_tests.txt
AloeareV: It was here beforehand, so beyond the scope of this PR...and I agree it clutters things up when running all tests, it's quite nice for running individual tests, though.

PR #659, opened 2023-11-06
Add compiler target indicator method

fluidvanadium:
builds on #658 

adds a .005 second sanity test

### DIFF
--- a/zingocli/tests/integration_tests.rs
+++ b/zingocli/tests/integration_tests.rs
@@ -3472,4 +3472,8 @@ mod slow {
         assert_eq!(list_before, list_after);
         assert_eq!(witness_before.unwrap(), witness_after.unwrap());
     }
+    #[tokio::test]
+    async fn os_target() {
+        dbg!(LightClient::is_mobile_target());
+    }
 }
--- a/zingocli/tests/integration_tests.rs
+++ b/zingocli/tests/integration_tests.rs
@@ -3472,8 +3472,4 @@ mod slow {
         assert_eq!(list_before, list_after);
         assert_eq!(witness_before.unwrap(), witness_after.unwrap());
     }
-    #[tokio::test]
-    async fn os_target() {
-        dbg!(LightClient::is_mobile_target());
-    }
 }
--- a/zingolib/src/lightclient/test_features.rs
+++ b/zingolib/src/lightclient/test_features.rs
@@ -12,3 +12,7 @@ impl LightClient {
         }
     }
 }
+#[tokio::test]
+async fn os_target() {
+    assert!(LightClient::is_mobile_target());
+}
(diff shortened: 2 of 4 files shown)


zingocli/tests/integration_tests.rs
```diff
@@ -3472,4 +3472,8 @@ mod slow {
         assert_eq!(list_before, list_after);
         assert_eq!(witness_before.unwrap(), witness_after.unwrap());
     }
+    #[tokio::test]
+    async fn os_target() {
+        dbg!(LightClient::is_mobile_target());
```
Oscar-Pepper, line 3477: do we need to merge this dbg into dev?

Oscar-Pepper (changes requested):

zingocli/tests/integration_tests.rs
fluidvanadium: there are many dbg!s in integration tests.

Oscar-Pepper: ok  i am of the opinion that dbgs clutter up the console and should be added for debugging and remove for merges into dev. but i suppose dbgs actually being directly inside a test is not so much of an issue for me. also what is this test actually testing, it looks like a unit test not an ingtergration test to me. could you add this to a test module in lightclient/test_features.rs as a unit test and have an assert which checks it returns false (as these tests will not run for mobile archs anyway)

Oscar-Pepper (changes requested):

zingocli/tests/integration_tests.rs
Oscar-Pepper: actually, could we architect the code so rather than creating a function which returns a boolean for if we are on mobile archs we compile gate so mobile specific code is not compiled for non-mobile and vice-versa? that would probably be safer and also make the binaries smaller

fluidvanadium: oh yeah, maybe its a unit test. i just wanted to be codamn sure that we are compiling integration tests for mobile. and this proved it to be. we dont use the is_mobile_target method anywhere else

fluidvanadium: moved to test_features

zingolib/src/lightclient/test_features.rs
```diff
@@ -0,0 +1,18 @@
...
+        }
+        #[cfg(any(target_os = "ios", target_os = "android"))]
+        {
+            false
+        }
+    }
+}
+#[tokio::test]
+async fn os_target() {
+    assert!(LightClient::is_mobile_target());
```
Oscar-Pepper, line 17: Why does this test not fail? This should evaluate as false and fail the assert?

fluidvanadium: just like you told me, we are building for a mobile target when we run the tests. so this passes

Oscar-Pepper: CI builds on a non-mobile target? i dont understand...

PR #661, opened 2023-11-07
fix 1153 block scenario hang

AloeareV:


fluidvanadium:
can confirm this created a benchmark

fluidvanadium (approved):
correct

PR #662, opened 2023-11-07
Add marking of unconfirmed_spent

### LINKED ISSUE #613: Ready for Halloween 🎃
**Transactions disappearing 👻 from history and zero balance.**

I report here on zingolib because I think it has to do with zingolib rather than Zingo Mobile or Zingo PC.

Sometimes closing and opening Zingo after the wallet syncs causes these problems:

1. **Transactions disappear**

I have compared with other wallets and yes, zingo-cli, zingo mobile and zingo-pc sometimes show a number of transactions in the list and when closing and opening again, the list of transactions changes and some of them are not displayed.

Before:

After:

In the images you can see that Zingo PC and YWalllet show two and more transactions than Zingo Mobile.

The fact that in this example the transactions do appear in Zingo PC does not mean that it does not fail there, as I have reported in the test room in Matrix, this error also happens in Zingo PC.

2. **Transaction details change**

This error has been reported before, after synchronization the transactions may show a correct amount and fee, but when you close and open the wallet again the amount and fee change to a very different, extremely large amount.

Before:

TXID: 5977961...

After:

TXID: 5877961...

3. **Balance disappears**

Just like the errors mentioned above, the wallet may show a correct balance when catching up with the synchronization, but when opening and closing the wallet and the balance usually changes, and may even be zero.

See that Zingo PC and YWallet show the correct balance, also it was so before closing and opening the application.

**I will be on the lookout for new changes for further testing.** 😎

AloeareV:
I'm hopeful this will fix #613, or at least shed some light on it!

### DIFF
--- a/zingolib/src/wallet/transactions.rs
+++ b/zingolib/src/wallet/transactions.rs
@@ -595,7 +595,7 @@ impl TransactionMetadataSet {
                     .iter_mut()
                     .find(|note| note.nullifier() == Some(spent_nullifier))
                 {
-                    *unconfirmed_spent_note.pending_spent_mut() = Some((txid, u32::from(height)))
+                    *unconfirmed_spent_note.pending_spent_mut() = Some((txid, u32::from(height)));
                 }
             }
         }
--- a/zingolib/src/wallet/transactions.rs
+++ b/zingolib/src/wallet/transactions.rs
@@ -667,7 +667,7 @@ impl TransactionMetadataSet {
                 .find(|u| u.txid == spent_txid && u.output_index == output_num as u64)
             {
                 if unconfirmed {
-                    spent_utxo.unconfirmed_spent = Some((source_txid, source_height as u32));
+                    spent_utxo.unconfirmed_spent = Some((source_txid, source_height));
                 } else {
                     // Mark this one as spent
                     spent_utxo.spent = Some(source_txid);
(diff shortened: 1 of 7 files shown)


fluidvanadium:
manual test promising for #613

fluidvanadium (commented):
should this be part of the conditional block on line 580?

fluidvanadium:
this is missing an e2e test

zingolib/src/wallet/transactions.rs
```diff
@@ -587,7 +587,17 @@ impl TransactionMetadataSet {
...
-            self.remove_witness_mark::<D>(height, txid, source_txid, output_index)
+            self.remove_witness_mark::<D>(height, txid, source_txid, output_index);
+        } else {
+            // Mark the unconfirmed_spent. Confirmed spends are already handled in update_notes
+            if let Some(transaction_spent_from) = self.current.get_mut(&source_txid) {
+                if let Some(unconfirmed_spent_note) = D::to_notes_vec_mut(transaction_spent_from)
+                    .iter_mut()
+                    .find(|note| note.nullifier() == Some(spent_nullifier))
+                {
+                    *unconfirmed_spent_note.pending_spent_mut() = Some((txid, u32::from(height)))
```
fluidvanadium, line 598: ;?

AloeareV: Semicolon would have no effect on the semantics of this...but would probably be clearer!

AloeareV:
> should this be part of the conditional block on line 580?

Which? This comment didn't link to any code

fluidvanadium (dismissed):
and a test proving a bug was fixed? AMAZING!

fluidvanadium (approved):
fixes #613

fluidvanadium:
okay, i should have tested this in cli. cli is broken and cannot see or send... i have to revert.

PR #676, opened 2023-11-13
Fix missing spaces and punctuation in error message

AArnott:
These messages were printing with "ofthis" and "sentmalicious" appearing at the console.


fluidvanadium (approved):
incrementally better!

PR #680, opened 2023-11-13
Process Spent Note error handling

fluidvanadium:
expand error handling to cover the function previously known as mark spent
and standardize search by output index rather than nullifier


AloeareV (approved):
Great!

PR #682, opened 2023-11-14
save command has been returned to dev

fluidvanadium:


juanky201271:
i can see the file is changing all the time while syncing... this mean, the `save` process is running implicit to the sync process, -> I don't need to run the command `save`... you can safely close this PR, and I'm so sorry for the Alarm.

PR #684, opened 2023-11-14
remove unused TransparentNote field

fluidvanadium:


AloeareV (approved):
Makes sense that this was never being used...as the height would be stored in the containing transaction anyway.

PR #686, opened 2023-11-14
Mv integration tests

zancas:
I move the integration tests into a new workspace member (crate) named `integration-tests`.

Some benefits:

* `cargo nextest run` in the top-level workspace directory runs 73 tests without explicitly passing a flag
This is so, because the `zingolib` dependency specified in `integration-tests` selects the `test` feature for compilation.

* integration tests for zinglib are no longer confusingly inside `zingo-cli`

NOTE:  Because this member isn't ever intended to be a standalone published crate I don't include "zingo" in its name.


zingocli/Cargo.toml
```diff
@@ -28,27 +28,6 @@ zingo-testutils = { path = "../zingo-testutils" }
...
-tracing-log = "0.1.3"
-futures-util = "0.3.28"
-orchard = { workspace = true }
-zcash_address = { workspace = true }
-zcash_client_backend = { workspace = true }
-shardtree = { workspace = true }
-serde_json = "1.0.107"
 
 [features]
 default = ["local_env"]
```
Oscar-Pepper, line 33: if you have removed local env is this needed anymore?

Oscar-Pepper (changes requested):

zingocli/Cargo.toml
zancas: Fixed.

AloeareV (approved):
Wooo!!!

PR #692, opened 2023-11-15
moved json dependency to workspace. todo: remove json in favour of se…

fluidvanadium:
…rde_json


zingolib/Cargo.toml
```diff
@@ -37,7 +37,7 @@ tower-http = { version = "0.2", features = ["add-extension"] }
 futures = "0.3.15"
 hex = "0.3"
 ring = "0.17.0"
-json = "0.12.4"
+json = { workspace = true }
```
Oscar-Pepper, line 40: why add to the workspace if this is the only crate that is using the workspace dependency?

fluidvanadium: so that zingo-status can use it

PR #696, opened 2023-11-17
Cratify darkside

zancas:
I believe that this PR is ready to land.   A minor regression is that lightwalletd must be synced from multiple locations and the error message is no longer as useful (it's incorrected for the darksidewalletd case).  I am currently researching alternative binary packaging methods to upgrade our tests, in a way that fixes both issues.  *Slow is Smooth*!!!

The relevant tracking is issue #699 .


pacu (approved):
Didn't test but looks good to me!

fluidvanadium (approved):
tested. works.

PR #701, opened 2023-11-20
Pacu advanced reorg

fluidvanadium:
all credit to @pacu . i rebased some of his test framework on our reorganized dev branch


pacu:
If you don't keep the commits I did then the following PRs will be  as hard as this one because they are based on each other

PR #712, opened 2023-11-23
Implement a reorg that removes an incoming transaction

pacu:
[https://github.com/zingolabs/zingolib/issues/614] Implement a reorg that removes an incoming transaction
Closes https://github.com/zingolabs/zingolib/issues/614

Builds atop #710

In this case the wallet receives a transaction and a reorg removes it
from the chain. This can happen if the "best chain" does not contain
that transaction, and the subsequent blocks don't include it either.

1. Setup w/ default dataset with a received transaction at `received_tx_height`
2. sync up to `received_tx_height` minus some blocks.
3. capture balance
4. sync up to `received_tx_height` or higher and capture txid of the received transaction
5. trigger the reorg that removes the received transaction from the chain
6. sync to chaintip (the wallet should recover from the reorg)7
7. verify that the balance is equal to step 3 and that the txid found on step 4
is not present anymore.

uses dataset: https://github.com/zingolabs/darksidewalletd-datasets/pull/14
Closes https://github.com/pacu/zwcd/issues/45


zancas (approved):
LGTM new darkside tests!!!!   Hell yeah!

PR #714, opened 2023-11-23
Add user balances struct and fn to LightClient

AArnott:
This is a port from the functionality originally added to Nerdbank.Zcash in [this change](https://github.com/nerdcash/Nerdbank.Cryptocurrencies/pull/150) and subsequently fixed in [this change](https://github.com/nerdcash/Nerdbank.Cryptocurrencies/pull/157).

### DIFF
--- a/zingolib/src/lightclient.rs
+++ b/zingolib/src/lightclient.rs
@@ -53,6 +53,8 @@ use zingoconfig::{ZingoConfig, MAX_REORG};
 
 static LOG_INIT: std::sync::Once = std::sync::Once::new();
 
+const MARGINAL_FEE: u64 = 5_000; // From ZIP-317
+
 #[derive(Clone, Debug, Default)]
 pub struct SyncResult {
     pub success: bool,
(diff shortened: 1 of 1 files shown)


zingolib/src/lightclient.rs
```diff
@@ -463,6 +528,123 @@ impl LightClient {
...
+                            if value > MARGINAL_FEE {
+                                useful_value += value;
+                                inbound_note_count_nodust += 1;
+                            } else {
+                                fairy_dust_value += value;
+                            }
+                        }
+                    });
+
+                tx.received_utxos
```
AArnott, line 600: `TransactionMetadata::received_utxos` is missing in this version of zingolib. What happened to it?

zingolib/src/lightclient.rs
```diff
@@ -463,6 +528,123 @@ impl LightClient {
         }
     }
 
+    pub async fn get_user_balances(&self) -> Result<UserBalances, LightWalletError> {
```
AArnott, line 531: What is your standard error enum type that I should use here?

zingolib/src/lightclient.rs
```diff
@@ -148,6 +150,69 @@ struct ZingoSaveBuffer {
...
+    /// The minimum fees that can be expected to spend all `spendable + immature_change` funds in the wallet.
+    /// This fee assumes all funds will be sent to a single note.
+    ///
+    /// Balances described by other fields in this struct are not included because they are not confirmed,
+    /// they may amount to dust, or because as `immature_income` funds they may require shielding which has a cost
+    /// and can change the amount of fees required to spend them (e.g. 3 UTXOs shielded together become only 1 note).
+    pub minimum_fees: u64,
+
+    /// The sum of non-change notes with a non-zero confirmation count that is less than the minimum required for spending,
+    /// and all UTXOs (considering that UTXOs must be shielded before spending).
```
AArnott, line 197: This policy makes perfect sense for the wallet library I'm building on zingolib, because I'm going to apply auto-shielding. But last I heard from the zingo team, zingolib wasn't going that route (nor were the zingo wallet apps). Is that still the case? If so, what do you want to do about the way this struct describes the balances?

We could potentially break out confirmed transparent income from this field to track it separately.

zingolib/src/lightclient.rs
AloeareV: Renamed to `transparent_notes`

zingolib/src/lightclient.rs
AloeareV: `error::ZingoLibError`

zingolib/src/lightclient.rs
AloeareV: Breaking out confirmed transparent balance to display separately is probably the way to go here.

AArnott: That calls into question whether `minimum_fees` includes UTXOs (which it didn't before), and makes it overall more confusing because `spendable` no longer represents all spendable funds, once you consider UTXOs to be spendable.
I'm going to investigate the option of just taking a parameter to indicate how to treat transparent funds, because given one or the other policy, I think we can fit into this struct in a consistent way, although the docs will get more complicated as it has to describe either case.

AloeareV (approved):
This looks great!

PR #717, opened 2023-11-23
Fixes arot makes sent tx disappear

zancas:
Hi @pacu ! Can we merge this instead of #715 it's just a clean rebase of that on latest dev.


pacu:
Yes, as you'd like!

PR #721, opened 2023-11-25
Add darkside spend orchard note test

Oscar-Pepper:


darkside-tests/src/utils.rs
```diff
@@ -384,3 +384,54 @@ impl TreeState {
         })
     }
 }
+
+pub async fn init_darksidewalletd() -> Result<(DarksideHandler, DarksideConnector), String> {
```
pacu, line 388: it seems that this initializes dwld with a specific state for a given set of tests. I would rename this function with a more suitable name for its use

darkside-tests/src/utils.rs
```diff
@@ -384,3 +384,54 @@ impl TreeState {
...
+    let handler = DarksideHandler::default();
+    let server_id = zingoconfig::construct_lightwalletd_uri(Some(format!(
+        "http://127.0.0.1:{}",
+        handler.grpc_port
+    )));
+    let connector = DarksideConnector(server_id);
+    let mut client = connector.get_client().await.unwrap();
+    // Setup prodedures.  Up to this point there's no communication between the client and the dswd
```
pacu, line 396:
```suggestion
    // Setup procedures.  Up to this point there's no communication between the client and the dlwd
```

darkside-tests/src/utils.rs
```diff
@@ -384,3 +384,54 @@ impl TreeState {
...
+        .stage_blocks_stream(vec![String::from(crate::constants::GENESIS_BLOCK)])
+        .await?;
+    connector
+        .add_tree_state(constants::first_tree_state())
+        .await
+        .unwrap();
+
+    Ok((handler, connector))
+}
+pub async fn stage_transaction(connector: &DarksideConnector, height: u64, hex_transaction: &str) {
```
pacu, line 416: This function stages a transaction and updates the tree state. is this something that will be done for every transaction? we probably a name that indicates that the transaction staged will also alter a tree state at height `height`. 

If this later action of updating the tree state is something that some tests will not use we probably want this to be wrapped on a flag to pass to the function. 

like `fn stage_transaction(connector: &DarksideConnector, height: u64, hex_transaction: &str, updating_tree_state: Bool)`

darkside-tests/tests/tests.rs
```diff
@@ -118,6 +51,70 @@ async fn simple_sync() {
     );
 }
 
+#[tokio::test]
+async fn spend_orchard_notes() {
```
pacu, line 55: I'd definitely add documentation on top of the test function so that the intent of the test is known beforehand

pacu (changes requested):

PR #729, opened 2023-11-27
Timed sync interrupt test

fluidvanadium:
this adds a test that uses tokio::select to drop the sync arbitrarily. it sometimes panics with a witness tree error! sometimes it fails with a sapling balance error


zancas (changes requested):
Let's #[ignore] the intermittently failing test for now, but go ahead and land this PR which expands our tool set.

fluidvanadium:
superceded by #754

PR #736, opened 2023-11-30
Darkside helpers

Oscar-Pepper:


zingolib/src/grpc_connector.rs
```diff
@@ -38,7 +38,7 @@ impl GrpcConnector {
         Self { uri }
     }
 
-    pub(crate) fn get_client(
+    pub fn get_client(
```
pacu, line 41: does this change in public interface need a CHANGELOG entry of some sort or additional docs?

darkside-tests/src/utils.rs
```diff
@@ -384,3 +388,151 @@ impl TreeState {
...
+            height: target_height as u64,
+            ..tree_state
+        })
+        .await
+        .unwrap();
+    connector.apply_staged(target_height).await.unwrap();
+    target_height
+}
+
+/// Stage a block and transaction, then update tree state.
```
pacu, line 451: is there a scenario in which updating the tree state is not desirable? If so this could be parameterized

pacu (commented):
Did a generic pass on this PR. waiting for more context

darkside-tests/src/utils.rs
zancas: Good non-blocking idea.

zancas (approved):

zancas (approved):

PR #742, opened 2023-12-02
move load_clientconfig from zingolib into zingoconfig

zancas:
The `load_clientconfig` fn is more ergonomic when located in `zingoconfig`.

This moves a `pub fn` from `zingolib` to `zingoconfig`, since it changes the public interface I'd like consumers to approve of it, before it merges.

Obviously we're incrementing the semver on this release.


AArnott (approved):

juanky201271 (approved):
nice.

PR #747, opened 2023-12-06
@fluidvanadium discovered that this wasn't used in the workspace

zancas:
This PR removes a `pub` fn.   I'd like to get feedback from @AArnott and @juanky201271 before merging it.


AArnott (approved):
It's not a fn that I use (directly anyway).

PR #749, opened 2023-12-08
updated readme to build correct features for zingo-cli

Oscar-Pepper:


fluidvanadium:
the second change i agree with

fluidvanadium (approved):
i see. this is correct. great job

PR #750, opened 2023-12-10
improved regtest error msg

fluidvanadium:
when zingocli is run in regtest mode, expect the specific error naming that ZCASHD must be placed into the correct directory


AloeareV:
Great! Tested locally, error message is comprehensible, and following the instructions makes it go away!

AloeareV (approved):

PR #751, opened 2023-12-10
added debug-only error panicker

fluidvanadium:
debug_assertions only works in debug. i have tested both sides of this by
introduce an error to lightclient
```
    pub async fn do_sync(&self, print_updates: bool) -> Result<SyncResult, String> {
        // test production
        let _: ZingoLibResult<()> = ZingoLibError::UnknownError.handle_or_panic();
```

cargo run panics
cargo run --release prints an error and keeps going

### DIFF
--- a/zingolib/src/blaze/fetch_full_transaction.rs
+++ b/zingolib/src/blaze/fetch_full_transaction.rs
@@ -491,7 +491,7 @@ impl TransactionContext {
                     }
                     Err(e) => {
                         let _memo_error: ZingoLibResult<()> =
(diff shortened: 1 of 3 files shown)


Oscar-Pepper (changes requested):
think we want it to panic for release in the case of running integration tests so im not sure this PR is doing the correct thing... thanks for introducing me to debug assertions though, nice tool to have..
 maybe a better way to structure this is instead of having a feature gate for the panic, we send the error up the stack and have integration tests call a function that panics if the result is error. however, this might take a lot of restructuring so maybe not a priority

PR #754, opened 2023-12-11
Spent process pr

fluidvanadium:
this builds upon Confenum 9 and interrupt_test

-- removes unconfirmed as a field of TransactionMetadata in favour of ConfirmationStatus

-- refactored the transactions.rs record keeping methods into a new mod. this creates a smaller problem space for any balance-related bugs to hide in, and made it easier to
-- create a unit test for get_some_txid_from_highest_wallet_block method, which is also dynamic thus no longer a field on TMS
-- spend marking uses the same functional flow regardless of status case, rather than being a concern of update notes, it is handled inside 'found_spent_nullifier'

this sets groundwork for a few immediate next steps

-- consolidate spent and unconfirmed_spent fields of notes to a txid pointer, as discussed on the airport drive


fluidvanadium:
have to address sporadic reorg failures

fluidvanadium:
> have to address sporadic reorg failures

documented and discussed

PR #761, opened 2023-12-14
No panic memo

fluidvanadium:
unreadable memo is not a bug that should break tests, so here it does not cause test to panic


Oscar-Pepper:
oh which tests do we have that reproduce this error?

PR #766, opened 2023-12-25
List tx summaries propagates error

fluidvanadium:
do list summaries currently ignores errors in fees. thats okay. this version adds a version of do_list_txsummaries_assert, a test feature that panics if any broken transactions are found.


zancas:
@fluidvanadium is this relevant?

PR #768, opened 2023-12-29
Debug clap cli

zancas:
6 out of 8 of these files were only changed to update to simpler/better seed_phrase naming.

The functional changes for this PR are in:

`zingocli/Cargo.toml` `zingocli/src/lib.rs`  and mostly `zingolib/src/wallet.rs`

Obviously since this is a change to application code, and library code **IT MUST BE REJECTED IF IT'S NOT TESTED**.

(There's a test.)

The point of the functional changes is to plug a hole in our data ingestion.  It used to be that there was _no_ validation of the data that was labeled "seed_phrase", now there is, and it's unit tested.


zancas:
I added a new feature which is env-variable management of seed-phrases.  I am thinking about a unit-test...  but maybe it just needs a manual test?

I manually tested by exporting an environment variable "SEED_PHRASE".

zancas:
The last commit allows back compatibility with users that are passing view-key Strings to `WalletBase::from_string`.

Note this view key is unsanitary.

zingocli/src/lib.rs
```diff
@@ -8,17 +8,36 @@ use log::{error, info};
...
+    let res3 = ufvk_to_string_helper(dropped_char);
+    assert!(res3.is_err())
+}
 pub fn build_clap_app() -> clap::ArgMatches {
     clap::Command::new("Zingo CLI").version(version::VERSION)
             .arg(Arg::new("nosync")
                 .help("By default, zingo-cli will sync the wallet at startup. Pass --nosync to prevent the automatic sync at startup.")
-                .long("nosync")
-                .short('n')
+                .long("nosync").short('n')
```
AloeareV, line 93: I'm surprised rustfmt likes this line...but it seems to be fine with it!

zingocli/src/lib.rs
```diff
@@ -30,31 +49,36 @@ pub fn build_clap_app() -> clap::ArgMatches {
...
-                .help("Create a new wallet with the given key. Can be a 24-word seed phrase or a viewkey. Will fail if wallet already exists"))
+                .help(r#"What chain to expect, if it's not inferrable from the server URI. One of "mainnet", "testnet", or "regtest""#))
+            .arg(Arg::new("seed-phrase")
+                .long("seed-phrase")
+                .hide(true)
+                .env("SEED_PHRASE")
+                .value_parser(WalletBase::parse_input_to_phrase)
+                .requires("birthday")
+                .help("Create a new wallet with the given seed phrase. Cannot be used with --view-key, or --wallet-dir."))
+            .arg(Arg::new("view-key")
```
AloeareV, line 123: This seems to be a reversion of https://github.com/zingolabs/zingolib/pull/426, I'm assuming this is intended?

zingocli/src/lib.rs
```diff
@@ -30,31 +49,36 @@ pub fn build_clap_app() -> clap::ArgMatches {
...
+                .env("VIEW_KEY")
+                .value_parser(ufvk_to_string_helper)
+                .requires("birthday")
+                .help("Create a new wallet with the given viewkey. Cannot be used with --seed-phrase, or --wallet-dir."))
             .arg(Arg::new("birthday")
                 .long("birthday")
                 .value_name("birthday")
                 .value_parser(clap::value_parser!(u32))
-                .help("Specify wallet birthday when restoring from seed. This is the earliest block height where the wallet has a transaction."))
+                .help("Specify wallet birthday when restoring from seed. This is the earlist block height where the wallet has a transaction."))
```
AloeareV: The only change to this line is a typo

zingocli/src/lib.rs
```diff
@@ -30,31 +49,36 @@ pub fn build_clap_app() -> clap::ArgMatches {
...
+            .arg(Arg::new("wallet-dir")
+                .long("wallet-dir")
+                .value_name("wallet-dir")
+                .help("Launch with an existing wallet specified by the value which is the absolute path to the wallet. Cannot be used with --seed-phrase, or --view-key."))
```
AloeareV: Why can't this be used with `--seed-phrase` or `--view-key`? How do you specify what directory to create the new wallet in?

AloeareV (dismissed):

zingocli/src/lib.rs
zancas: Yes.   `alias`s don't support multiple parsers, and I wanted to have a separate parser for each of the view key, and seedphrase.

I left from_string in, because it's `pub`, and I wasn't sure about consumers, but @fluidvanadium says that there probably aren't third party consumers, so we can remove it and simplify the code next time we update the semver.

I removed the `from` flag because of the Zen of Python (explicit is better than implicit).   Now the args are `--seed-phrase` and `--view-key`.    For convenience and back-compatibility they are exposed as command line options, but we *could* force the user to not use them (because the environment variable path is available).  For now the user can decide between a command line option and an environment variable.

zingocli/src/lib.rs
zancas: It's because the `wallet-dir` arg is overloaded.  If one exists it is used as the source of truth to load from..  which is mutually exclusive with passing a seed-phrase or view-key.

Really there should be an input-dir which is mutually exclusive with other inputs.

AloeareV: I think the current architecture of "if wallet-dir has wallet, load wallet (incompatible with seed-phrase/view-key), else create wallet in that location" is functional, and (if I understand the new architecture correctly) "if creating a wallet from seed/viewkey, it's not possible to specify a wallet location" isn't.

zancas: Try something like:

`cargo run -- --view-key="uview1gtxrj97cdrlf7u4c7pqa8qhsrle5y3uxd9axfrn40e6j4j6grct00xmns4tcfetg8dsl8hq5znu8pz60syrsxxxwc7m479grsfxj2nk7c0ef2k9xra5erf3nawn7qyzp585xfjqtzs4gcg53tdqpsladzezy2h9mdm75ud4wthnd9kj8fyel42x296clgr3dm4lwvduxly686z2npqhuyjcgs8adw7rpxh9f9s5tjyaj43fzxymv0fd6w868swuzwpay29xmrv3mx7htqeevehxd7e25z8ktknvwl6nmyz3j7sk4shu9nqqntskk9y4yadd8e0e65nhmd62h38ld3sxwwkmy2cv53lfngccnmlmvuhdqujdaek6hlsrzasut53pma9nenqcz6xlqag0w0wvqulyfs4as2vfmxl98k6gwr8ch659rcm0lhveyfk7mg0j3ldkqf4v55yat7938a7gj5ha530r57hpkruphj5sxnmsuqczsa0sq" --birthday=2363649 --fresh-wallet=./FOOOOOO`

zancas: OR

`cargo run -- --help`

zancas:
I set this back to Draft, the CHANGELOG must be updated before it can merge.

PR #779, opened 2024-01-12
Unit test of is incoming

zancas:
Provided that "is_change" is correct, the `is_incoming` method seems correct from inspection.   I wrote this during code review, when we thought this might in the critical path to balance calculation.

It _is_ relevant to @AArnott 's balance calculator.


fluidvanadium (approved):
unit test is good

PR #780, opened 2024-01-12
remove confusing comment

zancas:


zancas:
Hi @fluidvanadium I think this is kinda' against the wrong branch since we're working against spent_2024...  how do you propose that we handle it?  Is it already fixed on that branch?

Issue #663, opened 2023-11-08
unconfirmed is confusing and leads to double or triple negatives


Oscar-Pepper:
solution:
replace unconfirmed boolean with enum for confirmed or unconfirmed. confirmed(height) suggested by @fluidvanadium may also improve further

PR #783, opened 2024-01-15
Spent process 2024

fluidvanadium:
created integration tests that track that changing the structure of zingolib does not leave old wallet files incompatible
enumerated ConfirmationStatus with doc tests
organized and tested methods of TransactionRecord
streamlined TransactionRecord and its interface to the blockchain client


zancas:
Phewf...   big PR!   I am looking at it.

zingolib/src/wallet/transaction_record.rs
```diff
@@ -0,0 +1,343 @@
...
+
+    // List of all Utxos received in this Tx. Some of these might be change notes
+    pub transparent_notes: Vec<TransparentNote>,
+
+    // Total value of all the sapling nullifiers that were spent in this Tx
+    pub total_sapling_value_spent: u64,
+
+    // Total value of all the orchard nullifiers that were spent in this Tx
+    pub total_orchard_value_spent: u64,
+
```
AArnott: Between the names of these fields and the docs, it _looks_ like these values will always be available and reliable. But for _incoming_ transactions I think that cannot be the case. How can we set appropriate expectations there?
Maybe just change "all the (pool) nullifiers that were spent" to "all the (pool) nullifiers spent by this account".

zingolib/src/wallet/transaction_record.rs
```diff
@@ -0,0 +1,343 @@
...
+    pub status: ConfirmationStatus,
+
+    // Timestamp of Tx. Added in v4
+    pub datetime: u64,
+
+    // Txid of this transaction. It's duplicated here (It is also the Key in the HashMap that points to this
+    // WalletTx in LightWallet::txs)
+    pub txid: TxId,
+
+    // List of all nullifiers spent in this Tx. These nullifiers belong to the wallet.
```
AArnott: Clarify comment to: "...all nullifiers spent in this transaction by this account..." to account for the fact that we won't have any data for incoming transactions.

zingolib/src/wallet/transaction_record.rs
```diff
@@ -0,0 +1,343 @@
...
+    *,
+};
+
+///  Everything (SOMETHING) about a transaction
+#[derive(Debug)]
+pub struct TransactionRecord {
+    // the relationship of the transaction to the blockchain. can be either Broadcast (to mempool}, or Confirmed.
+    pub status: ConfirmationStatus,
+
+    // Timestamp of Tx. Added in v4
```
AArnott, line 18: I've always wondered how reliable this timestamp is. Is it actually from the transaction, or the block that mined it? If from the transaction, do the consensus rules guarantee that it falls within a reasonable time frame relative to the block? I think it would be good to clarify some of this in the docs.

zingolib/src/wallet/transaction_record.rs
```diff
@@ -0,0 +1,343 @@
...
+            PoolNullifier::Orchard(orchard_nullifier) => {
+                self.spent_orchard_nullifiers.push(orchard_nullifier);
+                self.total_orchard_value_spent += value;
+            }
+        }
+    }
+}
+//get
+impl TransactionRecord {
+    pub fn get_transparent_value_spent(&self) -> u64 {
```
AArnott, line 95: What purpose does this fn serve, considering the field it returns is already public? 
Less is more when it comes to public API -- particularly when the API doesn't add significant value.

zingolib/src/wallet/transaction_record.rs
```diff
@@ -0,0 +1,343 @@
...
+                self.total_value_spent(),
+                self.value_outgoing(),
+                self.total_change_returned(),
+                self,
+            ))
+            .handle()
+        }
+    }
+
+    // TODO: This is incorrect in the edge case where where we have a send-to-self with
```
AArnott, line 116: 'where where'
```suggestion
    // TODO: This is incorrect in the edge case where we have a send-to-self with
```

zingolib/src/wallet/transaction_record.rs
```diff
@@ -0,0 +1,343 @@
...
+    pub fn is_outgoing_transaction(&self) -> bool {
+        (!self.outgoing_tx_data.is_empty()) || self.total_value_spent() != 0
+    }
+    pub fn is_incoming_transaction(&self) -> bool {
+        self.sapling_notes.iter().any(|note| !note.is_change())
+            || self.orchard_notes.iter().any(|note| !note.is_change())
+            || !self.transparent_notes.is_empty()
+    }
+    pub fn net_spent(&self) -> u64 {
+        assert!(self.is_outgoing_transaction());
```
AArnott, line 127: Q: What does `assert!` do in release builds?
I worry that folks may call it for incoming transactions and get away without detection.
Maybe the function should return an `Option<u64>` or perhaps (my favorite) just return 0 for incoming transactions. But maybe that's already the net effect since I suspect the two functions involved in the expression would presumably both return 0 in this case.

zingolib/src/wallet/transaction_record.rs
```diff
@@ -0,0 +1,343 @@
...
+        self.value_spent_by_pool().iter().sum()
+    }
+
+    pub fn value_outgoing(&self) -> u64 {
+        self.outgoing_tx_data
+            .iter()
+            .fold(0, |running_total, tx_data| tx_data.value + running_total)
+    }
+
+    pub fn value_spent_by_pool(&self) -> [u64; 3] {
```
AArnott, line 171: This contract may be hard to maintain going forward. I note that 'sprout' isn't represented, so over time as pools may be deprecated in the future, how can a user of this API confidently know which pool is at which index?
IMO it would be better to return a struct where each pool is named.

zingolib/src/wallet/transaction_record.rs
fluidvanadium: i prefer a pattern where 0 means 0. total transparent_value_spent should be None until its filled in. however, this change will be a long way out.

zingolib/src/wallet/transaction_record.rs
fluidvanadium: ideally, we will implement a height to timestamp function and not store this here. thats a future upgrade

zingolib/src/wallet/transaction_record.rs
fluidvanadium: agree, it would be.

zingolib/src/wallet/transaction_record.rs
fluidvanadium: actually, this might be soon. because this value being unclear is causing inconsistencies

fluidvanadium: actually, fair warning, im about to rewrite this entire structure

zancas (dismissed):

zingolib/src/wallet/transactions/read_write.rs
```diff
@@ -0,0 +1,186 @@
...
+};
+use zcash_encoding::{Optional, Vector};
+use zcash_primitives::transaction::TxId;
+
+use crate::wallet::{data::TransactionRecord, keys::unified::WalletCapability, WitnessTrees};
+
+use super::TransactionMetadataSet;
+impl TransactionMetadataSet {
+    pub fn serialized_version() -> u64 {
+        22
```
Oscar-Pepper, line 14: should this match version of transactionRecord? i.e. 23? im not sure how we handle our versions but i want to point out they are different

zingo-testutils/src/interrupts.rs
```diff
@@ -0,0 +1,9 @@
+use zingolib::lightclient::LightClient;
+
```
zancas, line 2: Can you throw a doc-comment with a doc-testable examples clause on this?

Why is select called with `biased`?

zingo-status/src/confirmation_status.rs
```diff
@@ -0,0 +1,130 @@
...
+    pub fn from_blockheight_and_unconfirmed_bool(
+        blockheight: BlockHeight,
+        unconfirmed: bool,
+    ) -> Self {
+        if unconfirmed {
+            Self::Broadcast(blockheight)
+        } else {
+            Self::Confirmed(blockheight)
+        }
+    }
```
zancas, line 25: Hey!   Can you chuck a doc-comment with an example clause on this method?   I just ask ChatGPT to write them for me usually.  Thanks!

zingo-status/src/confirmation_status.rs
```diff
@@ -0,0 +1,130 @@
...
+    pub fn is_broadcast(&self) -> bool {
+        matches!(self, Self::Broadcast(_))
+    }
+    pub fn is_confirmed(&self) -> bool {
+        matches!(self, Self::Confirmed(_))
+    }
```
zancas, line 57: There're overlapping cases here.   Two different programmers could use this API can call two different calls that return true when the BlockHeight is "at"...   then later in life that calling code could have divergent intention.   Are these methods already called somewhere?

If they're not already explicitly used then I vote we remove them.

If they are used, then please unify the API so there's only one way to do a thing.

fluidvanadium: calling code can have different intentions: intended

zingo-status/src/confirmation_status.rs
fluidvanadium: doc tests included

zingo-testutils/src/interrupts.rs
fluidvanadium: doc tests included

integration-tests/tests/integrations.rs
```diff
@@ -2437,10 +2437,6 @@ mod slow {
             .unwrap();
 
         // 4. The transaction is not yet sent, it is just sitting in the test GRPC server, so remove it from there to make sure it doesn't get mined.
```
Oscar-Pepper: does this comment make sense?

integration-tests/tests/integrations.rs
```diff
@@ -2437,10 +2437,6 @@ mod slow {
...
-        assert_eq!(
-            do_maybe_recent_txid(&recipient).await["last_txid"],
-            sent_transaction_id
-        );
```
Oscar-Pepper, line 2443: why has this assert been removed? is this test even relevant to this PR?

integration-tests/tests/integrations.rs
```diff
@@ -3366,6 +3362,73 @@ mod slow {
...
+                .wallet
+                .transaction_context
+                .transaction_metadata_set
+                .read()
+                .await
+                .current
+        );
+
+        assert_eq!(bala_sim, bala_syn);
+    }
```
Oscar-Pepper, line 3500: interesting test, was this a failing test that was fixed by this PR? whats its purpose?

zingo-status/src/confirmation_status.rs
```diff
@@ -0,0 +1,208 @@
...
+    ///
+    /// let status = ConfirmationStatus::Broadcast(10.into());
+    /// assert_eq!(status.is_confirmed_after_or_at(&10.into()), false);
+    ///
+    /// let status = ConfirmationStatus::Confirmed(10.into());
+    /// assert_eq!(status.is_confirmed_after_or_at(&12.into()), false);
+    /// ```
+    pub fn is_confirmed_after_or_at(&self, comparison_height: &BlockHeight) -> bool {
+        match self {
+            Self::Confirmed(self_height) => self_height >= comparison_height,
```
Oscar-Pepper, line 76: self_height doesnt make sense to me, can these just be height? or can we make it clearer what the self part is for?

zingo-status/src/confirmation_status.rs
```diff
@@ -0,0 +1,208 @@
...
+    /// The transaction has been included in at-least one block mined to the zcash blockchain.
+    /// The height of a confirmed block that contains the transaction.
+    Confirmed(BlockHeight),
+}
+
+impl ConfirmationStatus {
+    /// Converts from a blockheight and `unconfirmed`. unconfirmed is deprecated and is only needed in loading from save.
+    pub fn from_blockheight_and_unconfirmed_bool(
+        blockheight: BlockHeight,
+        unconfirmed: bool,
```
Oscar-Pepper, line 18: i thought this PR was erradicating unconfirmed bools? i think if we are not erradicating unconfirmed bools then this must be accompanied by an issue to do so soon

Oscar-Pepper: i guess the comment mentioned its deprecatedness

zingo-status/src/confirmation_status.rs
```diff
@@ -0,0 +1,208 @@
...
+        match self {
+            Self::Broadcast(self_height) => *self_height,
+            Self::Confirmed(self_height) => *self_height,
+        }
+    }
+}
+
+impl std::fmt::Display for ConfirmationStatus {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        use ConfirmationStatus::*;
```
Oscar-Pepper: recent standup people were not happy with `*` in `use` statements right?￼

Oscar-Pepper: maybe this is an exception or this isnt the same case

zingolib/src/blaze/fetch_full_transaction.rs
```diff
@@ -417,20 +398,22 @@ impl TransactionContext {
...
-                        transaction.txid(),
-                        transaction_block_height,
-                        true, // this was "unconfirmed" but this fn is invoked inside `if unconfirmed` TODO: add regression test to protect against movement
-                        block_time,
-                        (*nf).into(),
-                        *value,
-                        *transaction_id,
-                        *output_index,
-                    );
+                    let _ = self
```
Oscar-Pepper, line 405: what is the `let _ =` for??

zingolib/src/blaze/fetch_full_transaction.rs
```diff
@@ -417,20 +398,22 @@ impl TransactionContext {
...
+                        .transaction_metadata_set
+                        .write()
+                        .await
+                        .found_spent_nullifier(
```
Oscar-Pepper, line 409: what is a spent nullifier?

Oscar-Pepper: it seems to match something in original code so i gues we can keep consistent

zingolib/src/blaze/update_notes.rs
```diff
@@ -123,32 +123,20 @@ impl UpdateNotes {
...
-                                &transaction_id_spent_in,
-                                spent_at_height,
-                                output_index,
-                            )
-                            .unwrap_or(0);
-
                         // Record the future transaction, the one that has spent the nullifiers received in this transaction in the wallet
-                        wallet_transactions_write_unlocked.add_new_spent(
+                        let status = ConfirmationStatus::Confirmed(spent_at_height);
+                        let _ = wallet_transactions_write_unlocked.found_spent_nullifier(
```
Oscar-Pepper, line 135: what is `let _ =` for?

zingolib/src/wallet/notes.rs
```diff
@@ -0,0 +1,491 @@
...
+
+    fn witnessed_position_mut(&mut self) -> &mut Option<Position> {
+        &mut self.witnessed_position
+    }
+
+    fn output_index(&self) -> &Option<u32> {
+        &self.output_index
+    }
+}
+
```
Oscar-Pepper, line 350: outputindex_mut isnt used anywhere? nice find!

Oscar-Pepper (dismissed):
incredible PR, this kind of work is exactly what we need to make our code easily managle, readable and more stable. great job!

i have some questions / minor changes so i am requesting changes but mostly looks ready to go

zingolib/src/wallet/transactions/read_write.rs
fluidvanadium: in this PR i kept the save files compatible... but it should probably be tested to prove it..

fluidvanadium: https://github.com/zingolabs/zingolib/pull/795

integration-tests/tests/integrations.rs
fluidvanadium: the assert was unnecessary and relied on some_txid_from_highest_block which was a misnomer

integration-tests/tests/integrations.rs
fluidvanadium: this test is new. i found it was failing interestingly on rare occasion. so i included it

zingo-status/src/confirmation_status.rs
fluidvanadium: it helped me write the binary comparators to be explicitly clear which height was self and which was the comparison.

zingo-status/src/confirmation_status.rs
fluidvanadium: fixed

zingolib/src/blaze/fetch_full_transaction.rs
fluidvanadium: this is an error case that is not handled. error handling is not part of this PR but this is upgraded in https://github.com/zingolabs/zingolib/pull/786

zingolib/src/blaze/update_notes.rs
fluidvanadium: this is an error case that is not handled. error handling is not part of this PR but this is upgraded in https://github.com/zingolabs/zingolib/pull/786

zingolib/src/blaze/fetch_full_transaction.rs
fluidvanadium: a nullifier that comes up as an input to a transaction.

zingolib/src/wallet/notes.rs
fluidvanadium: credit to @AloeareV

integration-tests/tests/integrations.rs
fluidvanadium: no. its gone

zingolib/src/blaze/fetch_full_transaction.rs
Oscar-Pepper: yes my guess would be that it is a nullifier in the tx that matches a nulliifer calulated from our notes in the wallet

fluidvanadium:
i dreamnt up a way to improve the wallet load tests

zancas (approved):

fluidvanadium:
fixes #663

PR #786, opened 2024-01-18
Error handle mark spent

fluidvanadium:
if mark_note_as_spent throws an error, sync will restart from the batch

a branch of spent_process_2024


fluidvanadium:
this is not necessary for release

fluidvanadium:
post release, this should go into dev

Oscar-Pepper (approved):
there are still a few unwraps in this function that would be good to remove from the code but this is an improvement for sure

PR #802, opened 2024-01-31
zip317 using zcash_client_backend traits

AloeareV:
So far this has implemented the first of the required InputSource methods.


zancas:
@AloeareV is this PR still useful?  Or should we close it?

PR #807, opened 2024-02-06
tag the build with a commit hash, even in the case where there's no t…

zancas:
…ag, and the case where the tag points at the tip

This change makes it be the case that an explicit commit hash is *always* included in the build version information.

The result is that users of the app can see a particular commit hash, for the zingolib version that was used in the build.


fluidvanadium (approved):
i looked up what these git flags do and they seem to increase consistency

PR #809, opened 2024-02-08
Note interface overhaul part 1 trait

fluidvanadium:
this adds a noteinterface trait and begins the process of bringing transparent notes in line with pool notes


zingolib/src/wallet.rs
Oscar-Pepper, line 1: the get_target_height_and_anchor_offset trait method should be in a different PR. this PR says ready to review but this seems to be unfinished owrk and not to do with note interface

Oscar-Pepper (changes requested):

Oscar-Pepper (approved):

PR #813, opened 2024-02-15
Zingoledger traits

fluidvanadium:


zingolib/src/wallet/notes/interface.rs
```diff
@@ -0,0 +1,70 @@
...
+    keys::unified::WalletCapability,
+    traits::{FromBytes, FromCommitment, Nullifier, ReadableWriteable, ToBytes},
+    Pool,
+};
+
+pub trait NoteInterface: Sized {
+    fn spent(&self) -> &Option<(TxId, u32)>;
+    fn is_spent(&self) -> bool {
+        Self::spent(self).is_some()
+    }
```
Oscar-Pepper, line 15: it makes logical sense to me to have is_spent as the first or last fn so spent and spent_mut etc. are grouped together. just an organisation thing

zingolib/src/wallet/notes/interface.rs
```diff
@@ -0,0 +1,70 @@
...
+    fn spent_mut(&mut self) -> &mut Option<(TxId, u32)>;
+    fn pending_spent(&self) -> &Option<(TxId, u32)>;
+    fn pending_spent_mut(&mut self) -> &mut Option<(TxId, u32)>;
+}
+
+///   All zingolib::wallet::traits::Notes are NoteInterface
+///   NoteInterface provides...
```
Oscar-Pepper: this doc comment is for NoteInterface not ShieldedNoteInterface.... also its not very helpful haha

i think everything in wallet::traits is for shielded right? or maybe the from bytes/ to bytes might be applicable to transprent im not sure

Oscar-Pepper: i see this comment was made before you created NoteInterface... interesting..

maybe change to ShieldedNoteInterface?

zingolib/src/wallet/notes/transparent.rs
```diff
@@ -0,0 +1,167 @@
...
+            address,
+            txid: transaction_id,
+            output_index,
+            script,
+            value,
+            spent: spent_tuple,
+            unconfirmed_spent: None,
+        })
+    }
+}
```
Oscar-Pepper, line 167: not sure why read and write were swapped in order but not an issue, just harder to review and dont see the benefit

zingolib/src/wallet/notes/transparent.rs
```diff
@@ -0,0 +1,167 @@
...
+    fn pending_spent(&self) -> &Option<(TxId, u32)> {
+        &self.unconfirmed_spent
+    }
+    fn pending_spent_mut(&mut self) -> &mut Option<(TxId, u32)> {
+        &mut self.unconfirmed_spent
+    }
+}
+
+impl TransparentNote {
+    pub fn new(
```
Oscar-Pepper: can we change this to `from_parts` to stay in line with sapling/orchard and also the LRZ format?

Oscar-Pepper (changes requested):
this is really great work, I have a  few questions/requests.
i noticed after running a global search on `transaction_meta` that there are many places where transaction_metadata is still used instead of transaction_record and also transaction_metadata_set is used instead of zingo_ledger. could you either fix this or put an issue out with the merging of this PR so its pulled in line soon?
also there is a mistake in wallet/traits.rs:399 where is says `transaction_metadata_set` but its a `transaction_metadata`

zingolib/src/wallet/notes/transparent.rs
zancas: If this was the result of an automation we should configure the tool to enforce the preferred order.

zingolib/src/wallet/notes/interface.rs
fluidvanadium: fixed in https://github.com/zingolabs/zingolib/pull/809

zingolib/src/wallet/notes/transparent.rs
fluidvanadium: sorry for any confusion. i feel strongly that write comes before read in a flow

PR #817, opened 2024-02-25
Initial integration with ledger application

neithanmo:
This draft PR introduces a foundational step towards secure hardware wallet support within zingolib by integrating a custom ledger application. 
### Changes
- **Keystore Abstraction:** Introduced a new abstraction layer, Keystore, as an enum with two variants: WalletCapability for the existing in-memory wallet functionality, and LedgerWalletCapability for the new ledger integration. This allows for seamless switching between wallet types and makes future extensions possible and easy.

- **Ledger Wallet Capability:** The LedgerWalletCapability encapsulates all interactions with the Ledger device, including transaction signing(not supported yet) and address generation.

- **Support for Sapling and Transparent Addresses:** The ledger application currently supports sapling and transparent addresses, orchard protocol is not supported yet by the ledger-app.

### Challenges
- **Integration with In-memory Key Architecture:** Adapting the ledger's secure key management approach to fit within zingolib's architecture, which currently relies on in-memory keys, poses a significant challenge. This required careful design to ensure that cryptographic operations are executed on the Ledger device, with zingolib managing only the operations that do not involve direct access to private keys, usually done through the TryFrom trait.
- **Partial Address Support:** Currently, the integration does not support orchard and unified addresses. The support for Orchard is not part of the scope of implementing this feature in zingolib but in the ledger-app.
- **Data storage:** Currently we are unable to read address data from wallet.dat file which is used to store important metadata like addresses and key parameters. This is due to the way the ledger app operates.
-  ** Understanding zingolib architecture:** The current integration effort can be improved by a better understanding of the zingo architecture to ensure a seamlessly integration that would facilitate further integration of new features.
- **Types differences:** There are some conflicts between our Rust library and zingolib, the differences are more related to versioning and API compatibility.

There are some TODOs in the code and comments explaining the reason.
This initial integration is still not functional, in order to get there more changes have to be done in:
- Defining how many metadata file the app supports and if differentiating from in memory or ledger-app ones is desirable.
- Proper changes on the UI
- Add support for sapling addresses
- Transaction signing would possibly require adding another abstraction, which for the ledger app is provided via ledger-zcash-rs
- Modifying ledger-zcash-rs library, in order to use raw data, could make easy further integration efforts, decoupling other libraries and relying on simple conversion, for example:  `PublicKey::from_slice()`, and so on.

### DIFF
--- a/zingolib/src/wallet.rs
+++ b/zingolib/src/wallet.rs
@@ -1011,7 +1011,6 @@ impl LightWallet {
             self.transaction_context.config.chain,
             submission_height,
             transaction::builder::BuildConfig::Standard {
-                // TODO: We probably need this
                 sapling_anchor: Some(sapling_anchor),
                 orchard_anchor: Some(orchard_anchor),
             },
--- a/zingolib/src/wallet.rs
+++ b/zingolib/src/wallet.rs
@@ -1,8 +1,10 @@
 //! In all cases in this file "external_version" refers to a serialization version that is interpreted
 //! from a source outside of the code-base e.g. a wallet-file.
 use crate::blaze::fetch_full_transaction::TransactionContext;
-use crate::wallet::data::{SpendableSaplingNote, TransactionRecord};
-use crate::wallet::notes::ShieldedNoteInterface;
+use crate::wallet::{
+    data::{SpendableSaplingNote, TransactionRecord},
+    notes::ShieldedNoteInterface,
+};
 
 use bip0039::Mnemonic;
 use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
@@ -51,7 +53,8 @@ use zingo_memo::create_wallet_internal_memo_version_0;
 use zingo_status::confirmation_status::ConfirmationStatus;
 
 use self::data::{SpendableOrchardNote, WitnessTrees, COMMITMENT_TREE_LEVELS, MAX_SHARD_LEVEL};
-use self::keys::unified::{Capability, WalletCapability};
+use self::keys::keystore::Keystore;
+use self::keys::unified::Capability;
 use self::traits::Recipient;
 use self::traits::{DomainWalletExt, SpendableNote};
 use self::utils::get_price;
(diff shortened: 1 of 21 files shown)


zingolib/src/wallet.rs
```diff
@@ -199,13 +202,17 @@ pub enum WalletBase {
...
 impl WalletBase {
     pub fn from_string(base: String) -> WalletBase {
-        if (&base[0..5]) == "uview" {
-            WalletBase::Ufvk(base)
-        } else {
-            WalletBase::MnemonicPhrase(base)
+        match base {
+            _ if &base[0..5] == "uview" => WalletBase::Ufvk(base.clone()),
+            #[cfg(feature = "ledger-support")]
+            _ if &base[0..5] == "ledger" => WalletBase::Ledger,
```
Oscar-Pepper, line 214: 0..6 (range is exclusive of end index)

AloeareV: ..= can be used to create an inclusive range, if 0..=5 would be clearer, also, the strings being matched are 5 and 6 characters respectively, so the ranges should match those lengths...

Actually now that I think about it, I think this will panic if base is shorter than the range, `starts_with` or a similar method would be more robust

fluidvanadium:
This is on my radar.

zingolib/src/wallet/keys/ledger.rs
```diff
@@ -0,0 +1,802 @@
...
+
+    pub const fn t_derivation_path(coin_type: u32, index: u32) -> [ChildIndex; 5] {
+        [
+            ChildIndex::hardened(44),
+            ChildIndex::hardened(coin_type),
+            ChildIndex::hardened(0),
+            // TODO: Bellow child_index in our lib are meant to be normal
+            // not hardened, but seems zingolib fork of this library
+            // uses only hardened fields in path????
+            ChildIndex::hardened(0),
```
AloeareV: This isn't the case, see https://github.com/zingolabs/zingolib/blob/1be3f866fceb12be4e06b04371d783e5bf85e777/zingolib/src/wallet/keys/extended_transparent.rs#L98-L108

The `zcash_primitives::ChildIndex` type being used here only supports hardened indexes because it's a sapling-specific type, and sapling doesn't need non-hardened indexes. We have our own KeyIndex type...that is a holdover from zecwallet-lite, and I think we can use `hdwallet::KeyIndex` in its place, which is the type used by librustzcash
for its transparent derivation

zingolib/src/wallet/keys/ledger.rs
```diff
@@ -0,0 +1,802 @@
...
+
+        let path = taddrs
+            .keys()
+            .last()
+            .cloned()
+            .map(|path| {
+                [
+                    // TODO: last two fields should not be hardened
+                    // but this is what the zcash-primitives provides
+                    ChildIndex::hardened(path[0]),
```
AloeareV: See above comment about KeyIndex vs ChildIndex

AloeareV (commented):
Leaving some notes here, most likely for myself

zancas:
We've consulted with ZonDax and this work is largely obsolete.  Is there important code to preserve @neithanmo ?

pacu:
Is this fully outdated or is this just hard to rebase?

neithanmo:
@zancas it is outdated, but the general integration idea holds, however this would depend on recent development the structure/design might be now outdated and could not be the best integration approach

pacu:
I'm giving it a go to see if I can bring this to life in a newer branch

zancas:
Superseded by #1352

PR #823, opened 2024-03-07
replaced missing transparent viewkey panic with handled error

fluidvanadium:
Should prevent rust panics. However, this may not prevent human panics.


zancas (approved):
Nice hack.

zancas:
I am going to test locally.

zancas:
I notice that this PR is into `dev`, but isn't this a hot-fix?   That should go into stable?

zancas (changes requested):
Let's add a test that covers this code.

PR #825, opened 2024-03-07
Hotfix transparent viewkey missing

fluidvanadium:
includes test that covers

### DIFF
--- a/integration-tests/tests/integrations.rs
+++ b/integration-tests/tests/integrations.rs
@@ -124,6 +124,9 @@ fn check_view_capability_bounds(
 }
 
 mod fast {
+    use zcash_address::unified::Encoding;
+    use zingolib::wallet::WalletBase;
+
     use super::*;
     #[tokio::test]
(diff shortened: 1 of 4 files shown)


fluidvanadium:
ignore the clippy here. its updated in dev but not worth hotfixing

juanky201271 (commented):
My concern about this is:
- A user can restore from vk, and after that when the user wants to check the vk, the App shows up an error message?
- `Restore from ufvk` & `exportufvk` would have to be synchronize, right? both with an error message or not, something coherent.

fluidvanadium:
right o. the new version should export the transparentless viewkey

fluidvanadium:
addresses the panic on line 179 of unified.rs

Oscar-Pepper (dismissed):
great job

Oscar-Pepper (approved):

juanky201271:
> right o. the new version should export the transparentless viewkey

Thanks, I appreciated that.

PR #834, opened 2024-03-11
stabilize rust toolchain

fluidvanadium:
The current stable version is already built this way. By locking it in, it becomes more stable.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
 channel = "1.75.0"
-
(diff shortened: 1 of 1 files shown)


rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "1.75.0"
```
zancas: Why not 1.76?

zancas (approved):

zancas:
I am running the tests locally.

PR #842, opened 2024-03-12
created a wallet::send module and sorted send functionality into it

fluidvanadium:
This is a nonfunctional change that should be merged if people want to build on it while i am.


zancas (dismissed):
utACK

PR #851, opened 2024-03-19
implement a new function necessary to transmute the Receivers into an…

fluidvanadium:
… LRZ-compatible TransactionRequest


zingolib/src/wallet/send.rs
```diff
@@ -72,6 +73,26 @@ impl SendProgress {
 }
 
 type Receivers = Vec<(address::Address, NonNegativeAmount, Option<MemoBytes>)>;
+
+// review! unit test this
```
fluidvanadium, line 77: this should be tested

AloeareV (approved):

PR #857, opened 2024-03-30
Unify zcash conf

zancas:
@Oscar-Pepper noticed multiple sources of truth for the configuration of "regtest" zcashd in interactive command line mode and in automated tests.

This PR properly unifies the source of truth.
unify_conf


rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.77.0"
+channel = "1.77"
```
Oscar-Pepper: this has now been superceeded by "stable" in your recent PR

zingo-testutils/src/regtest.rs
```diff
@@ -175,11 +175,22 @@ pub fn launch_lightwalletd(
...
 }
+fn write_zcash_conf(location: &PathBuf) {
+    // This is the only data we need to supply *to* the zcashd, the other files are created by zcashd and lightwalletd
+    use std::io::Write;
+    let conf_bytes: &'static [u8] = include_bytes!("../regtest/conf/zcash.conf");
+    File::create(location)
+        .unwrap()
+        .write_all(conf_bytes)
+        .unwrap();
+}
```
Oscar-Pepper, line 186: this is definately an improvement and i'm not requesting changes here, but it would be nice for zingo-cli to read all the the regtest files from zingo-testutils instead of copying them eventually

Oscar-Pepper (changes requested):

rust-toolchain.toml
zancas: OK.. but that PR hasn't landed yet.
When it does I will merge it into this branch.

zingo-testutils/src/regtest.rs
zancas: Since `zingo-cli` is a different crate that just depends on `zingo-testutils` it doesn't have access to data files that are part of `zingo-testutils`.

Probably a more elegant solution is to represent file contents as `const FILECONTENTS: &str = ..` inside of `zingo-testutils`.

zancas: I vote we merge this as is.

Oscar-Pepper: Sure. As I said im not requesting changes here.

rust-toolchain.toml
Oscar-Pepper: ok great

Oscar-Pepper (changes requested):
Great. Will approve once changes to `rust-toolchain.toml` are removed

Oscar-Pepper (approved):

PR #858, opened 2024-03-31
Fix is confirmed before bug

zancas:
This PR is mostly a cherry-pick of a commit from the spend_kit branch (see commit message).

I will break up my rationale for this PR into two parts.

## How I Fixed The Bug

  The bug was caused by a tiny smidge of wetness, and copy pasta, coupled with the false belief that the doc tests were protecting us.
  The fix is both correct, and DRYer because it uses composition of relevant functionality rather re-expression of the same thing.

(I also tightened boundary conditions and added assertions to the doc-tests.)
 
## Why I Propose This Way of Organizing PRs

   If this PR is accepted into `dev` it will have some nice properties:

(1)  It will have made the length of the feature branch (as measured in diff from dev) smaller.  I assume this is obviously good.
(2)  It will have decreased the length of distance from the tip of the feature, to the tip of any other feature.
     Here's the thought experiment:
        Oscar is working on some other feature (some sync thing) at the same time.  This PR lands (maybe Oscar reviews it).
   
   Now what?

   There's this chunk of code (on dev) that doesn't really do anything (on dev), *but* it's *probably* necessary for spend_kit (its "origin feature") and some_sync_thing is *aware* of it.   This means that (a) some_sync_thing won't accidentally put something else there.   Maybe some_sync_thing will actually use it.   I dunno'..  but at the end of the day all the `dev`eloping features need to work together.   By sucking pieces of them into `dev`, when it is convenient to do so, we ensure that features being developed against `dev` are aware of each other earlier.

### DIFF
--- a/zingo-status/src/confirmation_status.rs
+++ b/zingo-status/src/confirmation_status.rs
@@ -32,9 +32,11 @@ impl ConfirmationStatus {
     ///
     /// let status = ConfirmationStatus::Broadcast(10.into());
     /// assert_eq!(status.is_broadcast(), true);
+    /// assert_eq!(status.is_confirmed(), false);
     ///
     /// let status = ConfirmationStatus::Confirmed(10.into());
     /// assert_eq!(status.is_broadcast(), false);
+    /// assert_eq!(status.is_confirmed(), true);
     /// ```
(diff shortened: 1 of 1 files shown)


fluidvanadium (approved):
utACK

PR #859, opened 2024-03-31
trigger doctests

zancas:
(and other internal tests picked up by cargo test) in zingo-status/src directory.


fluidvanadium (approved):
utACK

PR #867, opened 2024-04-01
Removed unused struct

Oscar-Pepper:
This is already removed from dev but breaking CI for stable also


juanky201271 (dismissed):
good.

juanky201271 (approved):

.github/workflows/test.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
 
       - uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
-          toolchain: stable
```
zancas, line 25: stable is probably correct.

PR #870, opened 2024-04-01
set toolchains to stable

zancas:


zingocli/src/lib.rs
```diff
@@ -263,6 +263,7 @@ fn short_circuit_on_help(params: Vec<String>) {
     std::process::exit(0x0100);
 }
 use std::string::String;
+#[allow(dead_code)]
```
Oscar-Pepper: I think we need to get to the bottom of why this warning appears when these error variants are used in the code and the warnings only appear in CI

Oscar-Pepper (changes requested):

zingocli/src/lib.rs
zancas: Fixed.

Oscar-Pepper (approved):
We should improve the error handling here in the near future

PR #893, opened 2024-04-05
send command refactor and improved error handling

Oscar-Pepper:
- Moves parsing to dedicated command as prereq for zip317 propose command
- Cleans up parsing logic and send command
- Improves error handling
- Adds utils and error sub-mods to commands.rs
- Fixes bug in send help for sending using a json string
- Fixes bug where an incorrect address triggers the wrong error when there is some memo


fluidvanadium (approved):
its easier to understand

PR #899, opened 2024-04-06
Extend parse send args tests

zancas:
Here are more tests for the parse_send_args helper utility.

There are a few improvements:

* more error variants
* trim against from-single-arg int parsing

### DIFF
--- a/zingolib/src/commands/error.rs
+++ b/zingolib/src/commands/error.rs
@@ -28,7 +28,7 @@ impl fmt::Display for CommandError {
             }
             InvalidMemo(e) => write!(f, "failed to interpret memo. {}", e),
             NonJsonNumberForAmount(e) => write!(f, "Non Number input: {}", e),
-            SingleArgNotJsonArray(e) => write!(f, "{}", e),
+            SingleArgNotJsonArray(e) => write!(f, "argument cannot be parsed to a json array: {}", e),
         }
     }
 }
--- a/zingolib/src/commands/error.rs
+++ b/zingolib/src/commands/error.rs
@@ -2,8 +2,8 @@ use std::fmt;
 
 #[derive(Debug)]
 pub(crate) enum CommandError {
-    FailedJsonParsing(json::Error),
-    FailedIntParsing(std::num::ParseIntError),
+    ArgsNotJson(json::Error),
+    ParseIntFromString(std::num::ParseIntError),
     UnexpectedType(String),
     MissingKey(String),
     InvalidArguments,
@@ -16,8 +16,8 @@ impl fmt::Display for CommandError {
         use CommandError::*;
 
         match self {
-            FailedJsonParsing(e) => write!(f, "failed to parse argument. {}", e),
-            FailedIntParsing(e) => write!(f, "failed to parse argument. {}", e),
+            ArgsNotJson(e) => write!(f, "failed to parse argument. {}", e),
+            ParseIntFromString(e) => write!(f, "failed to parse argument. {}", e),
             UnexpectedType(e) => write!(f, "arguments cannot be parsed to expected type. {}", e),
             MissingKey(key) => write!(f, "json array is missing \"{}\" key.", key),
             InvalidArguments => write!(f, "arguments given are invalid."),
--- a/zingolib/src/commands/error.rs
+++ b/zingolib/src/commands/error.rs
@@ -28,7 +28,9 @@ impl fmt::Display for CommandError {
             }
             InvalidMemo(e) => write!(f, "failed to interpret memo. {}", e),
             NonJsonNumberForAmount(e) => write!(f, "Non Number input: {}", e),
-            SingleArgNotJsonArray(e) => write!(f, "argument cannot be parsed to a json array: {}", e),
+            SingleArgNotJsonArray(e) => {
+                write!(f, "argument cannot be parsed to a json array: {}", e)
+            }
         }
     }
 }
--- a/zingolib/src/commands/error.rs
+++ b/zingolib/src/commands/error.rs
@@ -3,6 +3,7 @@ use std::fmt;
 #[derive(Debug)]
 pub(crate) enum CommandError {
     ArgsNotJson(json::Error),
+    SingleArgNotJsonArray(String),
     ParseIntFromString(std::num::ParseIntError),
     UnexpectedType(String),
     MissingKey(String),
@@ -27,6 +28,7 @@ impl fmt::Display for CommandError {
             }
             InvalidMemo(e) => write!(f, "failed to interpret memo. {}", e),
             NonJsonNumberForAmount(e) => write!(f, "{}", e),
+            SingleArgNotJsonArray(e) => write!(f, "{}", e),
         }
     }
 }
--- a/zingolib/src/commands/error.rs
+++ b/zingolib/src/commands/error.rs
@@ -27,7 +27,7 @@ impl fmt::Display for CommandError {
                 write!(f, "memo's cannot be sent to transparent addresses.")
             }
             InvalidMemo(e) => write!(f, "failed to interpret memo. {}", e),
-            NonJsonNumberForAmount(e) => write!(f, "{}", e),
+            NonJsonNumberForAmount => write!(f,  "amount is not a json::number::Number"),
             SingleArgNotJsonArray(e) => write!(f, "{}", e),
         }
     }
--- a/zingolib/src/commands/error.rs
+++ b/zingolib/src/commands/error.rs
@@ -9,6 +9,7 @@ pub(crate) enum CommandError {
     InvalidArguments,
     IncompatibleMemo,
     InvalidMemo(String),
+    NonJsonNumberForAmount(String),
 }
 
 impl fmt::Display for CommandError {
@@ -25,6 +26,7 @@ impl fmt::Display for CommandError {
                 write!(f, "memo's cannot be sent to transparent addresses.")
             }
             InvalidMemo(e) => write!(f, "failed to interpret memo. {}", e),
+            NonJsonNumberForAmount(e) => write!(f, "{}", e),
         }
     }
 }
--- a/zingolib/src/commands/error.rs
+++ b/zingolib/src/commands/error.rs
@@ -27,7 +27,7 @@ impl fmt::Display for CommandError {
                 write!(f, "memo's cannot be sent to transparent addresses.")
             }
             InvalidMemo(e) => write!(f, "failed to interpret memo. {}", e),
-            NonJsonNumberForAmount => write!(f,  "amount is not a json::number::Number"),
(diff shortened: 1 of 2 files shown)


fluidvanadium:
noticed, but deferring to oscar

zingolib/src/commands/utils.rs
```diff
@@ -1,16 +1,20 @@
+#![forbid(unsafe_code)]
```
Oscar-Pepper: why is this attribute needed?

zingolib/src/commands/utils.rs
```diff
@@ -1,16 +1,20 @@
...
+//! Currently this mod contains a utility fn that's been factored out of SendCommand
```
Oscar-Pepper: I think there is a good chance additional functions could get added and get through review without realising this comment is now out of date.  Maybe its best the doc comment talks generally about what this mod is for?

```suggestion
//! Module containing utility functions for the commands interface
```

zingolib/src/commands/utils.rs
```diff
@@ -24,9 +28,19 @@ pub(super) fn parse_send_args(
...
                     .as_str()
-                    .ok_or(CommandError::UnexpectedType)?
+                    .ok_or(CommandError::UnexpectedType(
+                        "address not a Str!".to_string(),
+                    ))?
                     .to_string();
-                let amount = j["amount"].as_u64().ok_or(CommandError::UnexpectedType)?;
+                let amount = if !j["amount"].is_number() {
+                    return Err(CommandError::NonJsonNumberForAmount(
+                        "amount is not a json::number::Number".to_string(),
```
Oscar-Pepper: This additional string is unnecessary, the name of this error variant is so specific that this message should be in the error.rs fmt::Display

zingolib/src/commands/utils.rs
```diff
@@ -24,9 +28,19 @@ pub(super) fn parse_send_args(
...
+                    ));
+                } else {
+                    j["amount"].as_u64().ok_or(CommandError::UnexpectedType(
+                        "amount not a u64!".to_string(),
```
Oscar-Pepper, line 41: is there ever a case where `is_number()` is false but `as_u64()` would have returned `ok` incorrectly?
im not sure but maybe the `is_number()` check and `NonJsonNumberForAmount` is unneccessary?

zingolib/src/commands/utils.rs
```diff
@@ -66,7 +81,7 @@ mod tests {
     use crate::wallet;
 
     #[test]
-    fn parse_send_args() {
+    fn parse_send_args_test() {
```
Oscar-Pepper: I am of the opinion that the test attribute already describes this as a test and I quite like having a unit test with matching name to each function. But if this is your preferrence we can add the `_test`

zingolib/src/commands/utils.rs
```diff
@@ -103,4 +118,172 @@ mod tests {
             ]
         );
     }
+
+    use super::*;
```
Oscar-Pepper: can you replace `use crate::wallet;` instead so its at the top of the mod

zingolib/src/commands/utils.rs
```diff
@@ -103,4 +118,172 @@ mod tests {
...
+                    Err(CommandError::UnexpectedType(e)) => {
+                        assert_eq!(e, "address not a Str!".to_string())
+                    }
+                    _ => panic!(),
+                };
+            }
+            #[test]
+            fn non_u64_amount() {
+                let args =
+                ["[{\"address\": \"testaddress\", \"amount\": \"Oscar Pepper\", \"memo\": \"testmemo\"}]"];
```
Oscar-Pepper, line 189: Oscar Pepper? I think you mean Pesky Piper? :)

zingolib/src/commands/utils.rs
```diff
@@ -103,4 +118,172 @@ mod tests {
...
+                };
+            }
+        }
+        mod multi_string_args {
+            use super::*;
+            #[test]
+            fn two_args_wrong_amount() {
+                let args = ["testaddress", "foo"];
+                let result = parse_send_args(&args);
+                dbg!(&result);
```
Oscar-Pepper: please remove dbg

zingolib/src/commands/utils.rs
```diff
@@ -103,4 +118,172 @@ mod tests {
...
+                            "invalid digit found in string".to_string(),
+                            format!("{}", e)
+                        )
+                    }
+                    _ => panic!(),
+                };
+            }
+            #[test]
+            fn three_args_wrong_amount_show_trim_requirement() {
+                // Note the " " character after the 1.  The parser can handle by trimming, is that correct?
```
Oscar-Pepper: good comment but do we need the question?

zingolib/src/commands/utils.rs
```diff
@@ -103,4 +118,172 @@ mod tests {
...
+                let args = ["testaddress", "1 ", "whatever"];
+                let result = parse_send_args(&args);
+                dbg!(&result);
```
Oscar-Pepper: deboogs

zingolib/src/commands/utils.rs
```diff
@@ -103,4 +118,172 @@ mod tests {
...
+                match result {
+                    Ok(_) => (),
+                    _ => panic!(),
+                };
+            }
+            #[test]
+            fn wrong_number_of_args() {
+                let args = ["testaddress", "123", "3", "4"];
+                let result = parse_send_args(&args);
+                dbg!(&result);
```
Oscar-Pepper: and here

zingolib/src/commands/utils.rs
```diff
@@ -103,4 +118,172 @@ mod tests {
...
+                            )
+                        )
+                    }
+                    _ => panic!(),
+                };
+            }
+        }
+
+        #[test]
+        fn successful_parse_send_args_single_json() {
```
Oscar-Pepper: this is already tested for in `parse_send_args_test` ( and also in a mod called `failed_...`)

zingolib/src/commands/utils.rs
```diff
@@ -103,4 +118,172 @@ mod tests {
...
+                        assert_eq!(
+                            "invalid digit found in string".to_string(),
+                            format!("{}", e)
+                        )
+                    }
+                    _ => panic!(),
+                };
+            }
+            #[test]
+            fn three_args_wrong_amount_show_trim_requirement() {
```
Oscar-Pepper: as this is actually a test that is successful as oppose to verifying an error variant, can we move this to be a part of `parse_send_args_test` ?

zingolib/src/commands/error.rs
```diff
@@ -2,29 +2,33 @@ use std::fmt;
...
+            ArgsNotJson(e) => write!(f, "failed to parse argument. {}", e),
+            ParseIntFromString(e) => write!(f, "failed to parse argument. {}", e),
+            UnexpectedType(e) => write!(f, "arguments cannot be parsed to expected type. {}", e),
             MissingKey(key) => write!(f, "json array is missing \"{}\" key.", key),
             InvalidArguments => write!(f, "arguments given are invalid."),
             IncompatibleMemo => {
                 write!(f, "memo's cannot be sent to transparent addresses.")
             }
             InvalidMemo(e) => write!(f, "failed to interpret memo. {}", e),
+            NonJsonNumberForAmount(e) => write!(f, "{}", e),
```
Oscar-Pepper: see comment below

```suggestion
            NonJsonNumberForAmount => write!(f,  "amount is not a json::number::Number"),
```

zingolib/src/commands/error.rs
```diff
@@ -2,29 +2,33 @@ use std::fmt;
...
+            SingleArgNotJsonArray(e) => write!(f, "{}", e),
```
Oscar-Pepper: Missing display message

```suggestion
            SingleArgNotJsonArray(e) => write!(f, "argument cannot be parsed to a json array: {}", e),
```

zingolib/src/commands/utils.rs
Oscar-Pepper: alternatively we could remove the `fail_parse_send_args` module and split `parse_send_args_test` into separate tests in its corresponding json or multilple_args module? so we just have two modules in the test module (json and multiple_args) containing both tests that pass and tests that show the correct error

Oscar-Pepper (dismissed):
Good changes and really like the tests of each error variant. I have some change requests / questions

zingolib/src/commands/utils.rs
zancas: It's not necessary for the compiler, it's a readability/portability enhancement.  I'm meh on it.   Happy to remove.

zingolib/src/commands/utils.rs
zancas: Will fix.

zingolib/src/commands/utils.rs
zancas: Try removing `_test` and seeing what happens.

zingolib/src/commands/error.rs
zancas: I like your solution.

zingolib/src/commands/utils.rs
zancas: No.   If `is_number` is false then any invocation of `as_u64` would be `None`.

The (small) benefit of checking the "numberness" first is that a user will get a different message for `OscarPepper` (which isn't a JSON Number) and `-64` which *is* a JSON Number but not representable as a `u64`.   It's not more safe or correct, since both `-64` and `OscarPepper` are already rejected, but it's (sligthly) more informative (about *why*).

Is it worth the complexity, and readability costs?   I don't know.  I think that fewer instances of `UnexpectedType` being presented to the consumer is pretty valuable.   Clearly this isn't a complete solution.

zingolib/src/commands/utils.rs
zancas: The issue is that I stuffed my tests in with yours..  maybe they should be partitioned into a sub-mod?

Then I added "use super::*;`..  then I had a name collision.

I will have cake and eat too by...

zingolib/src/commands/utils.rs
zancas: I generally like more smaller tests, but I also like the unit-centric pattern you're using, where there's a "this unit works right" test.

So..  I went with your option 1.

zingolib/src/commands/utils.rs
zancas: I missed one!!?

zancas: No.. that's `outdated`.

zingolib/src/commands/utils.rs
zancas: I think we leave it in, because I don't have a totally unambiguous answer.

Oscar-Pepper (approved):
splendid

PR #900, opened 2024-04-07
removed an expect. no more expects are associated with send

fluidvanadium:
*associated

### DIFF
--- a/zingolib/src/wallet/send.rs
+++ b/zingolib/src/wallet/send.rs
@@ -274,7 +274,7 @@ impl LightWallet {
         let witness_trees = txmds_readlock
             .witness_trees
             .as_ref()
-            .expect("If we have spend capability we have trees");
+            .ok_or("No spend capability.")?;
         let (tx_builder, total_shielded_receivers) = match self
             .create_and_populate_tx_builder(
                 submission_height,


zingolib/src/wallet/send.rs
```diff
@@ -274,7 +274,7 @@ impl LightWallet {
         let witness_trees = txmds_readlock
             .witness_trees
             .as_ref()
-            .expect("If we have spend capability we have trees");
+            .ok_or("No spend capability.")?;
```
Oscar-Pepper, line 277: I think `ok_or_else` is preffered as its lazily evaluated
```suggestion
            .ok_or_else(|| "No spend capability.".to_string())?;
```

Oscar-Pepper (dismissed):

zingolib/src/wallet/send.rs
```diff
@@ -274,7 +274,7 @@ impl LightWallet {
         let witness_trees = txmds_readlock
             .witness_trees
             .as_ref()
-            .ok_or("No spend capability.")?;
+            .ok_or_else(|| "No spend capability.".to_string())?;
```
zancas: `?` is (almost certainly) lazily evaluating, or the `&str` ("No spend capability."), would not match the `..., String>` in the fn return type.   So we can go with the marginally more readable original version.

zancas (approved):

PR #901, opened 2024-04-07
added TransactionRecordMap

fluidvanadium:
Another non-functional change

### DIFF
--- a/zingolib/src/wallet/describe.rs
+++ b/zingolib/src/wallet/describe.rs
@@ -205,7 +205,7 @@ impl LightWallet {
         }
     }
 
-    /// Return a copy of the blocks current.maply in the wallet, needed to process possible reorgs
+    /// Return a copy of the blocks currently in the wallet, needed to process possible reorgs
     pub async fn get_blocks(&self) -> Vec<BlockData> {
         self.blocks.read().await.iter().cloned().collect()
     }
--- a/zingolib/src/wallet/describe.rs
+++ b/zingolib/src/wallet/describe.rs
@@ -53,6 +53,7 @@ impl LightWallet {
                 .read()
                 .await
                 .current
+                .map
                 .values()
                 .map(|transaction| {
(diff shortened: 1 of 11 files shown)


zingolib/src/wallet/describe.rs
```diff
@@ -204,7 +205,7 @@ impl LightWallet {
         }
     }
 
-    /// Return a copy of the blocks currently in the wallet, needed to process possible reorgs
+    /// Return a copy of the blocks current.maply in the wallet, needed to process possible reorgs
```
Oscar-Pepper: classic replace all :)

Oscar-Pepper (changes requested):
The diff wasnt checked, the replace all ran rampant!

Also I have no idea what the purpose of this is?
Current *IS* a map, why does it need a map field?
I assume its to do with seperating a spending from viewing wallet but more context in the description is needed here. It feels like to me there is already a lot of nested unnecessary complexity with Wallet -> Context -> TxMap -> Current where the map of transactions should probably be a direct field of the wallet? This PR adds an extra layer

zingolib/src/wallet/transactions/recording.rs
```diff
@@ -357,7 +367,7 @@ impl TxMapAndMaybeTrees {
         vout: &TxOut,
         output_num: u32,
     ) {
-        // Read or create the current TxId
+        // Read or create the current.map TxId
```
zancas, line 370: Typo?

zancas:
The benefit isn't obvious to me, but haven't tried too hard...

zancas (commented):
If this is not necessary to land 317, then let's deprioritize it.

zancas (changes requested):
Deprioritze non-essential work.

zancas:
Oh.. I think I see what's going on here...  I plan to reopen this..

zancas:
Superceded by #911

PR #911, opened 2024-04-09
Add transaction record map

zancas:
@fluidvanadium 's original hack was to add a native type that could have a foreign trait `zcash_client_backend::data_api::InputSource` implemented on it.

I've used a wrapper pattern with Deref and DerefMut implementations to all the "current" interface to be extended with the new interface (InputSource) while staying back compatible with the existing usage patterns.

This PR also pulls in the implementation of InputSource on TransactionRecordMap (I kept the same name as @fluidvanadium for the wrapper.)


zancas:
For the record all of the significant work in this PR was @fluidvanadium 's, I just cleaned it a little.

zancas:
OK, I stand by my original argument.  This type is added to provided an implementer for a foreign trait.

I vote we add this to dev, and I can hack on adding the functionality necessary to add the trait as a subsequent separate PR.

fluidvanadium (approved):
tested and approved

the Deref works out to save a lot of coding! how elegant!

PR #912, opened 2024-04-09
Add get received note from identifier method

zancas:
This PR builds on #911 if that one is rejected, then this one ought not to be considered.

This is the addition of ( get_received_note_from_identifier) an inherent method on TransactionRecordMap which is required to implement InputSource on TransactionRecordMap.


fluidvanadium (approved):
Right. Finally a clear pattern for querying the transaction database.

fluidvanadium (changes requested):
oops, the API changed and we need to redo this one.

fluidvanadium:
this will be superceded

PR #913, opened 2024-04-09
extend test framework

zancas:
In order to run the tests when I have `pwd` of zingolib/zingolib` that is, the *crate* root, I have to explicitly select the "test" feature like this:

`cargo nextest run --features test`

*BUT* if I run `cd $WORKSPACE_ROOT && cargo nextest run` it run just fine and picks up the tests inside `zingolib/zingolib`.

ChatGPT tells me that the `test` feature of the workspace member crate `zingolib/zingolib` is being picked up by "feature unification" in the workspace root.

I don't like this.    `cargo nextest run` in the workspace root shouldn't be running all of our tests *by accident*.

Debugging that feature unification issue might be interesting.. but..  if it's not borken...

The point of this PR is to build out some helpful things for writing **MUCH NEEDED** unit-tests.


fluidvanadium (approved):
tested ack

PR #922, opened 2024-04-10
transaction_record_map to new file

fluidvanadium:


zingolib/src/wallet.rs
```diff
@@ -53,6 +53,7 @@ pub mod notes;
 pub mod traits;
 pub mod transaction_context;
 pub mod transaction_record;
+pub mod transaction_record_map;
```
zancas: Are you sure this has to be `pub`?

zancas (approved):

PR #931, opened 2024-04-11
extracted mod witness_trees

fluidvanadium:
This is a refactor change.


zancas:
I think the `mod.rs` pattern is the old way of doing things.  I prefer not to have any file named `mod.rs`.

zingolib/src/data/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod witness_trees;
```
zancas: I think you can just put this in data.rs, and delete data/mod.rs.

zancas (changes requested):

zingolib/src/data/mod.rs
fluidvanadium: ok

zancas (approved):

PR #934, opened 2024-04-11
Enforce doc comments

idky137:
This PR will create clippy warnings for any pub functions that do not have Doc comments.

All current pub functions that do not already have a doc comment will be given "/// TODO: Add Doc Comment Here!". This way the code base will build without new warnings, but all new functions should be properly documented.


fluidvanadium:
can we try this against stable

zingo-netutils/src/lib.rs
```diff
@@ -1,3 +1,7 @@
+//! Zingo-Netutils
```
zancas, line 1: @AloeareV can you provide a more information doc-comment here?

zingo-status/src/confirmation_status.rs
```diff
@@ -1,3 +1,5 @@
+//! TODO: Add Mod Description Here!
```
zancas, line 1: @fluidvanadium can you make this doc-comment more useful?

zingo-status/src/lib.rs
```diff
@@ -1,4 +1,6 @@
-#[forbid(unsafe_code)]
+//! Zingo-Status
+//! TODO: Add Crate Discription Here!
```
zancas, line 2: @fluidvanadium can you please add a comment here?

idky137:
Finished adding todos. There is still one warning/error generated. This is because a function is being generated in the build process, I believe the way to solve this is to add "#[allow(missing_docs)]" to the module that is generating this function. However I am unaware of where this is.

zingo-memo/src/lib.rs
```diff
@@ -1,3 +1,9 @@
+//! Zingo-Memo
+//! Utilities for procedural creation and parsing of the Memo field
+//! These memos will likely never be directly exposed to the user,
```
zancas: I prefer to remove this speculative statement.

zingo-memo/src/lib.rs
```diff
@@ -1,3 +1,9 @@
...
+//! but instead write down things on-chain, for recovery after rescan, etc
```
zancas: It's not clear to me that things that are procedurally generated/parsed by zingo-memo will always be mutually exclusive with end-user relevant content.   Perhaps we should remove this line, or refine it to be clearer?

zingo-netutils/src/lib.rs
zancas: Are all of the current net-utils exclusively for test?

zancas (changes requested):

zingo-memo/src/lib.rs
AloeareV: Why?

zingo-memo/src/lib.rs
AloeareV: > It's not clear to me that things that are procedurally generated/parsed by zingo-memo will always be mutually exclusive with end-user relevant content. Perhaps we should remove this line, or refine it to be clearer?

It's presumptive...but we can always re-evaluate it if we come up with a use-case. I can reword it to be a 'currently'.

AloeareV: Reworded.

zingoconfig/src/lib.rs
```diff
@@ -1,4 +1,8 @@
+//! ZingConfig
```
zancas, line 1: @Oscar-Pepper can you please add a description of what this crate is, here?    No need for perfection, just enough to give a reader a general idea.  Fine to include `TODO`s for enhancements to the doc-comment.

zingolib/src/wallet.rs
```diff
@@ -190,23 +218,23 @@ pub struct LightWallet {
...
+    /// Progress of an outgoing transaction
     send_progress: Arc<RwLock<SendProgress>>,
 
-    // The current price of ZEC. (time_fetched, price in USD)
+    /// The current price of ZEC. (time_fetched, price in USD)
     pub price: Arc<RwLock<WalletZecPriceInfo>>,
 
-    // Local state needed to submit [compact]block-requests to the proxy
-    // and interpret responses
+    /// Local state needed to submit [compact]block-requests to the proxy
```
AloeareV: Square brackets have special meaning in doc comments as intra-doc links, which causes this to fail cargo doc, as this is a broken link.

AloeareV (dismissed):

zingolib/src/wallet.rs
AloeareV: Oh hang on I will just fix this, that's easier than leaving a comment.

fluidvanadium:
how do i run this doc-comment completeness test locally?

zancas (dismissed):

zancas (dismissed):

zancas (dismissed):

zancas (approved):

PR #939, opened 2024-04-11
Input source

fluidvanadium:
This implements input source on TransactionRecordsById and tests it.

The test fails.


fluidvanadium:
looks like this is an agglomeration of multiple changes
-mocks
-move some methods from transactions.rs into transaction_records_by_id.rs
-impl InputSource on transaction_records_by_id and test

should be easy enough to separate. thats my next goal

PR #942, opened 2024-04-12
Expand mocks

fluidvanadium:
lives on top of #940 

all internal structs are mocked in their respective files and this mod mocks is handles external library structs.

PR #940, opened 2024-04-12
moved transaction_records_by_id methods into it

fluidvanadium:
This is a step toward implementing input source.

A number of unit tests are incoming

PR #944, opened 2024-04-12
Expand mocks 2

fluidvanadium:
lives on top of #940 #941 #942 
Fix some test features flags and added a test to TRBI


zingolib/src/lightclient.rs
```diff
@@ -668,6 +668,7 @@ async fn get_recent_median_price_from_gemini() -> Result<f64, PriceFetchError> {
 }
 
 #[cfg(test)]
+#[cfg(feature = "test-features")]
```
AloeareV, line 671: From a look at the rust reference, I believe that multiple cfg attributes are additive. This would be much clearer if it's a single `#[cfg(all(test, feature = "test-features"))]`

AloeareV (changes requested):

zingolib/src/lightclient.rs
AloeareV: I'm starting to think that I'm wrong and `#[cfg(any(test, feature = "test-features"))]` is correct, in which case that would still also be much clearer!

zingolib/src/test_framework/mocks.rs
```diff
@@ -1,62 +1,88 @@
...
+                recipient: address,
+                value: NoteValue::from_raw(1000000),
+                rseed: Rseed::AfterZip212([7; 32]),
+            }
         }
     }
 }
+
+#[allow(dead_code)]
+pub(crate) fn mock_sapling_crypto_note() -> sapling_crypto::Note {
```
fluidvanadium, line 81: can we enable a feature to pull a mock straight from a slightly modded LRZ?

PR #957, opened 2024-04-18
Add derive builder

zancas:


zingolib/src/test_framework/mocks.rs
```diff
@@ -73,58 +73,30 @@ pub fn random_zaddr() -> (
...
-            }
-        }
-
-        // Methods to set each field
-        build_method!(recipient, PaymentAddress);
-        build_method!(value, NoteValue);
-        build_method!(rseed, Rseed);
-
-        /// For any old zcaddr!
-        pub fn randomize_recipient(self) -> Self {
```
fluidvanadium: this function may be necessary to distinguish notes in unit tests

zingolib/src/wallet/notes/sapling.rs
```diff
@@ -189,104 +204,37 @@ impl ShieldedNoteInterface for SaplingNote {
 #[cfg(any(test, feature = "test-features"))]
 pub mod mocks {
```
fluidvanadium: if this mod is empty, lets remove it

zingolib/src/wallet/notes/sapling.rs
```diff
@@ -189,104 +204,37 @@ impl ShieldedNoteInterface for SaplingNote {
...
-    }
 }
 
 #[cfg(test)]
 mod tests {
-    use super::mocks::SaplingNoteBuilder;
+    use crate::test_framework::mocks::LRZSaplingNoteBuilder;
 
     #[test]
     pub fn build_sapling_note() {
```
fluidvanadium: this test needs to build a SaplingNote, not just an LRZSaplingNote.

fluidvanadium (changes requested):
its an improvement! it also needs some

zingolib/src/wallet/notes/sapling.rs
zancas: It does...   though.. I am not sure that a LRZSaplingNoteBuilder is still necessary.

zancas: The `to_lrz` method converts to a Sapling Note.

zingolib/src/wallet/notes/sapling.rs
zancas: I think we'll want to reorganize mods a bit, perhaps as it's own PR (including removing this empty one).   I thought it should be it's own mod/name PR.

zingolib/src/wallet/notes/sapling.rs
fluidvanadium: the to_lrz method converts to a sapling_crypto::note

zingolib/src/test_framework/mocks.rs
```diff
@@ -71,60 +70,32 @@ pub fn random_zaddr() -> (
 }
 
 // Sapling Note Mocker
-mod sapling_note {
+pub(crate) mod sapling_note {
```
fluidvanadium: this mod should NOT be called sapling_note!!

zingolib/src/wallet/notes/sapling.rs
fluidvanadium: ok, open more PRs for these good ideas

zingolib/src/test_framework/mocks.rs
fluidvanadium: MY BAD!

zingolib/src/lightclient/deprecated.rs
```diff
@@ -72,7 +72,7 @@ impl LightClient {
             .sapling_notes
             .iter()
             .filter(|nd| nd.is_change)
-            .map(|nd| nd.note.value().inner())
+            .map(|nd| nd.sapling_crypto_note.value().inner())
```
AloeareV, line 75: :smiley:

fluidvanadium (dismissed):
I love #[derive(Builder

however, we cannot derive Builder on sapling_crypto::Note.
instead, i want us to set aside this portion of the changes and focus on landing add_derive_builder into point where it works best, which is zingolib/src/wallet/notes/sapling.rs SaplingNote

zingolib/src/wallet/notes/sapling.rs
```diff
@@ -1,57 +1,75 @@
+#![allow(missing_docs)]
```
AloeareV: I don't like this in the sapling mod. Why is this here? Is it the derive-generated SaplingNoteBuilder? Surely there's a way around putting this on the whole mod.

zingolib/src/wallet/notes/sapling.rs
```diff
@@ -1,57 +1,75 @@
...
+use crate::wallet::{notes::ShieldedNoteInterface, traits::FromBytes};
+
 use super::{
     super::{data::TransactionRecord, Pool},
-    NoteInterface, ShieldedNoteInterface,
+    NoteInterface,
 };
 
 /// TODO: Add Doc Comment Here!
+#[derive(Builder)]
```
AloeareV, line 14: I don't think this wants to exit in production code. Can we toss this behind a [cfg_attr](https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg_attr-attribute)?

AloeareV (changes requested):

zingolib/src/wallet/notes/sapling.rs
zancas: I can shrink the scope.

zingolib/src/wallet/notes/sapling.rs
zancas: Yes.  It can.
Once that's done then I push the default impls into a feature-gated explicit impl block.

PR #962, opened 2024-04-18
added conditional compilation to os-specific save methods

fluidvanadium:
This is an improvement to take advantage of conditional compilation.


zingolib/src/lightclient/save.rs
```diff
@@ -79,6 +85,7 @@ impl LightClient {
             .map_err(String::from)
     }
 
+    #[cfg(not(any(target_os = "ios", target_os = "android")))]
     pub(super) fn write_file_if_not_exists(dir: &Path, name: &str, bytes: &[u8]) -> io::Result<()> {
```
Oscar-Pepper, line 92: this function is called in `set_sapling_params` which also calls `get_zcash_params_path` which includes code for mobile archs. this is function is therefore broken for ios/android. could you investigate?

Oscar-Pepper (changes requested):

zingolib/src/lightclient/save.rs
fluidvanadium: oh crap! this would not have compiled for mobile

fluidvanadium: feature-gated that block

PR #963, opened 2024-04-18
Impl inp src for trbid

AloeareV:
@fluidvanadium I think this is the minimal functionality needed to get the input source implementation working (the commit history is wild enough that I'd consider some sort of rebase-like git commit cleanup, but the changes that actually made it to the diff are the right ones). Maybe the helper functions should be in a precurser pr...but they'd be temporarily dead code. 

@zancas You were asking about unit tests,  these input_source functions all need unit tests...I was trying to write a unit test for get_spendable_transparent_output recently and was blocked on it. #957 probably wants to land before those unit tests, iiuc you're working on getting that one working with the explicit default branch you mentioned to me somewhere? 

@myself, this pr is still draft because get_spendable_transparent_notes (with an s) is still todoed, I'm hoping to get that done tomorrow and undraft this.


zingolib/src/wallet/traits.rs
```diff
@@ -532,6 +533,9 @@ where
     /// TODO: Add Doc Comment Here!
     fn get_tree(tree_state: &TreeState) -> &String;
 
+    /// The [zcash_protocol::ShieldedProtocol] this domain represents
```
zancas: Odd that a `DomainWalletExt` has a protocol method that's specific to ShieldedProtocols.   This seems like a name mismatch.

zingolib/src/wallet/traits.rs
```diff
@@ -599,6 +603,10 @@ impl DomainWalletExt for SaplingDomain {
         &tree_state.sapling_tree
     }
 
+    fn protocol() -> ShieldedProtocol {
+        ShieldedProtocol::Sapling
```
zancas: Given that this trait is being implemented on a domain that has a type including Domain specitic information..   why aren't the trait methods referencing something via (Self)?   Maybe the same thing for each implementation..  maybe even a default implementation?

zingolib/src/wallet/traits.rs
```diff
@@ -674,6 +682,10 @@ impl DomainWalletExt for OrchardDomain {
         &tree_state.orchard_tree
     }
 
+    fn protocol() -> ShieldedProtocol {
+        ShieldedProtocol::Orchard
```
zancas: As with Sapling....

zancas (changes requested):

zingolib/src/wallet/transaction_records_by_id.rs
```diff
@@ -39,16 +49,252 @@ impl TransactionRecordsById {
...
+            NoteRecordIdentifier,
+            <D as zcash_note_encryption::Domain>::Note,
+        >,
+    >
+    where
+        <D as zcash_note_encryption::Domain>::Note: PartialEq + Clone,
+        <D as zcash_note_encryption::Domain>::Recipient: super::traits::Recipient,
+    {
+        let transaction = self.get(&note_record_reference.txid);
+        transaction.and_then(|transaction_record| {
```
zancas: This seems to treat a found transparent note in the same way as no note.  Is that correct?!

zancas: In both cases this function returns `None`.

zingolib/src/wallet/transaction_records_by_id.rs
```diff
@@ -39,16 +49,252 @@ impl TransactionRecordsById {
...
+        })
+    }
 }
+
+impl InputSource for TransactionRecordsById {
+    type Error = ZingoLibError;
+    type AccountId = zcash_primitives::zip32::AccountId;
+    type NoteRef = NoteRecordIdentifier;
+
+    fn get_spendable_note(
```
zancas: This convention using the word "note" to mean "shielded value transfer mechanism" is confusing.

AloeareV: "Note" is the established zcash term used to mean "Shielded store of received value", so yes, I think describing a "note" as a "shielded value transfer mechanism" is accurate.

zingolib/src/wallet/transaction_records_by_id.rs
AloeareV: > This seems to treat a found transparent note in the same way as no note. Is that correct?!

This function can't find a "transparent note". We need to fix our terminology in some other parts of this crate, as "notes" are, by definition, shielded. `D` is a (shielded) `Domain`, and this method returns a note of that domain's associated note type.

zingolib/src/wallet/traits.rs
AloeareV: shielded_protocol would be a more specific (and therefore clearer) name for this function, yes.

zingolib/src/wallet/traits.rs
AloeareV: We could make this an associated constant instead of a method I think, yeah,

zingolib/src/wallet/transaction_record.rs
```diff
@@ -203,6 +205,39 @@ impl TransactionRecord {
         }
     }
 
+    /// For each Shielded note received in this transactions,
+    /// pair it with a NoteRecordIdentifier identifying the note
+    /// and return the list
+    // TODO: Make these generic, this is wet code
+    pub fn select_unspent_shnotes_and_ids<D>(
```
fluidvanadium, line 225: please add unit test

zingolib/src/wallet/transaction_record.rs
```diff
@@ -282,6 +317,44 @@ impl TransactionRecord {
             self.total_orchard_value_spent,
         ]
     }
+
+    /// Gets a received note, by index and domain
+    pub fn get_received_note<D>(
```
fluidvanadium, line 337: please add unit test

zingolib/src/wallet/transaction_records_by_id.rs
```diff
@@ -40,13 +42,36 @@ impl TransactionRecordsById {
...
-    // Constructs a TransactionRecordMap from a HashMap
+    /// Constructs a TransactionRecordsById from a HashMap
     pub fn from_map(map: HashMap<TxId, TransactionRecord>) -> Self {
         TransactionRecordsById(map)
     }
 }
-/// Methods to modify the map.
+
+/// Methods to query and modify the map.
 impl TransactionRecordsById {
```
fluidvanadium, line 52: please add unit test

zingolib/src/wallet/transaction_records_by_id/input_source.rs
```diff
@@ -0,0 +1,317 @@
...
+                    "insufficient funds, short {}",
+                    missing_value_after_orchard.into_u64()
+                )));
+            };
+        };
+
+        Ok(SpendableNotes::new(sapling_notes, orchard_notes))
+    }
+
+    fn get_unspent_transparent_output(
```
fluidvanadium, line 185: please add unit test

zingolib/src/wallet/transaction_records_by_id/input_source.rs
```diff
@@ -0,0 +1,317 @@
...
+    error::{ZingoLibError, ZingoLibResult},
+    wallet::{notes::ShNoteId, transaction_records_by_id::TransactionRecordsById},
+};
+
+impl InputSource for TransactionRecordsById {
+    type Error = ZingoLibError;
+    type AccountId = zcash_primitives::zip32::AccountId;
+    type NoteRef = ShNoteId;
+
+    fn get_spendable_note(
```
fluidvanadium, line 52: please add unit test

fluidvanadium (changes requested):
it is really coming together! every new method should have its own unit test

zingolib/src/wallet/transaction_record.rs
```diff
@@ -561,15 +634,35 @@ pub mod mocks {
                 .build(),
         );
 
+        transaction_record
+    }
+    /// creates a transaction_record with one of
+    /// each mock note type in it
+    pub fn setup_mock_transaction_record() -> TransactionRecord {
+        let mut transaction_record = TransactionRecordBuilder::default().build();
+        transaction_record
```
fluidvanadium: there is a fluent builder pattern that would streamline this function

zingolib/src/wallet/transaction_records_by_id.rs
```diff
@@ -40,13 +42,36 @@ impl TransactionRecordsById {
...
+    pub fn get_received_note_from_identifier<D: DomainWalletExt>(
```
fluidvanadium: is there a unit test for this?

zingolib/src/wallet/transaction_records_by_id.rs
```diff
@@ -507,4 +514,31 @@ mod tests {
...
+        trbid.insert_transaction_record(setup_mock_transaction_record());
+
+        let received_note = trbid.get_received_note_from_identifier::<SaplingDomain>(
+            crate::wallet::notes::ShNoteId {
+                txid: default_txid(),
+                shpool: ShieldedProtocol::Sapling,
+                index: 0,
+            },
+        );
+
```
fluidvanadium, line 607: can we expand this test to prove the correct note is selected between 2 options?

zingolib/src/wallet/transaction_record.rs
fluidvanadium: nice

zingolib/src/wallet/transaction_record.rs
```diff
@@ -731,4 +817,50 @@ mod tests {
...
         assert_eq!(transaction_record.total_value_received(), old_total);
     }
+
+    #[test]
+    fn select_unspent_shnotes_and_ids() {
+        let transaction_record = setup_mock_transaction_record();
+
+        let sapling_notes = transaction_record.select_unspent_shnotes_and_ids::<SaplingDomain>();
+        assert_eq!(
+            sapling_notes.first().unwrap().0,
```
fluidvanadium, line 889: first is good, but can we prove that this skips a spent note?

zingolib/src/wallet/transaction_records_by_id.rs
```diff
@@ -40,13 +42,36 @@ impl TransactionRecordsById {
...
+            <D as zcash_note_encryption::Domain>::Note,
+        >,
+    >
+    where
+        <D as zcash_note_encryption::Domain>::Note: PartialEq + Clone,
+        <D as zcash_note_encryption::Domain>::Recipient: super::traits::Recipient,
+    {
+        let transaction = self.get(&note_record_reference.txid);
+        transaction.and_then(|transaction_record| {
+            if note_record_reference.shpool == D::SHIELDED_PROTOCOL {
```
fluidvanadium: this if statement should go before getting the transaction

zingolib/src/wallet/transaction_records_by_id.rs
```diff
@@ -482,4 +514,31 @@ mod tests {
 
         assert_eq!(transaction_records_by_id.len(), 1);
     }
+
+    #[test]
+    fn get_received_note_from_identifier() {
+        let mut trbid = TransactionRecordsById::new();
+        trbid.insert_transaction_record(setup_mock_transaction_record());
```
fluidvanadium: this test would also be more powerful against nine

zingolib/src/wallet/transaction_records_by_id/input_source.rs
```diff
@@ -0,0 +1,368 @@
...
+use crate::{
+    error::{ZingoLibError, ZingoLibResult},
+    wallet::{notes::ShNoteId, transaction_records_by_id::TransactionRecordsById},
+};
+
+impl InputSource for TransactionRecordsById {
+    type Error = ZingoLibError;
+    type AccountId = zcash_primitives::zip32::AccountId;
+    type NoteRef = ShNoteId;
+
```
fluidvanadium, line 45: it would be great to add doc-comments

zingolib/src/wallet/transaction_records_by_id/input_source.rs
```diff
@@ -0,0 +1,368 @@
...
+        }
+        let mut sapling_notes = Vec::<ReceivedNote<ShNoteId, sapling_crypto::Note>>::new();
+        let mut orchard_notes = Vec::<ReceivedNote<ShNoteId, orchard::Note>>::new();
+        if let Some(missing_value_after_sapling) = sapling_note_noteref_pairs.into_iter().rev(/*biggest first*/).try_fold(
+            Some(target_value),
+            |rolling_target, (note, noteref)| match rolling_target {
+                Some(targ) => {
+                    sapling_notes.push(
+                        self.get(&noteref.txid).and_then(|tr| tr.get_received_note::<SaplingDomain>(noteref.index))
+                            .ok_or_else(|| ZingoLibError::Error("missing note".to_string()))?
```
fluidvanadium: this error implies we somehow misrepresented the txid across this function. it should never happen, so im fine with this for now

zingolib/src/wallet/transaction_records_by_id/input_source.rs
```diff
@@ -0,0 +1,368 @@
...
+                    );
+                    Ok(targ
+                        - NonNegativeAmount::from_u64(note.value().inner())
```
fluidvanadium: if this is expected to become None, why does it need error handling?

zingolib/src/wallet/transaction_records_by_id/input_source.rs
```diff
@@ -0,0 +1,368 @@
...
+                            .map_err(|e| ZingoLibError::Error(e.to_string()))?)
+                }
+                None => Ok(None),
+            },
+        )? {
+            if let Some(missing_value_after_orchard) = orchard_note_noteref_pairs.into_iter().rev(/*biggest first*/).try_fold(
+            Some(missing_value_after_sapling),
```
fluidvanadium: pls indent

zingolib/src/wallet/transaction_records_by_id/input_source.rs
```diff
@@ -0,0 +1,368 @@
...
+        assert_eq!(
+            real_single_note.note(),
+            &zcash_client_backend::wallet::Note::Sapling(
+                SaplingCryptoNoteBuilder::default().build()
+            )
+        )
+    }
+
+    #[test]
+    fn select_spendable_notes() {
```
fluidvanadium: needs proptest

zingolib/src/wallet/transaction_records_by_id/input_source.rs
AloeareV: This error would be if note.value().inner() was not a valid value for NonNegativeAmount::from_u64

fluidvanadium (approved):
fantastic!

PR #964, opened 2024-04-18
update note names

zancas:


zingolib/src/lightclient/deprecated.rs
```diff
@@ -72,13 +72,13 @@ impl LightClient {
             .sapling_notes
             .iter()
             .filter(|nd| nd.is_change)
-            .map(|nd| nd.note.value().inner())
+            .map(|nd| nd.sapling_note_crypto.value().inner())
```
Oscar-Pepper: this is a note from the sapling_crypto crate so we should call all these notes `saping_crypto_note` instead of `sapling_note_crypto`

zingolib/src/lightclient/deprecated.rs
```diff
@@ -72,13 +72,13 @@ impl LightClient {
...
             .sum::<u64>()
             + wallet_transaction
                 .orchard_notes
                 .iter()
                 .filter(|nd| nd.is_change)
-                .map(|nd| nd.note.value().inner())
+                .map(|nd| nd.orchard_note_crypto.value().inner())
```
Oscar-Pepper: the orchard crate is just called `orchard` not `orchard_crypto`. do we change this to `orchard_crypto_note` for consistency, or `orchard_note` to match the crate or ALSO change to only `sapling_note` for consistency .... OR ... we choose a new name for all zcash stuff like `orchard_zcash_note` and `sapling_zcash_note`

the choice be yours...

Oscar-Pepper (dismissed):
I think it a good change but words be a bit muddled IMO

fluidvanadium (approved):
claro!

PR #965, opened 2024-04-19
prefer transaction id over txid

zancas:


fluidvanadium:
in this case, i disagree because TxId is the LRZ type we rely on.

zancas:
I see your point, but couldn't that actually be an argument in favor?   The developer can *read* the foreign crate type in the signature..  now..  without further action it's *disambiguated* by the `ident` as "a transaction" as opposed to e.g. "a transmission".   So our more-explicit naming convention produces an `ident` that helps to (locally) fix an ambiguity introduced by the slightly less careful name being used elsewhere?

PR #966, opened 2024-04-19
Wallet read part 1

fluidvanadium:
this implements part of WalletRead on TxMapAndMaybeTrees


zancas:
OK here's my thought...  you're basically interfacing with the lrz "Account" idea.

In order to do so, you need to add a bunch of new dependencies.

I propose a whole new crate:  `zingo-account` with all the new functionality contained inside.   Then it can have lots of separation of concern, encapsulation, modularization, etc.   What say you?

zancas (commented):
I propose breaking ZingoAccount into a micro-crate.

zingolib/src/wallet/transactions.rs
```diff
@@ -38,3 +50,255 @@ impl TxMapAndMaybeTrees {
         self.witness_trees.as_mut().map(WitnessTrees::clear);
     }
 }
+
+pub struct ZingoAccount(AccountId, UnifiedFullViewingKey);
```
zancas, line 54: I vote we move this thing somewhere else..   like a `zingo_account` mod.   It's only (minimally) implemented to fulfill an API constraint, and leaving it sitting around here makes the "transactions.rs" mod less readable, and more cluttered.

zancas (changes requested):

zingolib/src/wallet/transactions.rs
```diff
@@ -38,3 +50,255 @@ impl TxMapAndMaybeTrees {
...
+    fn ufvk(&self) -> Option<&UnifiedFullViewingKey> {
+        Some(&self.1)
+    }
+
+    fn uivk(&self) -> zcash_keys::keys::UnifiedIncomingViewingKey {
+        unimplemented!()
+    }
+}
+
+impl WalletRead for TxMapAndMaybeTrees {
```
zancas, line 74: It seems like these methods are all getters.  Maybe this implementation properly belongs in `get.rs` ?

zancas (changes requested):

zingolib/src/wallet/transactions.rs
zancas: This comment is superseded by:  #980

PR #967, opened 2024-04-19
Note query

fluidvanadium:
Adds a struct for querying notes properties.

### DIFF
--- a/zingolib/src/wallet/notes.rs
+++ b/zingolib/src/wallet/notes.rs
@@ -8,6 +8,7 @@ pub mod sapling;
 pub use sapling::SaplingNote;
 pub mod orchard;
 pub use orchard::OrchardNote;
+pub mod query;
 
 use zcash_client_backend::PoolType;
 use zcash_primitives::transaction::TxId;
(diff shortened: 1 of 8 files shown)


zingolib/src/wallet/notes.rs
```diff
@@ -97,3 +98,91 @@ pub mod mocks {
...
+    #[test]
+    fn note_queries() {
+        let spend = Some((default_txid(), 112358));
+
+        let transparent_unspent_note = TransparentNoteBuilder::default().build();
+        let transparent_pending_spent_note = TransparentNoteBuilder::default()
+            .unconfirmed_spent(spend)
+            .build();
+        let transparent_spent_note = TransparentNoteBuilder::default().spent(spend).build();
+        let sapling_unspent_note = SaplingNoteBuilder::default().build();
```
zancas, line 123: I don't see this testing against orchard notes for some reason?

zancas: Not a blocker we can add that.

zancas (approved):
It seems like OrchardNote isn't tested?   Maybe I just missed it?  It's not a blocker if it's missing we can add it.

PR #970, opened 2024-04-20
Parse send args to final types

Oscar-Pepper:
Problem statement: In order to create a `send all` command for a zip317 compatibility we need to be able to distinguish a single argument address (in the case `send all` has no memo) from a single argument json. currently we decode the address much later in `do_send` and therefore if it not json parsable we must assume it is a valid address.

This PR solves this issue by parsing the command arguments into their correct types (or returns a specific error) immediately. This also means there are less stages of conversion in the codebase.

This PRs:
- adds converters for LRZ Address and NonNegativeAmount (Zatoshi) types
- moves send/shield command argument parsing to the top layer, removing `map_tos_to_receivers` in `do_send`

As a by-product:
- changes tests and test framework to use `do_send_test_only` and `do_shield_test_only`
- changes tests to give memos as &str
- fixes a typo and adds doc comments on affected entities
- helperises shield command parsing (with unit test!)
- moves checking memo compatibility into parse_send_args which makes sense and also declutters SendCommand
- improves error handling
- deprecates `is_shielded_address`

### DIFF
--- a/darkside-tests/src/utils.rs
+++ b/darkside-tests/src/utils.rs
@@ -722,7 +722,7 @@ pub mod scenarios {
                 DarksideSender::ExternalClient(lc) => lc,
             };
             lightclient
-                .do_shield(&[pool_to_shield], None)
+                .do_shield_test_only(&[pool_to_shield], None)
                 .await
                 .unwrap();
             let mut streamed_raw_txns = self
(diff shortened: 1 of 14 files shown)


fluidvanadium (dismissed):
excellent

fluidvanadium (approved):
lightclient/send.rs actually shrinks! awesome

PR #971, opened 2024-04-20
also fix quicksend description

Oscar-Pepper:


juanky201271 (approved):
Thanks, this is perfect.

PR #972, opened 2024-04-20
Poc microcrate account

zancas:
This is an experiment.  I don't have a great idea for what the tradeoffs might be, but all other things being equal, I think more encapsulated is better.

### DIFF
--- a/zingo-account/src/lib.rs
+++ b/zingo-account/src/lib.rs
@@ -1,6 +1,9 @@
 //! A micro crate to provide a wrapper around the zcash_client_backend account trait.
 //! By constraining the wrapper to be independent of internal zingolib functionality
 //! we expose a shareable component, and refine the definition of zingolib.
+
+/// The a view-only abstraction that provides a management interface for
+/// TODO:  Explain more about our Account abstraction.
 pub struct ZingoAccount(
     pub zip32::AccountId,
     pub zcash_keys::keys::UnifiedFullViewingKey,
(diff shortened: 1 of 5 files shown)


zancas:
I like it that there are no `use` statements.   I like it that it can be plugged into `zingolib` as a "completely" contained component.    It should be easy to integration test..

zingo-account/src/lib.rs
```diff
@@ -0,0 +1,28 @@
+//! A micro crate to provide a wrapper around the zcash_client_backend account trait.
+//! By constraining the wrapper to be independent of internal zingolib functionality
+//! we expose a shareable component, and refine the definition of zingolib.
+
+/// The a view-only abstraction that provides a management interface for
```
Oscar-Pepper, line 5: "The a view-only" is a typo? also comment seems to just ...end. I am ok to chalk that up to the TODO if you want

zingo-account/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+/// TODO:  Explain more about our Account abstraction.
+pub struct ZingoAccount(
+    pub zip32::AccountId,
+    pub zcash_keys::keys::UnifiedFullViewingKey,
```
Oscar-Pepper, line 9: I dont think we should make fields pub unless there is a really good reason? is there one?
the `id` and `ufvk` methods are private though which is surprising. Have you looked at how this Account trait is used in LRZ?

Oscar-Pepper:
I think we should probably have a crate for many "primitive" wallet objects like account. I think one crate for just this one struct will end up with a silly amount of crates. Or we just have the struct in zingolib.

PR #973, opened 2024-04-20
Break error into micro crate

zancas:
An argument in favor:   The codebase inside zingolib is shrinking.


Oscar-Pepper (commented):
NAck

I would like some actual advantages before we consider this...
as you said a "crate" is a stand-a-lone part so having lots of crates all depend on one error crate doesnt seem right to me. Having the errors of one crate in another crate doesnt seem right either. also I cant actually think of an application of a workspace-wide error. most likely if our codebase was as suggested int his PR, I would think its a good idea to open a PR which sorts the big heap of random error in this new crate into the crates it belongs in. therefore this seems like an anti-pattern to me

PR #977, opened 2024-04-22
Query sum value record

fluidvanadium:
This adds a new method to transaction_record and unit tests it.


fluidvanadium:
closed because the merge-base was changed too much. reopening

PR #978, opened 2024-04-22
Add propose all command

Oscar-Pepper:


fluidvanadium (approved):
this looks awesome! thanks for fixing the changelog
good that feature gated methods are starting to align

zingolib/src/commands/utils.rs
```diff
@@ -317,19 +327,10 @@ mod tests {
...
-                            e,
-                            format!(
-                                "Error creating output. Memo '\"{}\"' is too long",
-                                long_513_byte_memo
-                            )
-                        )
-                    }
-                    _ => panic!(),
-                };
+                assert!(matches!(
```
zancas, line 330: I don't like the `matches!` macro..  what's itz purpose?

PR #980, opened 2024-04-22
Migrate functionality from TxMapAndMaybeTrees to TRBID where appropriate.

zancas:
Draft PR #982 builds on this PR.

I tried to note where methods could not because of their dependence on witness trees.

Methods that *have* moved are tested and shown to be dependent only on `transaction_records_by_id` (and therefore should be implented on a trbid).


fluidvanadium (dismissed):
this is awesome! it had to happen at some point, the sooner the better

zingolib/src/wallet/transactions/recording.rs
```diff
@@ -21,33 +21,12 @@ use crate::{
         traits::{self, DomainWalletExt, Nullifier, Recipient},
     },
 };
-
-use super::TxMapAndMaybeTrees;
-impl TxMapAndMaybeTrees {
```
fluidvanadium, line 26: it is confusing to me that this file has impl blocks for both

fluidvanadium (changes requested):
i really like what is going on here, however i feel very strongly that impl TransactionRecordsById does not go in transactions/recording.rs next to TxMapAndMaybeTrees

zancas:
That's where the methods were implemented before.  In "recording.rs".  I'm pretty sure you put them there.

zingolib/src/wallet/transactions/recording.rs
```diff
@@ -446,16 +293,172 @@ impl TxMapAndMaybeTrees {
...
+            }
+        } else {
+            ZingoLibError::UnknownError.handle()?
+        }) // todO add special error variant
     }
 }
 
 // shardtree
-impl TxMapAndMaybeTrees {
+impl super::TxMapAndMaybeTrees {
```
AloeareV: I believe we're moving in favor of absolute paths over super. This is more verbose, is is clearer?
```suggestion
impl crate::wallet::transactions::TxMapAndMaybeTrees {
```

AloeareV (dismissed):

zingolib/src/wallet/transactions/recording.rs
```diff
@@ -21,33 +21,12 @@ use crate::{
...
-                .truncate_removing_checkpoint(&(reorg_height - 1))
-                .expect("Infallible");
-            t.witness_tree_orchard
-                .truncate_removing_checkpoint(&(reorg_height - 1))
-                .expect("Infallible");
-            t.add_checkpoint(reorg_height - 1);
-        }
-    }
-
+impl super::TransactionRecordsById {
```
fluidvanadium: please move this impl block to either transaction_records_by_id.rs or a submod of transaction_records_by_id

fluidvanadium (changes requested):

zingolib/src/wallet/transactions/recording.rs
zancas: Done.

zingolib/src/wallet/transactions/recording.rs
```diff
@@ -42,67 +35,9 @@ impl TxMapAndMaybeTrees {
         }
     }
 
-    /// Invalidates all those transactions which were broadcast but never 'confirmed' accepted by a miner.
-    pub(crate) fn clear_expired_mempool(&mut self, latest_height: u64) {
```
fluidvanadium, line 46: i have confirmed that this function does not effect witness trees

fluidvanadium (approved):
amazing!

PR #995, opened 2024-04-25
Allow caller to explicitly name stipulations, factor logic into predicate fn for readability

zancas:


fluidvanadium (approved):
excellente

PR #1008, opened 2024-04-26
Propose

fluidvanadium:
needs #1006 and #1007 and error typing

### DIFF
--- a/zingolib/src/lightclient/send.rs
+++ b/zingolib/src/lightclient/send.rs
@@ -1,5 +1,5 @@
 //! TODO: Add Mod Description Here!
-use std::num::NonZeroU32;
+use std::{num::NonZeroU32, ops::DerefMut};
 
 use log::debug;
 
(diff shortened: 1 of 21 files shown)


zingolib/src/lightclient/send.rs
```diff
@@ -30,11 +66,44 @@ impl LightClient {
...
+            ChainType,
+            GISKit,
+            ZingoLibError,
+        >(
+            tmamt.deref_mut(),
+            &self.wallet.transaction_context.config.chain,
+            zcash_primitives::zip32::AccountId::ZERO,
+            &input_selector,
+            request,
+            NonZeroU32::new(1).expect("yeep yop"), //review! be more specific
```
AloeareV:
```suggestion
            NonZeroU32::MIN
```
Would do this without the expect. For non-one numbers, here's also the possibility of moving the panic into a helper `const fn` and that way would only be hit at compile time

AloeareV (changes requested):

AloeareV (approved):
Errors need work, but that can be a follow-up pr. We now have a proposal being made!

PR #1010, opened 2024-04-29
Propose plus thiserror

fluidvanadium:
propose mod includes
  - DoProposeError
  - do_propose
  - do_propose_shielding

### DIFF
--- a/zingolib/src/lightclient/send.rs
+++ b/zingolib/src/lightclient/send.rs
@@ -51,6 +51,15 @@ pub fn receivers_becomes_transaction_request(
     TransactionRequest::new(payments)
 }
 
+use thiserror::Error;
+#[derive(Debug, Error)]
+pub enum DoProposeError {
+    #[error("{0}")]
+    Receiver(zcash_client_backend::zip321::Zip321Error),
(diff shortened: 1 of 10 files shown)


zingolib/src/lightclient/send.rs
```diff
@@ -255,4 +142,22 @@ impl LightClient {
 
         result.map(|(transaction_id, _)| transaction_id)
     }
+
+    #[cfg(feature = "zip317")]
+    /// Unstable function to expose the zip317 interface for development
+    // TODO: add correct functionality and doc comments / tests
+    pub async fn do_send_proposal(&self) -> Result<Vec<TxId>, String> {
```
AloeareV, line 149: The feature gates and imports would be cleaner if this went into the propose mod.

AloeareV (changes requested):

zingolib/src/lightclient/send.rs
fluidvanadium: its not part of that mod. propose mod is for creating the proposal. send mod includes sending it.

AloeareV (approved):

PR #1014, opened 2024-04-29
Read sapling params

fluidvanadium:
lives on #962


zingolib/src/lightclient.rs
```diff
@@ -536,6 +533,7 @@ impl LightClient {
         }
 
         // Ensure that the sapling params are stored on disk properly as well. Only on desktop
```
zancas, line 535: Is this doc-comment still correct?

zancas: I guess it's not ios/Android.

zingolib/src/lightclient/save.rs
```diff
@@ -36,28 +48,22 @@ impl LightClient {
         Ok(())
     }
 
-    /// If possible, write to disk. todo conditionally compile this function
```
zancas, line 39: Is this comment roughly the inspiration for this PR?

zancas (approved):

PR #1026, opened 2024-05-01
Add default zingoconfig

zancas:


zingoconfig/src/lib.rs
```diff
@@ -506,6 +506,13 @@ pub enum ChainType {
     Mainnet,
 }
 
+impl Default for ChainType {
```
AloeareV, line 509: My gut would expect a default ChainType to be Mainnet. This (and by extension, the default zingoconfig) strikes me as counterintuitive.

AloeareV (changes requested):

PR #1043, opened 2024-05-09
Error display

fluidvanadium:
improved error display.
impl Display for NoteRef -> ProposeError can be fully displayed

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,14 +17,14 @@ members = [
 resolver = "2"
 
 [workspace.dependencies]
-zcash_address = { git = "https://github.com/zingolabs/librustzcash.git", tag = "zcash_client_backend-0.12.1-1-g11ce156b_avoid_wallet_write" } # git tag -s $(git describe --dirty)
-zcash_client_backend = { git = "https://github.com/zingolabs/librustzcash.git", tag = "zcash_client_backend-0.12.1-1-g11ce156b_avoid_wallet_write", features = ["lightwalletd-tonic", "orchard", "transparent-inputs"] }
-zcash_encoding = { git = "https://github.com/zingolabs/librustzcash.git", tag = "zcash_client_backend-0.12.1-1-g11ce156b_avoid_wallet_write" }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -17,14 +17,14 @@ members = [
 resolver = "2"
 
 [workspace.dependencies]
-zcash_address = { git = "https://github.com/zingolabs/librustzcash.git", tag = "zcash_client_backend-0.12.1-1-g11ce156b_avoid_wallet_write" } # git tag -s $(git describe --dirty)
-zcash_client_backend = { git = "https://github.com/zingolabs/librustzcash.git", tag = "zcash_client_backend-0.12.1-1-g11ce156b_avoid_wallet_write", features = ["lightwalletd-tonic", "orchard", "transparent-inputs"] }
-zcash_encoding = { git = "https://github.com/zingolabs/librustzcash.git", tag = "zcash_client_backend-0.12.1-1-g11ce156b_avoid_wallet_write" }
-zcash_keys = { git = "https://github.com/zingolabs/librustzcash.git", tag = "zcash_client_backend-0.12.1-1-g11ce156b_avoid_wallet_write", features = ["orchard"] }
+zcash_address = { git = "https://github.com/zingolabs/librustzcash.git", rev = "84469ad71a3f389aaf839e2d4c40e612fa8b64a0" }
```
zancas, line 20: I think we should create and use human-readable `tags` in our dev manifest.

zancas (changes requested):
Perhaps the inclusion of Cargo.toml was accidental.   I am -1 on opaque commit hashes in the manifest.

Cargo.toml
fluidvanadium: what do you want me to call it?

fluidvanadium:
No. the Cargo.toml update is necessary to improve error display

Cargo.toml
zancas: I like to incorporate the output of `git describe --dirty` into tags..  of course there's a recursion issue...  IIRC

PR #1047, opened 2024-05-09
Generic test pooltype

fluidvanadium:


zancas (approved):
I am going to tinker with proptest in integration-tests

PR #1050, opened 2024-05-10
Unit test propose shield

zancas:


zingolib/src/lightclient/propose.rs
```diff
@@ -260,7 +260,7 @@ impl LightClient {
             // make it configurable?
             0,
         )
-        .map_err(ProposeShieldError::ShieldProposal)?;
+        .map_err(ProposeShieldError::Component)?;
```
fluidvanadium, line 263: not sure about this name

fluidvanadium (approved):

zingolib/src/lightclient/propose.rs
zancas: Me neither, but now it's not colliding with a type name.

PR #1053, opened 2024-05-10
Zip317 unify propose for test and interface

Oscar-Pepper:
follow on from #1052. Aims to unify testing and command interface once again by having clear layers of conversion where we are converting rust primitive types (i.e. str and u64s) to receivers in the testing/interface layer and then creating transaction requests from the receivers in the proposal layer. This will also help ease the switchover from do_send to propose when zip317 goes online as the inputs are aligned

- reworks raw_to_transaction_request to create a transaction request from receivers
- moves this function to within the propose lightclient methods
- removes uneccessary duplication of code associated with converting raw types and its corresponding error types
- aligns outputs of commands with inputs of propose lightclient methods

### DIFF
--- a/zingolib/src/lightclient/propose.rs
+++ b/zingolib/src/lightclient/propose.rs
@@ -11,7 +11,6 @@ use zcash_client_backend::zip321::Zip321Error;
 use zcash_client_backend::ShieldedProtocol;
 use zcash_keys::address::Address;
 use zcash_primitives::memo::MemoBytes;
-use zcash_primitives::transaction::components::amount::BalanceError;
 use zcash_primitives::transaction::components::amount::NonNegativeAmount;
 
 use thiserror::Error;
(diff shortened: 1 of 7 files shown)


zingolib/src/lightclient/propose.rs
```diff
@@ -273,6 +228,26 @@ impl LightClient {
         Ok(proposal)
     }
 }
+
+/// Creates a [`zcash_client_backend::zip321::TransactionRequest`] from receivers.
+fn transaction_request_from_receivers(
+    receivers: Vec<(Address, NonNegativeAmount, Option<MemoBytes>)>,
+) -> Result<TransactionRequest, ProposeSendError> {
```
fluidvanadium: how about this returns a zip321 error

zingolib/src/lightclient/propose.rs
```diff
@@ -120,17 +70,21 @@ pub enum ProposeShieldError {
...
+
     /// Unstable function to expose the zip317 interface for development
     // TOdo: add correct functionality and doc comments / tests
     // TODO: Add migrate_sapling_to_orchard argument
     pub(crate) async fn propose_send(
         &self,
-        request: TransactionRequest,
+        receivers: Vec<(Address, NonNegativeAmount, Option<MemoBytes>)>,
     ) -> Result<TransferProposal, ProposeSendError> {
+        let request = transaction_request_from_receivers(receivers)?;
```
fluidvanadium: and this has a map_err

fluidvanadium (changes requested):
please dont break https://github.com/zingolabs/zingolib/pull/1055

fluidvanadium (approved):
okay. still not sure about how i feel about _test_only(

PR #1063, opened 2024-05-10
Untested complete_and_broadcast impl

AloeareV:


zingolib/src/lightclient/send.rs
```diff
@@ -133,16 +133,18 @@ impl LightClient {
 #[cfg(feature = "zip317")]
 /// patterns for newfangled propose flow
 pub mod send_with_proposal {
-    use std::convert::Infallible;
+    use std::{convert::Infallible, ops::DerefMut as _};
```
fluidvanadium: can we just use DerufMut without the _?

AloeareV: We can, I like the convention of using `_` for traits that aren't named in the file, to indicate that they are imported for side effects (i.e. getting their trait methods into scope).

zingolib/src/lightclient/send.rs
```diff
@@ -199,6 +220,61 @@ pub mod send_with_proposal {
...
+            let (sapling_output, sapling_spend): (Vec<u8>, Vec<u8>) = self
+                .read_sapling_params()
+                .map_err(CompleteAndBroadcastError::SaplingParams)?;
+            let sapling_prover = LocalTxProver::from_bytes(&sapling_spend, &sapling_output);
+            let unified_spend_key =
+                UnifiedSpendingKey::try_from(self.wallet.wallet_capability().as_ref())
+                    .map_err(CompleteAndBroadcastError::UnifiedSpendKey)?;
+
+            let mut step_results = Vec::with_capacity(proposal.steps().len());
+            for step in proposal.steps() {
```
zancas: This is premature support for 320.   Let's remove it.

zingolib/src/lightclient/send.rs
```diff
@@ -199,6 +220,61 @@ pub mod send_with_proposal {
...
+                        step,
+                    )
+                    .map_err(CompleteAndBroadcastError::Calculation)?
+                };
+
+                step_results.push((step, step_result));
+            }
+            for (_step, step_result) in step_results {
+                self.wallet
+                    .send_to_addresses_inner(
```
fluidvanadium: we need to digest sent transaction BEFORE calculating the second proposal
AND ideally we ougta digest BEFORE broadcasting

fluidvanadium (changes requested):

zingolib/src/lightclient/send.rs
fluidvanadium: this is stable or upgrade

fluidvanadium (dismissed):
excellente!! incredible!

fluidvanadium:
oh no,  failed due to a merge conflict

zingolib/src/lightclient/send.rs
```diff
@@ -137,25 +137,49 @@ impl LightClient {
 #[cfg(feature = "zip317")]
 /// patterns for newfangled propose flow
 pub mod send_with_proposal {
```
Oscar-Pepper: i cant see this mod in the changelog, if its added all other pub in this mod will be covered implicitly

Oscar-Pepper (dismissed):

zingolib/src/lightclient/send.rs
AloeareV: Perhaps....but this PR adds no public methods to the interface.

fluidvanadium:
is `fulfill` a better word than 'complete' for changing a Proposal into a Transaction?

zingolib/src/lightclient/send.rs
Oscar-Pepper: well my point is it changes `pub enum CompleteAndBroadcastError` and therefore the quest of how do i update the changelog should resolve at "oops this whole mod was not added in the first place"

AloeareV: Oh right, the error is public. D'oh.

zancas (approved):

PR #1069, opened 2024-05-13
Trait name conduct chain

fluidvanadium:
/// A Lightclient test may involve hosting a server to send data to the LightClient. This trait can be asked to set simple scenarios where a mock LightServer sends data showing a note to a LightClient, the LightClient updates and responds by sending the note, and the Lightserver accepts the transaction and rebroadcasts it...  
/// The initial two implementors are
/// lib-to-node, which links a lightserver to a zcashd in regtest mode. see `impl ConductChain for LibtoNode`
/// darkside, a mode for the lightserver which mocks zcashd. search 'impl ConductChain for DarksideScenario`

### DIFF
--- a/darkside-tests/src/chain_generic_tests.rs
+++ b/darkside-tests/src/chain_generic_tests.rs
@@ -16,6 +16,28 @@ use crate::constants::DARKSIDE_SEED;
 use crate::utils::scenarios::DarksideEnvironment;
 use crate::utils::update_tree_states_for_transaction;
 
+#[tokio::test]
+#[ignore] // darkside cant handle transparent?
+async fn darkside_send_40_000_to_transparent() {
+    send_value_to_pool::<DarksideEnvironment>(40_000, Transparent).await;
+}
+
+proptest! {
+    #![proptest_config(proptest::test_runner::Config::with_cases(4))]
+    #[test]
+    fn send_pvalue_to_orchard(value in 0..90u32) {
+        Runtime::new().unwrap().block_on(async {
+    send_value_to_pool::<DarksideEnvironment>(value * 1_000, Shielded(Orchard)).await;
+        });
+     }
+    #[test]
+    fn send_pvalue_to_sapling(value in 0..90u32) {
+        Runtime::new().unwrap().block_on(async {
+    send_value_to_pool::<DarksideEnvironment>(value * 1_000, Shielded(Sapling)).await;
+        });
+     }
+}
+
 impl ConductChain for DarksideEnvironment {
     async fn setup() -> Self {
         DarksideEnvironment::new(None).await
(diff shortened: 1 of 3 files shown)


darkside-tests/src/chain_generic_tests.rs
```diff
@@ -16,7 +16,29 @@ use crate::constants::DARKSIDE_SEED;
 use crate::utils::scenarios::DarksideEnvironment;
 use crate::utils::update_tree_states_for_transaction;
 
-impl ManageScenario for DarksideEnvironment {
+#[tokio::test]
```
Oscar-Pepper, line 19: tests belong in the `tests` dir not `src`

integration-tests/tests/chain_generic_tests.rs
```diff
@@ -5,18 +5,31 @@ use zcash_client_backend::ShieldedProtocol::Orchard;
...
+    send_value_to_pool::<LibtonodeEnvironment>(40_000, Shielded(Orchard)).await;
+}
+
 struct LibtonodeEnvironment {
     regtest_network: RegtestNetwork,
     scenario_builder: ScenarioBuilder,
 }
 
-impl ManageScenario for LibtonodeEnvironment {
+impl ConductChain for LibtonodeEnvironment {
```
Oscar-Pepper, line 34: impls belong in test utils in `src` not `tests` directory

zingo-testutils/src/chain_generic_tests.rs
```diff
@@ -8,11 +8,63 @@ use zingolib::wallet::notes::query::OutputQuery;
...
 use zingolib::{get_base_address, wallet::notes::query::OutputPoolQuery};
 
+/// runs a send-to-receiver and receives it in a chain-generic context
+pub async fn propose_and_broadcast_value_to_pool<TE>(send_value: u32, pooltype: PoolType)
+where
+    TE: ConductChain,
+{
+    let mut environment = TE::setup().await;
+
+    dbg!("chain set up, funding client now");
```
Oscar-Pepper, line 18: I feel quite stongly about dbg! NEVER making it into dev. I use `println!` in test where i think there should be relavent information in the console when a test fails or runs with --no-capture

Oscar-Pepper: or even using logs and assigning it to debug level is a good compromise?

Oscar-Pepper (commented):
I am not requesting changes as technically they are not directly related to this PR but i think they are valid comments and should be addressed

integration-tests/tests/chain_generic_tests.rs
zancas: What does ConductChain mean?

zancas: Tests in `src` should be in unit-tests.

zancas (approved):

zingo-testutils/src/chain_generic_tests.rs
AloeareV: > or even using logs and assigning it to debug level is a good compromise?

Logs aren't viewable from tests, unfortunately. Agree that we shouldn't have dbg!s in non-test code.

PR #1071, opened 2024-05-13
Mod in chain generic

fluidvanadium:
This is is legibility PR. Set up some mods that will make moving trait implementations between files easier later. Now the use statements wont get tangled between the trait and the tests.

### DIFF
--- a/darkside-tests/src/chain_generic_tests.rs
+++ b/darkside-tests/src/chain_generic_tests.rs
@@ -7,14 +7,8 @@ use zcash_client_backend::ShieldedProtocol::Orchard;
 use zcash_client_backend::ShieldedProtocol::Sapling;
 
 use zingo_testutils::chain_generic_tests::send_value_to_pool;
-use zingo_testutils::chain_generic_tests::ConductChain;
-use zingolib::lightclient::LightClient;
-use zingolib::wallet::WalletBase;
 
-use crate::constants::ABANDON_TO_DARKSIDE_SAP_10_000_000_ZAT;
-use crate::constants::DARKSIDE_SEED;
 use crate::utils::scenarios::DarksideEnvironment;
-use crate::utils::update_tree_states_for_transaction;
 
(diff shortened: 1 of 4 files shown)


integration-tests/tests/chain_generic_tests.rs
Oscar-Pepper, line 1: effort has gone into moving this code but not out of the `tests` directory ? this is test framework not a test itself so doesnt belong here

darkside-tests/src/chain_generic_tests.rs
```diff
@@ -42,83 +36,94 @@ proptest! {
 ///   - transparent sends do not work
 ///   - txids are regenerated randomly. zingo can optionally accept_server_txid
 /// these tests cannot portray the full range of network weather.
-impl ConductChain for DarksideEnvironment {
-    async fn setup() -> Self {
-        DarksideEnvironment::new(None).await
-    }
+pub(crate) mod impl_conduct_chain_for_darkside_environment {
```
Oscar-Pepper, line 39: I think symmetry is important, can we name this `darkside_environment` to match `libtonode_environment` which serves the same purpose of including the impl for conduct chain? this mod name seems too long and specific

Oscar-Pepper (changes requested):
my comments from previous PRs werent addressed and that code has been moved again but not to the right place

darkside-tests/src/chain_generic_tests.rs
fluidvanadium: well, this mod doesnt include the actual definition of the struct, so it is importantly different from the other

integration-tests/tests/chain_generic_tests.rs
fluidvanadium: i have created a mod. the mod will be easy to move later

Oscar-Pepper (approved):
i will organise this when we have time, nothing is broken here

PR #1073, opened 2024-05-13
Remove duplicate raw to transaction code and replace with helper

Oscar-Pepper:


fluidvanadium:
good and hoping i can merge first

zingolib/src/lightclient/test_features.rs
```diff
@@ -85,4 +82,15 @@ impl LightClient {
...
+
+    /// Creates a [`zcash_client_backend::zip321::TransactionRequest`] from rust primitives for simplified test writing.
+    pub fn transaction_request_from_send_inputs(
+        &self,
+        address_amount_memo_tuples: Vec<(&str, u64, Option<&str>)>,
+    ) -> TransactionRequest {
+        let receivers =
+            receivers_from_send_inputs(address_amount_memo_tuples, &self.config().chain);
+        transaction_request_from_receivers(receivers)
+            .expect("should be able to create a transaction request as receivers are valid.")
```
fluidvanadium: can we return an error value instead of expecting? its possible to get an error here.

fluidvanadium (dismissed):
changing u32 to u64 is correct. i forgot to factor in the satoshis when i did the calculation.

zingolib/src/lightclient/test_features.rs
fluidvanadium: valid receivers can still result in invalid request in some cases.

Oscar-Pepper: I dont have an issue with test only functions containing expects. if for example there were too many payments in a test, then a panic here would be enough information to correct the test. anyways I have added the return of the zip321 error for you

zancas (approved):

fluidvanadium (approved):
i like having transaction_request_from_send_inputs

fluidvanadium:
re-opened as #1078

PR #1074, opened 2024-05-13
Remove unwraps get client

AloeareV:


fluidvanadium:
i see fewer unwraps but still some. i am not sure what this code does so i am not qualified to review it yet

zancas (approved):

zancas (commented):
LGTM

zingo-netutils/src/lib.rs
```diff
@@ -84,43 +101,51 @@ impl GrpcConnector {
...
-                let uri = uri.clone();
                 let svc = tower::ServiceBuilder::new()
                     //Here, we take all the pieces of our uri, and add in the path from the Requests's uri
-                    .map_request(move |mut req: http::Request<tonic::body::BoxBody>| {
+                    .map_request(move |mut request: http::Request<tonic::body::BoxBody>| {
+                        let path_and_query = request
+                            .uri()
+                            .path_and_query()
+                            .cloned()
+                            .unwrap_or(PathAndQuery::from_static("/"));
```
zancas, line 137: Do we really want this default?

AloeareV: Not really, but map_request is infallible, and I don't know that there's an alternative. I wasn't able to find one in ~15 minutes of looking..not sure that it's worth spending more time on it now. The path and query comes from the name of the RPC being called, so it will only be missing in the case of the proto-generated code messing up somehow, is my understanding.

zancas: So..  don't use map_request?   What's correct behavior here?   Can I trick a client into connecting to the wrong resource?   Is this worse than a panic!?

zingo-netutils/src/lib.rs
```diff
@@ -84,43 +101,51 @@ impl GrpcConnector {
...
-                            .scheme(uri.scheme().unwrap().clone())
-                            .authority(uri.authority().unwrap().clone())
+                            .scheme(scheme.clone())
+                            .authority(authority.clone())
                             //here. The Request's uri contains the path to the GRPC sever and
                             //the method being called
-                            .path_and_query(req.uri().path_and_query().unwrap().clone())
+                            .path_and_query(path_and_query)
                             .build()
                             .unwrap();
```
zancas, line 145: Here's another `unwrap`..

zingo-netutils/src/lib.rs
zancas:
```
let path_and_query = match request.uri().path_and_query().cloned() {
          Err(e) => panic!("Holy CRAP, this means that our gRPC generator is borken!!")
          Ok(great) => great,
};
```

zancas: But..   now define an explicit error variant and return that, instead of panic!?

zingo-netutils/src/lib.rs
```diff
@@ -84,43 +101,51 @@ impl GrpcConnector {
...
-                let uri = uri.clone();
                 let svc = tower::ServiceBuilder::new()
                     //Here, we take all the pieces of our uri, and add in the path from the Requests's uri
-                    .map_request(move |mut req: http::Request<tonic::body::BoxBody>| {
+                    .map_request(move |mut request: http::Request<tonic::body::BoxBody>| {
+                        let path_and_query = request
+                            .uri()
+                            .path_and_query()
+                            .cloned()
+                            .unwrap_or(PathAndQuery::from_static("/"));
```
zancas, line 111: Why can't these be new error variants?

PR #1081, opened 2024-05-15
Assertions mod

fluidvanadium:
these are assertions about whether a proposal was fulfilled, packaged up nicely

### DIFF
--- a/zingolib/src/wallet/notes/query.rs
+++ b/zingolib/src/wallet/notes/query.rs
@@ -38,13 +38,13 @@ impl OutputSpendStatusQuery {
 pub struct OutputPoolQuery {
     /// will the query include transparent notes? (coins)
     #[getset(get = "pub")]
-    transparent: bool,
+    pub transparent: bool,
     /// will the query include sapling notes?
     #[getset(get = "pub")]
-    sapling: bool,
+    pub sapling: bool,
     /// will the query include orchard notes?
     #[getset(get = "pub")]
-    orchard: bool,
+    pub orchard: bool,
 }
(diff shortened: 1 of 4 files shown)


zingolib/src/wallet/notes/query.rs
```diff
@@ -38,13 +38,13 @@ impl OutputSpendStatusQuery {
 pub struct OutputPoolQuery {
     /// will the query include transparent notes? (coins)
     #[getset(get = "pub")]
-    transparent: bool,
+    pub transparent: bool,
```
Oscar-Pepper, line 41: i dont understand why the pub getset isnt good enough and how making these pub relates to this PR? i probably have missed something

Oscar-Pepper (dismissed):

zingolib/src/wallet/notes/query.rs
fluidvanadium: its so that we can use naive struct instantiation

fluidvanadium: oh, i guess i didnt end up using it in this branch ...

AloeareV: Initializing an `OutputPoolQuery` with struct literal notation, i.e. 
```rust
let opq = OutputPoolQuery {
    transparent: true,
    sapling: true,
    orchard: false,
}
```
is clearer to me than something like `OutputPoolQuery::new(true, true, false)` or `OutputPoolQuery::default().set_transparent(true)...`

zancas (approved):

zancas (approved):

PR #1084, opened 2024-05-15
upgraded zip317 user interface to a much cleaner send and confirm flow

Oscar-Pepper:
Problem statement: 
While updating the command interface to recent updates such as the addition of zip317 shielding, it become clear the user interface of zingo-cli would be unclear and also cumbersome with long commands such as "proposesendall". There is also other issues of clarity such as proposing and then confirming a shield would require the user to call "proposeshield" followed by "send" which breaks the concept of a transfer being a "shield" OR a "send".
Although it seems logical to seperate a transfer into a "proposal" and "complete and broadcast" in the codebase, this seems unnecessarily confusing for a user of zingo-cli who wants to send funds. This is made worse by the fact that 3 commands all create proposals so we cant have a clear "propose" command.

Solution:
the solution i propose in this PR is that from a user interface perspective they should call "send" or "shield" followed by "confirm". Therefore, "send" or "shield" (or "sendall") will create and store a proposal via. propose_send / propose_shield etc. and display the important parts of this proposal to the user, prompting them to type "confirm" if they want to complete the transfer. Then, "confirm" will complete and broadcast the latest proposal. this also aligns well with zingo-mobile as clicking "send" will call the "send" command and then clicking "confirm" will call the "confirm" command

This PR also completes incomplete/unimplemented commands such as quickshield and adds a zip317 shield command

### DIFF
--- a/zingolib/CHANGELOG.md
+++ b/zingolib/CHANGELOG.md
@@ -41,7 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `wallet::keys::is_shielded_address` takes a `&ChainType` instead of a `&ZingoConfig`
 - `wallet::transaction_record_map::TransactionRecordMap` -> `wallet::transaction_records_by_id::TransactionRecordsById`
 - `commands`:
-  - `get_commands` added propose and quicksend to entries behind "zip317" feature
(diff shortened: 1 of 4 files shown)


fluidvanadium (approved):
awesome! i will tell juanky

juanky201271:
> awesome! i will tell juanky

Thanks, I think I got it. But any explanation will be well received

PR #1093, opened 2024-05-16
the type information is in the type system

zancas:
Use the ident to explain higher layer semantics

### DIFF
--- a/zingolib/src/lightclient.rs
+++ b/zingolib/src/lightclient.rs
@@ -696,6 +696,6 @@ mod tests {
 #[cfg(feature = "lightclient-deprecated")]
 mod deprecated;
 
-/// TODO: Add Doc Comment Here!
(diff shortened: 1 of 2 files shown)


Oscar-Pepper (dismissed):
I think this looks cleaner but it is not clear from the function definition what the &str's and u64's represent. please can you add to the doc comments to explain what the three parts of a raw receiver is in the correct order.

also if we are calling these `raw receivers` maybe we could go further and change the function names from `send_inputs` to `raw_receivers` i.e. `send_from_raw_receivers` ?

zancas:
I intend to go further with this refactor in other PRs.   I want to keep the scope of this to this name change to make sure that we're on the same page, and to keep review lightweight.

fluidvanadium (approved):
seems like requested changes have been made

PR #1095, opened 2024-05-16
Deprecate do_send

zancas:
This causes the integration tests to run against quicksend (wrapped by send).   I get 26 test fails.  I am looking into them now.

### DIFF
--- a/zingolib/src/lightclient/send.rs
+++ b/zingolib/src/lightclient/send.rs
@@ -1,15 +1,6 @@
 //! TODO: Add Mod Description Here!
-use log::debug;
 
-use zcash_client_backend::address::Address;
 use zcash_primitives::consensus::BlockHeight;
-use zcash_primitives::transaction::fees::zip317::MINIMUM_FEE;
-use zcash_primitives::transaction::TxId;
-use zcash_proofs::prover::LocalTxProver;
-
-use crate::data::receivers::Receivers;
-use crate::utils::conversion::zatoshis_from_u64;
-use crate::wallet::Pool;
 
 use super::LightClient;
 use super::LightWalletSendProgress;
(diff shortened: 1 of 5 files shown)


zingolib/src/lightclient/send.rs
```diff
@@ -23,35 +22,23 @@ impl LightClient {
         ) + 1)
     }
 
+    #[deprecated(note = "superceded by quick_send")]
```
Oscar-Pepper: my opinion is that do_send should be left intact and deprecated when zip317 is fully tested and a viable alternative. as far as testing goes we dont use do_send we use send_from_send_inputs (a wrapper on do_send), and this function should be replaced by quick_send_from_send_inputs throughout. if do_send is heading towards deprecation it makes sense to make this switch in our testing now anyway

zingolib/src/lightclient/send.rs
```diff
@@ -23,35 +22,23 @@ impl LightClient {
...
-            )
-            .await
+        let request = match crate::data::receivers::transaction_request_from_receivers(receivers) {
+            Ok(req) => req,
+            Err(e) => {
+                return Err(format!(
+                    "Error: {}\nTry 'help send' for correct usage and examples.",
+                    e
+                ));
+            }
```
Oscar-Pepper: for what its worth, the correct way of handling our errors moving forward is returning an error struct/enum rather than a string so these "Error: {} yada yadas" are only concocted in the commands.rs / top layer and the Display::fmt implemented on these error types will format the error message

Oscar-Pepper (commented):
I have made a couple of comments about our switchover strategy.
I am going to open a new PR that purely switches over send_from_send_inputs to quick_send_from_send_inputs and starts correcting fee errors etc.
when we are left with tests that are truly failing we can merge that PR into our feature branchs temporarily to check if we have correctly fixed them. eventually, once all tests pass we can merge the PR

zancas:
This now deletes `do_send` and `do_shield`.

PR #1098, opened 2024-05-16
Merge Confirmation status mod

fluidvanadium:
This should allow crates that already depend on zingolib to explicitly name the ConfirmationStatus enum.
Also organized a bunch of headers


zancas (changes requested):
There's a lot of change here that's not necessary for the re-homing of `confirmation_status`.

I'd prefer to isolate that re-homing into it's own PR.

PR #1096, opened 2024-05-16
zip317_update_integration_tests

Oscar-Pepper:
Replaces all calls to do_shield and do_send with quick_send and quick_shield in all integration tests.

This PR should be merged into dev when all test pass. This draft PR can be used as a tool to temp merge into a feature branch to check if it is correctly fixing a failing test

this PR should not include any fixes to the codebase but it should include fixes to wrong assertions in tests such as wrong expected fees etc.

PR #1100, opened 2024-05-16
tests run including zingolib unit-tests without test-features flag

zancas:
Builds on #1101

This PR continues the work begun in #1101.

The intent is to fully separate unit test utilities from integration test utilities.

A subsequent PR is expected to mv and rename the "test_features" code into `zingo-testutils`.


zingolib/src/lightclient/propose.rs
```diff
@@ -132,12 +130,13 @@ impl LightClient {
     /// Unstable function to expose the zip317 interface for development
     // TOdo: add correct functionality and doc comments / tests
     // TODO: Add migrate_sapling_to_orchard argument
+    #[cfg(test)]
```
Oscar-Pepper, line 134: this is not intended to be a test function this is intended to expose the send_all interface to juanky for zip317 development. we can comment this out like the others if you prefer

zingolib/src/lightclient/test_features.rs
```diff
@@ -4,22 +4,53 @@
...
+use zingoconfig::ChainType;
+
+use crate::data::receivers::Receivers;
+
+/// Converts primitive rust types to zcash receiver types for constructing transaction requests.
+///
+/// # Panics
+///
+/// Panics if the address, amount or memo conversion fails.
+pub fn receivers_from_send_inputs(
```
Oscar-Pepper, line 30: the reason why I created a conversion module is to home functions like this that should not belong in lightclient because they are not a method of lightclient. this doesnt seem right to me maybe we shuld discuss

zingolib/src/lightclient/test_features.rs
```diff
@@ -4,22 +4,53 @@
 //! raw_receiver.0:   A &str representing the receiver address
 //! raw_receiver.1:   A u64 representing the number of zats to be sent
 //! raw_receiver.2:   An Option<&str> that contains memo data if not None
```
Oscar-Pepper, line 6: im not sure who wrote this and when it merged but this comment is wrong. a raw_receiver is only used in integration testing not from the command line. also maybe "stuple" is a spelling mistake?

zingolib/src/lightclient/test_features.rs
```diff
@@ -33,28 +64,6 @@ impl LightClient {
...
-    pub async fn quick_send_from_send_inputs(
-        &self,
-        raw_receivers: Vec<(&str, u64, Option<&str>)>,
-    ) -> Result<NonEmpty<TxId>, QuickSendError> {
-        let request = self
-            .transaction_request_from_send_inputs(raw_receivers)
-            .expect("should be able to create a transaction request as receivers are valid.");
-        self.quick_send(request).await
-    }
-
```
Oscar-Pepper, line 57: these two functions are necessary and im not sure why they are being deleted?
#1096 replaces send_from_send_inputs with quick_send_from_send_inputs throughout all integ tests to test zip317

Oscar-Pepper (changes requested):
i have left some comments we need to discuss before this can merge

zingolib/src/lightclient/propose.rs
zancas: I think @juanky201271 hid the send-all function for the first release of ZIP317.

zancas: Commented.

zingolib/src/lightclient/test_features.rs
zancas: `stuple` decapitated

zancas: e.g. removed

zingolib/src/lightclient/test_features.rs
zancas: I agree that it is an odd duck.   I am OK with it moving more.   I just moved it because I knew it was only in test.   In a subsequent PR I moved it into zingo-testutils..   I am not clear on where it will ultimately live.   I am a fan of putting where it's used.

zingolib/src/lightclient/test_features.rs
zancas: My bad.   I followed your suggestion and added _all_ necessary functionality to zingo-testutils/src/lightclient.rs

Oscar-Pepper (approved):
nice work

PR #1102, opened 2024-05-17
complete removal of test-features gate

zancas:
This builds on #1100 and #1101 to complete the removal of the test-features gate from the codebase.

This results in a clean separation between integration test utilities and unit-test utilities.

Some tradeoffs:   There're more pub interfaces in this feature, and that may be a blocker for merge.

My primary motivation was a desire to have a clear understanding of where different kinds of functionality would live.   Logic that is for the purpose of testing a full lightclient operating in an environment should not be sourced in `zingolib`.


zingo-testutils/src/lightclient.rs
```diff
@@ -0,0 +1,90 @@
...
+        });
+        shielder
+            .do_shield(pools_to_shield, address)
+            .await
+            .map(|txid| txid.to_string())
+    }
+
+    /// Creates a [`zcash_client_backend::zip321::TransactionRequest`] from rust primitives for simplified test writing.
+    pub fn transaction_request_from_send_inputs(
+        requester: &zingolib::lightclient::LightClient,
```
Oscar-Pepper: maybe i prefer `sender` as its more clear to me but non-blocking. I'm pretty sure this previously had ChainType as the input which I liked because you shouldnt need to create an entire lightclient to create a transaction request in testing but I understand the organisation here and im not going to block on this either

zingo-testutils/src/lightclient.rs
```diff
@@ -0,0 +1,90 @@
...
+        PoolType::Shielded(ShieldedProtocol::Orchard) => {
+            client.do_addresses().await[0]["address"].take().to_string()
+        }
+    }
+}
+/// Helpers to provide raw_receivers to lightclients for send and shield, etc.
+pub mod from_inputs {
+    use zingolib::lightclient::LightClient;
+
+    /// Panics if the address, amount or memo conversion fails.
```
Oscar-Pepper, line 45: doc comment seems incomplete to me, I thought i actually wrote a full comment for this that seems to have got lost

darkside-tests/src/utils.rs
```diff
@@ -674,10 +674,12 @@ pub mod scenarios {
                 DarksideSender::IndexedClient(n) => self.get_lightclient(n),
                 DarksideSender::ExternalClient(lc) => lc,
             };
-            lightclient
-                .send_from_send_inputs(vec![(receiver_address, value, None)])
-                .await
-                .unwrap();
+            zingo_testutils::lightclient::from_inputs::send(
```
Oscar-Pepper, line 677: I like your creativity to move lightclient test methods into zingo-testutils in terms of code organisation and simplification, although I think including the full path on every send/shield looks awful and is crazy. I am very happy in test writing to actually just write `send` with `use::zingo_testutils::from_inputs::send` imported. that will actually clean up our test code rather than turn `recipient.send` into a monster on many lines of test code

zingolib/src/lightclient/propose.rs
```diff
@@ -132,12 +130,13 @@ impl LightClient {
     /// Unstable function to expose the zip317 interface for development
     // TOdo: add correct functionality and doc comments / tests
     // TODO: Add migrate_sapling_to_orchard argument
+    #[cfg(test)]
     pub async fn propose_send_all(
         &self,
-        _address: Address,
-        _memo: Option<MemoBytes>,
+        _address: zcash_keys::address::Address,
+        _memo: Option<zcash_primitives::memo::MemoBytes>,
```
Oscar-Pepper: I intentionally wrote this function with implicit imports because it looks much better to me, i dont like code that looks unnecesarrily complex or convoluted. my opinion is that we shouldnt have our own Address or MemoBytes structs, we should use LRZ's types, and should be very familiar with LRZ types, so when we see `Address` it is already ingrained this is an LRZ address type. I think the best way for our split opinions on this to live together is i wont change your functions and you dont change mine haha

zingolib/src/utils/conversion.rs
```diff
@@ -29,54 +29,13 @@ pub fn txid_from_hex_encoded_str(txid: &str) -> Result<TxId, TxIdFromHexEncodedS
     Ok(TxId::from_bytes(txid_bytes))
 }
 
-pub(crate) fn address_from_str(
-    address: &str,
-    chain: &ChainType,
-) -> Result<Address, ConversionError> {
+/// Coverrt a &str to an Adddress
```
Oscar-Pepper: typo -> Covert a &str to an Address

zingolib/src/utils/conversion.rs
```diff
@@ -29,54 +29,13 @@ pub fn txid_from_hex_encoded_str(txid: &str) -> Result<TxId, TxIdFromHexEncodedS
...
+pub fn address_from_str(address: &str, chain: &ChainType) -> Result<Address, ConversionError> {
     Address::decode(chain, address)
         .ok_or_else(|| ConversionError::InvalidAddress(address.to_string()))
 }
 
-pub(crate) fn zatoshis_from_u64(amount: u64) -> Result<NonNegativeAmount, ConversionError> {
+/// Have a u64?  Need Zatoshis?
```
Oscar-Pepper: haha plus one humor, minus one on looking profesh

Oscar-Pepper (changes requested):

zingo-testutils/src/lightclient.rs
zancas: Happy to add that back in as needed.

darkside-tests/src/utils.rs
zancas: I agree with you.   I like:

```Rust
use::zingo_testutils::from_inputs;
...
from_inputs::FOO(spam...
```

I am aiming to consistently apply that pattern moving forward.

zingo-testutils/src/lightclient.rs
zancas: I moved some commentary to the top of the mod, and removed the repetition of `test` and `panic` since panic is explicitly still in the remaining doc-comment, and the code now lives in a crate called `testutils`.

zingolib/src/lightclient/propose.rs
zancas: I see your point.  I don't remember with certainty why I made this change, but I suspect it was that I was moving the function and didn't want to add-and-delete "`use FOO`" in **addition** to moving the fn itself.

I agree with your assertion that we should use LRZ types wherever possible.

zingolib/src/utils/conversion.rs
```diff
@@ -29,14 +29,13 @@ pub fn txid_from_hex_encoded_str(txid: &str) -> Result<TxId, TxIdFromHexEncodedS
...
-    chain: &ChainType,
-) -> Result<Address, ConversionError> {
+/// Convert a &str to an Adddress
+pub fn address_from_str(address: &str, chain: &ChainType) -> Result<Address, ConversionError> {
     Address::decode(chain, address)
         .ok_or_else(|| ConversionError::InvalidAddress(address.to_string()))
 }
 
-pub(crate) fn zatoshis_from_u64(amount: u64) -> Result<NonNegativeAmount, ConversionError> {
+/// Convery a valid u64 into Zatoshis.
```
Oscar-Pepper: I like it. typo though .. lets see if i can spell `convert` properly. You passed the test and corrected my correction on `address_from_str`

Oscar-Pepper (changes requested):

Oscar-Pepper (approved):

PR #1105, opened 2024-05-18
Remove derive_more

zancas:
Remove unnecessary type, and use explicit type construction to expose meaningful field names at call sites.


zancas:
Superseded by #1109

fluidvanadium (approved):
good catch, thanks!

PR #1114, opened 2024-05-20
Ignore Shielded Dust

fluidvanadium:
ignores dust inputs. passes the ignore_dust_inputs test.
prevents the sufficient insufficient fee bug

builds on #1111 and #1113


zingolib/src/wallet/transaction_records_by_id/trait_inputsource.rs
```diff
@@ -168,6 +175,9 @@ impl InputSource for TransactionRecordsById {
                     Some(missing_value_after_sapling),
                     |rolling_target, (note, note_id)| match rolling_target {
                         Some(targ) if targ == NonNegativeAmount::ZERO => Ok(None),
+                        Some(targ) if note.value().inner() <= MARGINAL_FEE.into_u64() => {
+                            Ok(Some(targ))
```
zancas, line 185: What if the spend is cross-pool?   The e.g. 7_000 (> MARGINAL) is still dust if it's transparent to shielded..

fluidvanadium: this is a really good question. it is asking for a good discussion

fluidvanadium: this function is only for shielded pools

fluidvanadium:
superceded by #1142

PR #1124, opened 2024-05-23
DRY mock implementation with declarative macro

zancas:


fluidvanadium (approved):
reduces lines of code!!

PR #1125, opened 2024-05-23
Zip317 send all

Oscar-Pepper:
ontop of #1185 

- propose_send_all lightclient method for creating and storing a proposal for sending total confirmed shielded balance
- added send all tests including fail cases
- send all command
- simple helpers for calculating total amount and fee from proposal with tests

a following PR with helperise spendable balance and add a spendable balance command to protect accidental "sending all funds when trying to find spendable balance" bugs


zingolib/src/commands.rs
```diff
@@ -892,11 +893,13 @@ impl Command for SendCommand {
             }
         };
         RT.block_on(async move {
-            match lightclient
-                .propose_send(request)
-                .await {
+            match lightclient.propose_send(request).await {
```
zancas, line 896: I don't understand why a Command named "Send" is actually a proposal creator.

zingolib/src/data/proposal.rs
```diff
@@ -25,3 +29,53 @@ pub(crate) enum ZingoProposal {
...
+        })
+}
+
+#[cfg(test)]
+mod tests {
+    use zcash_primitives::transaction::components::amount::NonNegativeAmount;
+
+    use crate::mocks;
+
+    #[test]
```
zancas, line 65: Nice unit tests.

zancas (approved):

PR #1127, opened 2024-05-23
Assertions 3: shield with assertions

fluidvanadium:
builds on #1126

adds a version of shield that calls asserts intended for test cases

PR #1129, opened 2024-05-23
Assertions 40: recipient balance

fluidvanadium:
builds on #1127 

checks that recipient has recieved the expected balance


zingo-testutils/src/assertions.rs
```diff
@@ -33,3 +32,31 @@ pub async fn assert_send_outputs_match_client<NoteId>(
         );
     }
 }
+
+/// currently only checks if the fee matches
+/// this currently fails for any broadcast but not confirmed transaction: it seems like get_transaction_fee does not recognize pending spends
```
zancas, line 37: Let's always use "pending" when we want an adjective that modifies a transaction, and reserve the word "broadcast" to be used as a verb, e.g. in the name of the function "propose_and_broadcast".

zancas (approved):

PR #1131, opened 2024-05-23
Assertions 400: expected fee

fluidvanadium:
based on #1129
-expected fee is now asserted to be equal to
-proposed fee
-recorded fee

### DIFF
--- a/zingo-testutils/src/chain_generic_tests.rs
+++ b/zingo-testutils/src/chain_generic_tests.rs
@@ -89,7 +89,7 @@ pub mod fixtures {
         let expected_fee = MARGINAL_FEE.into_u64()
             * match pooltype {
                 Transparent => 3,
-                Shielded(Sapling) => 4,
+                Shielded(Sapling) => 4, // but according to my reading of https://zips.z.cash/zip-0317, this should be 3
                 Shielded(Orchard) => 2,
             };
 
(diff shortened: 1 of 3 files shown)


zingo-testutils/src/chain_generic_tests.rs
```diff
@@ -89,7 +89,7 @@ pub mod fixtures {
         let expected_fee = MARGINAL_FEE.into_u64()
             * match pooltype {
                 Transparent => 3,
-                Shielded(Sapling) => 4,
+                Shielded(Sapling) => 4, // but according to my reading of https://zips.z.cash/zip-0317, this should be 3
                 Shielded(Orchard) => 2,
             };
```
Oscar-Pepper, line 108: this expected fee calculator cannot return the correct fee in this state, lets discuss at codeshare but i think we should have a dedicated PR that adds an expected fee calculator to zingo-testutils

Oscar-Pepper (changes requested):

zingo-testutils/src/chain_generic_tests.rs
fluidvanadium: Out of date...

fluidvanadium: This instance of expected fee is correct in this micro context, which is good enough for this test for now

zancas (approved):

PR #1135, opened 2024-05-25
rename, and doc-comment, and elide unused methods

zancas:
# Rationale:

Rename components of the TransactionRecord to emphasize how they are related to the TransactionRecord.

This significantly clarifies the code.

# Comments

Input (in this context) means "input of value to the TransactionRecord".

By highlighting that the Record is a Capability-Specific dataset, I hope further clarity is gained.

By marking the differential publicity of different pieces of data I hope to add clarity, but also to suggest an enhancement in the form of new language-level categorization of this data. 

Some of the elided methods might need to be re-added when the `pub` of the record is removed or diminished, but I think it's better to "burn that bridge when we come to it" rather than carry extra kruft.

### DIFF
--- a/zingolib/src/wallet/transaction_record.rs
+++ b/zingolib/src/wallet/transaction_record.rs
@@ -25,47 +25,67 @@ use crate::{
     },
 };
 
-///  Everything (SOMETHING) about a transaction
+/// This Capabilities perspective on a Transaction
+/// Key:   [ChainPublic|Outgoing|Incoming]
+///   ChainPublic: this data can be obtained directly from the chain, it has no privacy properties
+///   Outgoing:    this data can be Viewed by the Capability that createed this Transaction
+///   Incoming:    this data can be Viewed by the Recipient Capability that is encumbered by the note
+///  NOTE: A received note on an outgoing transaction was formerly called "change".
 #[derive(Debug)]
 pub struct TransactionRecord {
     /// the relationship of the transaction to the blockchain. can be either Broadcast (to mempool}, or Confirmed.
+    ///  # ChainPublic
     pub status: zingo_status::confirmation_status::ConfirmationStatus,
 
     /// Timestamp of Tx. Added in v4
+    ///  # ChainPublic
     pub datetime: u64,
 
     /// Txid of this transaction. It's duplicated here (It is also the Key in the HashMap that points to this
-    /// WalletTx in LightWallet::txs)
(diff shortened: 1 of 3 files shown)


Oscar-Pepper:
NAck: comments in codeshare for discussion

zingolib/src/wallet/transaction_record.rs
```diff
@@ -25,47 +25,67 @@ use crate::{
...
     /// Txid of this transaction. It's duplicated here (It is also the Key in the HashMap that points to this
-    /// WalletTx in LightWallet::txs)
+    /// TransactionRecord in LightWallet::TransactionRecordsById)
+    ///  # ChainPublic
     pub txid: TxId,
 
-    /// List of all nullifiers spent by this wallet in this Tx.
-    pub spent_sapling_nullifiers: Vec<sapling_crypto::Nullifier>,
+    /// List of all Sapling nullifiers used by this Capability to add value to the Sapling Value Pool.
+    ///  # Outgoing
```
Oscar-Pepper, line 53: lets have a discussion about this PR, i have left some comments in chat. Outgoing here highlights the type of confusion we might unleash by defining inputs as nullifiers, (and only nullifiers). nullifiers are actually public on chain so should fall into the category of public. but i suppose there is confusion here that the value comes from a note/output only seen by the "Outgoing" capability. however, a nullifier is public and a link to that output, not the output itself or the value it holds

Oscar-Pepper (changes requested):
I think this PR needs discussion

zingolib/src/wallet/transaction_record.rs
fluidvanadium: i like the clarity that these nullifiers are inputs to THIS transaction

fluidvanadium (dismissed):
+4 for illustrative clear comments

zingolib/src/wallet/transaction_record.rs
zancas: The 32-byte number (IIRC) nullifier value is public, but which output it nullifies is **critically** private.   So I agree that there's  room to improve these doc-comments, but I think we should merge and iterate on them.

Oscar-Pepper: sorry i dont agree with merging new code that is confusing/incorrect on the basis we can change it later. lets change it now. nullifiers are public and when we add an extra field to transaction records that stores a ref to the notes spent in this transaction we can call that field Outgoing

PR #1136, opened 2024-05-25
Update value transfer and txsummaries

zancas:
Builds on #1135 
Preamble to debugging:

https://github.com/zingolabs/zingolib/blob/dev/zingolib/src/lightclient/describe.rs#L249-L284


Oscar-Pepper (changes requested):
if you can cherry-pick the commit(s) that just changes do_list_txsummaries on dev i will approve

PR #1139, opened 2024-05-27
Proptest libtonode

fluidvanadium:
This tests different values in the libtonode tests


Oscar-Pepper (dismissed):

Oscar-Pepper (changes requested):
NAck.
Proptest enforcing to move integration tests to src/lib.rs makes it pretty clear to me that proptests are for unit tests.

PR #1141, opened 2024-05-27
Transparent dust guard

fluidvanadium:
From a user perspective, this effects what happens when they Propose Shielding while they have few transparent funds

If they would create a note with value <= SHIELDING_CUTOFF, stop. propose -> error
SHIELDING_CUTOFF = 5_000

### DIFF
--- a/zingo-testutils/src/assertions.rs
+++ b/zingo-testutils/src/assertions.rs
@@ -68,6 +68,7 @@ pub async fn assert_cant_shield(client: &zingolib::lightclient::LightClient) {
                 println!("available {}, required more than {}", available.into_u64(), required.into_u64());
                 true
             },
+            zingolib::lightclient::propose::ProposeShieldError::Dusty => true,
             e => {
                 dbg!(e);
                 false
--- a/zingo-testutils/src/assertions.rs
+++ b/zingo-testutils/src/assertions.rs
@@ -65,7 +65,7 @@ pub async fn assert_send_outputs_match_recipient<NoteId>(
 pub async fn assert_cant_shield(client: &zingolib::lightclient::LightClient) {
     assert!(match client.propose_shield().await.unwrap_err() {
             zingolib::lightclient::propose::ProposeShieldError::Component(zcash_client_backend::data_api::error::Error::NoteSelection(zcash_client_backend::data_api::wallet::input_selection::GreedyInputSelectorError::Change(zcash_client_backend::fees::ChangeError::InsufficientFunds { available, required }))) => {
-                println!("available {}, required {}", available.into_u64(), required.into_u64());
+                println!("available {}, required more than {}", available.into_u64(), required.into_u64());
                 true
             },
             e => {
--- a/zingo-testutils/src/assertions.rs
+++ b/zingo-testutils/src/assertions.rs
@@ -69,7 +69,7 @@ pub async fn assert_cant_shield(client: &zingolib::lightclient::LightClient) {
                 true
             },
             zingolib::lightclient::propose::ProposeShieldError::BelowThreshold(amount) => {
-                println!("possible note {}, threshold {}", amount, zingolib::lightclient::propose::SHIELDING_THRESHOLD);
+                println!("possible note {}, threshold {}", amount, zingolib::lightclient::propose::SHIELDING_CUTOFF);
                 true},
             e => {
                 dbg!(e);
--- a/zingo-testutils/src/assertions.rs
+++ b/zingo-testutils/src/assertions.rs
@@ -62,7 +62,7 @@ pub async fn assert_send_outputs_match_recipient<NoteId>(
 }
 
 /// assert that proposing shield causes a specific error
-pub async fn assert_cant_shield(client: zingolib::lightclient::LightClient, balance: u64) {
+pub async fn assert_cant_shield(client: &zingolib::lightclient::LightClient, balance: u64) {
     assert!(match client.propose_shield().await.unwrap_err() {
             zingolib::lightclient::propose::ProposeShieldError::Component(zcash_client_backend::data_api::error::Error::NoteSelection(zcash_client_backend::data_api::wallet::input_selection::GreedyInputSelectorError::Change(zcash_client_backend::fees::ChangeError::InsufficientFunds { available, required }))) => {
                 println!("available {}, required {}", available.into_u64(), required.into_u64());
--- a/zingo-testutils/src/assertions.rs
+++ b/zingo-testutils/src/assertions.rs
@@ -68,7 +68,9 @@ pub async fn assert_cant_shield(client: &zingolib::lightclient::LightClient) {
                 println!("available {}, required more than {}", available.into_u64(), required.into_u64());
                 true
             },
-            zingolib::lightclient::propose::ProposeShieldError::Dusty => true,
+            zingolib::lightclient::propose::ProposeShieldError::BelowThreshold(amount) => {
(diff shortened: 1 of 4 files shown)


zingo-testutils/src/chain_generic_tests.rs
```diff
@@ -128,6 +129,92 @@ pub mod fixtures {
...
+                    (Transparent, 1_000),
+                    (Transparent, 1_000),
+                    (Transparent, 1_000),
+                    (Transparent, 1_000),
+                    (Transparent, 1_000),
+                    (Transparent, 1_000)
+                ],
+            )
+            .await,
+            40_000
```
Oscar-Pepper: can we atleast have these in multiples of marginal fee so the reviewer can count easier. I have said in many other PRs I dont think we should have magic number expected fees. we need an expected fee calculator that reviewed and used thourghout our testing

zingolib/src/mocks.rs
```diff
@@ -1,5 +1,3 @@
-#![allow(dead_code)]
```
Oscar-Pepper: see PR House 1

Oscar-Pepper (changes requested):

zingo-testutils/src/chain_generic_tests.rs
fluidvanadium: addressed

fluidvanadium: haha, 'magic number expected fees'

fluidvanadium: good point, though there is also an internal assert that proposed = recorded

fluidvanadium: fixed slightly

zingo-testutils/src/assertions.rs
```diff
@@ -60,3 +60,20 @@ pub async fn assert_send_outputs_match_recipient<NoteId>(
...
+    assert!(match client.propose_shield().await.unwrap_err() {
+            zingolib::lightclient::propose::ProposeShieldError::Component(zcash_client_backend::data_api::error::Error::NoteSelection(zcash_client_backend::data_api::wallet::input_selection::GreedyInputSelectorError::Change(zcash_client_backend::fees::ChangeError::InsufficientFunds { available, required }))) => {
+                println!("available {}, required more than {}", available.into_u64(), required.into_u64());
+                true
+            },
+            zingolib::lightclient::propose::ProposeShieldError::BelowThreshold(amount) => {
+                println!("possible note {}, threshold {}", amount, zingolib::lightclient::propose::SHIELDING_CUTOFF);
+                true},
+            e => {
+                dbg!(e);
```
Oscar-Pepper, line 75: i think we should use println! for test printouts that are meant to be permenant useful data when a test fails

zingo-testutils/src/assertions.rs
```diff
@@ -60,3 +60,20 @@ pub async fn assert_send_outputs_match_recipient<NoteId>(
...
+/// assert that proposing shield causes a specific error
+pub async fn assert_cant_shield(client: &zingolib::lightclient::LightClient) {
+    assert!(match client.propose_shield().await.unwrap_err() {
+            zingolib::lightclient::propose::ProposeShieldError::Component(zcash_client_backend::data_api::error::Error::NoteSelection(zcash_client_backend::data_api::wallet::input_selection::GreedyInputSelectorError::Change(zcash_client_backend::fees::ChangeError::InsufficientFunds { available, required }))) => {
+                println!("available {}, required more than {}", available.into_u64(), required.into_u64());
+                true
+            },
+            zingolib::lightclient::propose::ProposeShieldError::BelowThreshold(amount) => {
+                println!("possible note {}, threshold {}", amount, zingolib::lightclient::propose::SHIELDING_CUTOFF);
+                true},
```
Oscar-Pepper, line 73: I think asserting that a specific expected error is the correct way to test error cases, this test will pass when the shield fails in a way that wasnt expected.

i like the pattern:
assert!(matches!(client.propose_shield().await, Err(zingolib::lightclient::propose::ProposeShieldError::BelowThreshold(_))))

this way we dont need to waste time trying to write the error messages out and fixing all tests when the message changes etc.

this is my preference but I think tests are more readable when this assert is directly in the test itself rather than helperised as this format is quite concise anyway

zingo-testutils/src/chain_generic_tests.rs
```diff
@@ -128,6 +129,92 @@ pub mod fixtures {
...
+            )
+            .await,
+            3 * MARGINAL_FEE.into_u64()
+        );
+
+        // the recipient cannot propose shielding
+        assertions::assert_cant_shield(&shielder).await;
+
+        // creates an output for recipient which combined with the previous is sendable, but would create dust: still not enough to be worth a shield
+        let nondust_2 = 14_000u64;
```
Oscar-Pepper, line 182: maybe im getting this wrong but if all dust is ignored, non dust notes sum = 6_000 + 14_000 = 20_000
logical actions = 2 for 2 t-note spends plus 1 for orchard action = 3
fee = 3 * 5_000 = 15_000
received note value = 20_000 - 15_000 = 5_000 which is *not* dust
unless im missing something can we change the values to make the note 3_000 or something?

also a caveat is the orchard actions may pad to 2 which makes the received note value 0, still need to clarify this as I haven't found this in the proposal code but in testing this seems to occur

zingolib/src/lightclient/propose.rs
```diff
@@ -211,9 +219,26 @@ impl LightClient {
...
+                std::cmp::max(biggest_change_value_value, change_value.value().into_u64())
+            });
+
+        if biggest_change > SHIELDING_CUTOFF {
+            self.store_proposal(ZingoProposal::Shield(proposal.clone()))
+                .await;
+            Ok(proposal)
+        } else {
+            Err(ProposeShieldError::BelowThreshold(biggest_change))
+        }
```
Oscar-Pepper, line 241: none of this makes sense to me. due to how @AloeareV constructed the `value_to_shield` in the ShieldCommand It is my understanding that fees for shielding can't be paid for from the transparent pool. therefore, if we *were* to check the value we are shielding we should do it the same way we do in commands.rs, by summing the transparent inputs. However ..... we have a shielding threshold on `propose_shielding` which means we don't shield dust anyway .. so this is a no-op right?

Oscar-Pepper (changes requested):

zingolib/src/lightclient/propose.rs
Oscar-Pepper: i have just found that `propose_shield` also calls the same `compute_balance` so if we want to prevent dust change we can just rely on the `DustAction::Reject` policy. this doesnt need re-implementing

Oscar-Pepper: > none of this makes sense to me. due to how @AloeareV constructed the `value_to_shield` in the ShieldCommand It is my understanding that fees for shielding can't be paid for from the transparent pool. therefore, if we _were_ to check the value we are shielding we should do it the same way we do in commands.rs, by summing the transparent inputs. However ..... we have a shielding threshold on `propose_shielding` which means we don't shield dust anyway .. so this is a no-op right?

looking at the shield code in LRZ my assumptions were wrong and the shielding amount in commands.rs must also be incorrect. There are no shielded inputs added in a shielding proposal and therefore all fees must be covered by transparent inputs.

Oscar-Pepper:
test and test framework to be implemented in a suitably named following PR
@fluidvanadium

PR #1143, opened 2024-05-27
Select Spendable Streamlined

fluidvanadium:
This is a rewrite of select_spendable_notes, using what we have learned over the past couple of weeks.

Notable improvement is that the ignore_dust test goes from passing to failing for zip317

### DIFF
--- a/zingo-testutils/src/chain_generic_tests.rs
+++ b/zingo-testutils/src/chain_generic_tests.rs
@@ -216,50 +216,42 @@ pub mod fixtures {
         CC: ConductChain,
     {
         let mut environment = CC::setup().await;
+
         let primary = environment.fund_client_orchard(120_000).await;
-
-        let primary_address_orchard = get_base_address(&primary, Shielded(Orchard)).await;
-
         let secondary = environment.create_client().await;
-        // let secondary_address_sapling = secondary.get_base_address(Shielded(Sapling)).await;
-        let secondary_address_sapling = get_base_address(&secondary, Shielded(Sapling)).await;
-        let secondary_address_orchard = get_base_address(&secondary, Shielded(Orchard)).await;
 
-        from_inputs::send(
-            &primary,
-            vec![
-                (secondary_address_sapling.as_str(), 1_000, None),
-                (secondary_address_orchard.as_str(), 1_000, None),
-                (secondary_address_sapling.as_str(), 1_000, None),
-                (secondary_address_orchard.as_str(), 1_000, None),
-                (secondary_address_sapling.as_str(), 1_000, None),
-                (secondary_address_orchard.as_str(), 1_000, None),
-                (secondary_address_sapling.as_str(), 1_000, None),
-                (secondary_address_orchard.as_str(), 1_000, None),
-                (secondary_address_sapling.as_str(), 1_000, None),
-                (secondary_address_orchard.as_str(), 1_000, None),
-                (secondary_address_sapling.as_str(), 10_000, None),
-                (secondary_address_orchard.as_str(), 10_000, None),
-            ],
-        )
-        .await
-        .unwrap();
+        assert_eq!(
+            with_assertions::propose_send_bump_sync_recipient(
+                &mut environment,
+                &primary,
+                &secondary,
+                vec![
+                    (Shielded(Sapling), 1_000),
+                    (Shielded(Sapling), 1_000),
+                    (Shielded(Sapling), 1_000),
+                    (Shielded(Sapling), 1_000),
+                    (Shielded(Sapling), 15_000),
+                    (Shielded(Orchard), 1_000),
+                    (Shielded(Orchard), 1_000),
+                    (Shielded(Orchard), 1_000),
+                    (Shielded(Orchard), 1_000),
+                    (Shielded(Orchard), 15_000),
+                ],
+            )
+            .await,
+            11 * MARGINAL_FEE.into_u64()
+        );
 
-        environment.bump_chain().await;
-        secondary.do_sync(false).await.unwrap();
-
-        check_client_balances!(secondary, o: 15_000 s: 15_000 t: 0);
-
-        from_inputs::send(
-            &secondary,
-            vec![(primary_address_orchard.as_str(), 5_001, None)],
-        )
-        .await
-        .unwrap();
-
-        environment.bump_chain().await;
-        secondary.do_sync(false).await.unwrap();
-        check_client_balances!(secondary, o: 9_999 s: 5_000 t: 0);
+        assert_eq!(
+            with_assertions::propose_send_bump_sync_recipient(
+                &mut environment,
+                &secondary,
+                &primary,
+                vec![(Shielded(Orchard), 10_000),],
+            )
+            .await,
+            4 * MARGINAL_FEE.into_u64()
+        );
     }
 
     /// creates a proposal, sends it and receives it (upcoming: compares that it was executed correctly) in a chain-generic context
--- a/zingo-testutils/src/chain_generic_tests.rs
+++ b/zingo-testutils/src/chain_generic_tests.rs
@@ -220,6 +220,7 @@ pub mod fixtures {
         let primary = environment.fund_client_orchard(120_000).await;
         let secondary = environment.create_client().await;
 
+        // send a bunch of dust
         assert_eq!(
             with_assertions::propose_send_bump_sync_recipient(
                 &mut environment,
@@ -242,6 +243,7 @@ pub mod fixtures {
             11 * MARGINAL_FEE.into_u64()
         );
 
+        // combine the only valid sapling note with the only valid orchard note to send
         assert_eq!(
             with_assertions::propose_send_bump_sync_recipient(
                 &mut environment,
--- a/zingolib/src/wallet/transaction_record.rs
+++ b/zingolib/src/wallet/transaction_record.rs
@@ -401,7 +401,7 @@ impl TransactionRecord {
                 .into_iter()
                 .find(|note| *note.output_index() == Some(id.output_index().into()))
                 .map(|zingo_orchard_note| {
-                    crate::data::notes::Note::Orchard(zingo_orchard_note.note().clone())
+                    crate::data::notes::Note::Orchard(*zingo_orchard_note.note())
                 })
             }
         }
--- a/zingolib/src/wallet/transaction_record.rs
+++ b/zingolib/src/wallet/transaction_record.rs
@@ -381,6 +381,7 @@ impl TransactionRecord {
         })
     }
 
+    /// get a Note from a NoteId
     pub(crate) fn get_note_from_id(&self, id: NoteId) -> Option<crate::data::notes::Note> {
         match id.protocol() {
             zcash_client_backend::ShieldedProtocol::Sapling => {
--- a/zingolib/src/wallet/transaction_record.rs
+++ b/zingolib/src/wallet/transaction_record.rs
@@ -825,7 +825,6 @@ mod tests {
     use proptest::prelude::proptest;
     use test_case::test_matrix;
 
-    use orchard::note_encryption::OrchardDomain;
     use sapling_crypto::note_encryption::SaplingDomain;
     use zcash_client_backend::wallet::NoteId;
     use zcash_client_backend::ShieldedProtocol::{Orchard, Sapling};
--- a/zingolib/src/wallet/transaction_record.rs
+++ b/zingolib/src/wallet/transaction_record.rs
@@ -21,10 +21,8 @@ use crate::{
         data::{OutgoingTxData, PoolNullifier, COMMITMENT_TREE_LEVELS},
         keys::unified::WalletCapability,
         notes::{
-            self,
-            query::{OutputQuery, OutputSpendStatusQuery},
-            OrchardNote, OutputId, OutputInterface, SaplingNote, ShieldedNoteInterface,
-            TransparentOutput,
+            query::OutputQuery, OrchardNote, OutputId, OutputInterface, SaplingNote,
+            ShieldedNoteInterface, TransparentOutput,
         },
         traits::{DomainWalletExt, ReadableWriteable as _},
     },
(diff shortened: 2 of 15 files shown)


zingolib/src/wallet/transaction_record.rs
```diff
@@ -379,6 +347,44 @@ impl TransactionRecord {
...
+        if sources.contains(&Sapling) {
+            self.sapling_notes.iter().for_each(|zingo_sapling_note| {
+                if zingo_sapling_note.is_unspent() {
+                    if let Some(output_index) = zingo_sapling_note.output_index() {
+                        let id = NoteId::new(self.txid, Sapling, *output_index as u16);
+                        if !exclude.contains(&id) {
+                            all.push((id, zingo_sapling_note.value()));
+                        }
+                    } else {
+                        println!("note has no index");
```
zancas, line 367: What case does this match?

zingolib/src/wallet/transaction_records_by_id/trait_inputsource.rs
```diff
@@ -118,89 +114,75 @@ impl InputSource for TransactionRecordsById {
...
-                        .filter(|note_ref_pair| !exclude.contains(&note_ref_pair.1)),
-                );
+        let mut unselected =
+            self.get_spendable_note_ids_and_values(sources, anchor_height, exclude);
+
+        unselected.sort_by_key(|(_id, value)| *value);
+
+        let mut selected = vec![];
+
+        let mut unselected_iterator = 0;
```
zancas: Let's call this `counter`?  Iterator is too overloaded.

zingolib/src/wallet/transaction_records_by_id/trait_inputsource.rs
```diff
@@ -118,89 +114,75 @@ impl InputSource for TransactionRecordsById {
...
+                    // the iterator went off the end of the vector without finding a note big enough to complete the transaction... add the biggest note and reset the iteraton
+                    selected.push(unselected.pop().expect("nonempty"));
+                    unselected_iterator = 0;
+                    continue;
+                }
+                Some(smallest_unselected) => {
+                    // selected a note to test if it has enough value to complete the transaction on its own
+                    if smallest_unselected.1
+                        >= target_value.into_u64()
+                            - selected.iter().fold(0, |sum, (_id, value)| sum + *value)
```
zancas, line 150: It would be nice if this were a local variable with a clear name like:

`already_selected_value`  or..  something.

zingolib/src/wallet/transaction_records_by_id/trait_inputsource.rs
```diff
@@ -118,89 +114,75 @@ impl InputSource for TransactionRecordsById {
...
+                    {
+                        selected.push(*smallest_unselected);
+                        unselected.remove(unselected_iterator);
```
zancas: This line hints that `unselected_iterator` is better named `index_into_unselected`.

zingolib/src/wallet/transaction_records_by_id/trait_inputsource.rs
```diff
@@ -118,89 +114,75 @@ impl InputSource for TransactionRecordsById {
...
-                        - NonNegativeAmount::from_u64(note.value().inner()).map_err(|e| {
-                            InputSourceError::OutputTooBig((note.value().inner(), e))
-                        })?)
+        let mut selected_sapling = Vec::<ReceivedNote<NoteId, sapling_crypto::Note>>::new();
+        let mut selected_orchard = Vec::<ReceivedNote<NoteId, orchard::Note>>::new();
+        selected
+            .iter()
+            .try_for_each(|(id, _value)| match id.protocol() {
+                zcash_client_backend::ShieldedProtocol::Sapling => {
+                    match self.get(id.txid()).and_then(|transaction_record| {
```
zancas: Helperize?

zingolib/src/wallet/transaction_records_by_id/trait_inputsource.rs
```diff
@@ -118,89 +114,75 @@ impl InputSource for TransactionRecordsById {
...
-                                        )
-                                    })
-                                    .ok_or(InputSourceError::NoteCannotBeIdentified(note_id))?,
-                            );
-                            Ok(targ
-                                - NonNegativeAmount::from_u64(note.value().inner()).map_err(
-                                    |e| InputSourceError::OutputTooBig((note.value().inner(), e)),
-                                )?)
+                zcash_client_backend::ShieldedProtocol::Orchard => {
+                    match self.get(id.txid()).and_then(|transaction_record| {
```
zancas: For DRYness?

zancas (changes requested):
I left some notes.

zingolib/src/wallet/transaction_records_by_id/trait_inputsource.rs
fluidvanadium: i chose to write the comment on line 138 instead, as a stylistic choice to separated the mathematical aspects from the human explanation

fluidvanadium: i will write a good explanation for this loop

zingolib/src/wallet/transaction_records_by_id/trait_inputsource.rs
fluidvanadium: oh, i can use the transitive property to simplify this

zingolib/src/wallet/transaction_record.rs
fluidvanadium: loaded an old wallet with incomplete information about notes... when we update wallet format we will do targeted rescan.

zingolib/src/wallet/transaction_records_by_id/trait_inputsource.rs
fluidvanadium: dried from 33 to 21 lines

zingo-testutils/src/chain_generic_tests.rs
```diff
@@ -216,50 +216,42 @@ pub mod fixtures {
...
+                    (Shielded(Sapling), 15_000),
+                    (Shielded(Orchard), 1_000),
+                    (Shielded(Orchard), 1_000),
+                    (Shielded(Orchard), 1_000),
+                    (Shielded(Orchard), 1_000),
+                    (Shielded(Orchard), 15_000),
+                ],
+            )
+            .await,
+            11 * MARGINAL_FEE.into_u64()
```
Oscar-Pepper, line 243: 11 marginal fees suggests the change from all these payments our returned in one note? I wasnt aware of that.

zingo-testutils/src/chain_generic_tests.rs
```diff
@@ -216,50 +216,42 @@ pub mod fixtures {
...
+                    (Shielded(Sapling), 1_000),
+                    (Shielded(Sapling), 1_000),
+                    (Shielded(Sapling), 1_000),
+                    (Shielded(Sapling), 1_000),
+                    (Shielded(Sapling), 15_000),
+                    (Shielded(Orchard), 1_000),
+                    (Shielded(Orchard), 1_000),
+                    (Shielded(Orchard), 1_000),
+                    (Shielded(Orchard), 1_000),
+                    (Shielded(Orchard), 15_000),
```
Oscar-Pepper, line 239: how does the faucet have enough notes for this send? interesting

zingo-testutils/src/chain_generic_tests.rs
```diff
@@ -216,50 +216,42 @@ pub mod fixtures {
...
-        check_client_balances!(secondary, o: 9_999 s: 5_000 t: 0);
+        assert_eq!(
+            with_assertions::propose_send_bump_sync_recipient(
+                &mut environment,
+                &secondary,
+                &primary,
+                vec![(Shielded(Orchard), 10_000),],
+            )
+            .await,
+            4 * MARGINAL_FEE.into_u64()
```
Oscar-Pepper, line 255: orchard -> orchard = 1 logical action
orchard change pads to 1 logical action
fee = 2 logical actions * marginal fee

why 4???

zingolib/src/wallet/transaction_record.rs
```diff
@@ -379,6 +347,44 @@ impl TransactionRecord {
             })
         })
     }
+
+    /// get a list of unspent NoteIds with associated note values
+    pub(crate) fn get_spendable_note_ids_and_values(
```
Oscar-Pepper, line 352: much better name IMO. nice

zingolib/src/wallet/transaction_records_by_id/trait_inputsource.rs
```diff
@@ -25,20 +25,15 @@ use zcash_client_backend::wallet::NoteId;
 use zcash_primitives::transaction::components::amount::BalanceError;
 
 /// TODO: Add Doc Comment Here!
+#[allow(missing_docs)] // error types document themselves
```
Oscar-Pepper: oh interesting, because of thiserror?

zingolib/src/wallet/transaction_records_by_id/trait_inputsource.rs
```diff
@@ -118,89 +122,70 @@ impl InputSource for TransactionRecordsById {
...
-                        .into_iter()
-                        .filter(|note_ref_pair| !exclude.contains(&note_ref_pair.1)),
-                );
+            match unselected.get(index_of_unselected) {
+                None => {
+                    // the iterator went off the end of the vector without finding a note big enough to complete the transaction... add the biggest note and reset the iteraton
+                    selected.push(unselected.pop().expect("nonempty"));
+                    index_of_unselected = 0;
+                    continue;
+                }
```
Oscar-Pepper, line 145: there is sometimes cases where it makes sense to put None before Some but this doesnt seem right to me

zingolib/src/wallet/transaction_records_by_id/trait_inputsource.rs
```diff
@@ -118,89 +122,70 @@ impl InputSource for TransactionRecordsById {
...
+                    }),
+                zcash_client_backend::ShieldedProtocol::Orchard => transaction
+                    .and_then(|transaction_record| {
+                        transaction_record.get_received_note::<OrchardDomain>(output_index)
+                    })
+                    .map(|received_note| {
+                        selected_orchard.push(received_note);
+                    }),
+            }
+            .ok_or(InputSourceError::NoteCannotBeIdentified(*id))
```
Oscar-Pepper, line 186: this works but looks messy to me, could we handle each option and return an error in place to avoid using `and_then` and `map` and then returning an Option<()> ? I think it would be much more readable
```suggestion
                    // transform each NoteId to a ReceivedNote
        for (id, _value) in selected {
            let transaction = self
                .get(id.txid())
                .ok_or(InputSourceError::NoteCannotBeIdentified(id))?;
            let output_index = id.output_index() as u32;
            match id.protocol() {
                zcash_client_backend::ShieldedProtocol::Sapling => {
                    let received_note = transaction
                        .get_received_note::<SaplingDomain>(output_index)
                        .ok_or(InputSourceError::NoteCannotBeIdentified(id))?;
                    selected_sapling.push(received_note);
                }
                zcash_client_backend::ShieldedProtocol::Orchard => {
                    let received_note = transaction
                        .get_received_note::<OrchardDomain>(output_index)
                        .ok_or(InputSourceError::NoteCannotBeIdentified(id))?;
                    selected_orchard.push(received_note);
                }
            }
        }

```

Oscar-Pepper (changes requested):
I have left some comments which i think are worth implementing but non-blocking. The reason why I am requesting changes is because our expected fees seem to just be inputted by the test writer with no real reviewed expected fee calculator and some of the expected fees seem like they might be wrong to me. ALSO the new select_spendable_notes has lost control of selecting notes based on pool which i think is more important than value as it effects our pool policies and also is needed to reduce fees for user. although I havent worked out how this is even possible with the current LRZ code as the use_orchard / use_sapling booleans dont seem to work correctly or get fed into this function, so non-blocking... maybe we should implement this soon though.

zingo-testutils/src/chain_generic_tests.rs
Oscar-Pepper: I imagine what is happening is because select_spendable_notes is not selecting by pool it is selecting a sapling note, therefore this should maybe be set to 2 * MARGINAL_FEE and the failing test should be fixed to correctly prioritise orchard?

zingolib/src/wallet/transaction_record.rs
fluidvanadium: well, its also a much better function. it doesnt have the same signature either

zingolib/src/wallet/transaction_records_by_id/trait_inputsource.rs
fluidvanadium: i like the pattern with and_then. updating identifier for clarity

zingolib/src/wallet/transaction_records_by_id/trait_inputsource.rs
fluidvanadium: I guess the argument for Some first is that we expect the first iteration of the loop to hit the Some case... however there are multiple ways to exit this loop
I normally put None first because it is a simpler case

fluidvanadium: in this case, i typically expect the None case to notes before the if condition in the Some case to break up the loop. so i like this way

zingolib/src/wallet/transaction_records_by_id/trait_inputsource.rs
fluidvanadium: oops! didnt work as i expected. fixing

fluidvanadium: fixed

fluidvanadium: the Display Implementation... though this doesnt work. fixed

zingo-testutils/src/chain_generic_tests.rs
fluidvanadium: uh, it gets funded by 1 big note, then sends off 11 to the secondary client

zingo-testutils/src/chain_generic_tests.rs
fluidvanadium: 10 outgoing outputs and 1 change i think

zingo-testutils/src/chain_generic_tests.rs
fluidvanadium: we have to select the sapling note to send this much

zancas (approved):

Oscar-Pepper (approved):
wonderful

PR #1144, opened 2024-05-27
update name from do_list_txsummaries to list_txsummaries

zancas:
Also, actually call do_list_transactions in macro


fluidvanadium (approved):
just what it says on the tin

PR #1148, opened 2024-05-28
House 1: Fix a Clippy Warning

fluidvanadium:


zingolib/src/mocks.rs
```diff
@@ -1,5 +1,3 @@
-#![allow(dead_code)]
```
Oscar-Pepper: I added this because it is going to be a common occurance that mock builders are created in a PR and then used in a following PR. if we remove this it will likely get added again soon so I dont see the point

Oscar-Pepper (changes requested):
Can you detail the compiler warning, I can't recreate it so I dont know what this is fixing or how this could be fixing anything

zingolib/src/mocks.rs
fluidvanadium: okay

fluidvanadium:
warning: methods with the following characteristics: (`to_*` and `self` type is not `Copy`) usually take `self` by reference
   --> zingolib/src/mocks.rs:8:22
    |
8   | ...   pub fn $name(&mut self, $name: $loc...
    |                    ^^^^^^^^^
    |
   ::: zingolib/src/wallet/data.rs:695:9
    |
695 |         build_method!(to_address, String);
    |         --------------------------------- in this macro invocation
    |
    = help: consider choosing a less ambiguous name
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention
    = note: `#[warn(clippy::wrong_self_convention)]` on by default
    = note: this warning originates in the macro `build_method` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: `zingolib` (lib test) generated 1 warning

fluidvanadium:
this means that to_address is an bad method name

fluidvanadium:
rewritten to use macro, changing field name

Oscar-Pepper (commented):
I much prefer `recipient_address` as this keeps in tie with the rest of our codebase

Oscar-Pepper (approved):
nice!

PR #1159, opened 2024-05-30
Confirmed balance excluding dust

Oscar-Pepper:
- adds new wallet balance method for determining confirmed shielded balance not counting dust (<5000 zats) necessary for send_all and therefore correct spendable balance
- adds unit test to verify correct function
- deprecates two wrongly named and unecessary balance functions
- switches the two test calls of these balance functions to `shielded_balance`
- default OrchardCryptoNoteBuilder now has a fixed recipient as its impossible for a wallet to have decrypted outputs from a different seed
- default TransactionRecordBuilder has random txid to avoid overwriting when pushing multiple defaults to TransactionRecordsById


fluidvanadium (approved):
specific function and unit tested. i approve

PR #1168, opened 2024-05-31
Self send transparent receipt

zancas:
This isn't ready to land.   Two legacy tests that have to do with promoting funds from before the heartwoord epoch fail.   I suspect that may have to do with the definition of the OVK.

Any help appreciated!

RUN:

`cargo nextest run -p libtonode-tests heartwood`

### DIFF
--- a/zingolib/src/wallet/transaction_context.rs
+++ b/zingolib/src/wallet/transaction_context.rs
@@ -87,12 +87,6 @@ pub mod decrypt_transaction {
             let mut is_outgoing_transaction = false;
             // Collect our t-addresses for easy checking
             let taddrs_set = self.key.get_all_taddrs(&self.config);
-            // Process t-address outputs
-            //  THE FOLLOWING LINE INTENDS TO SAY:  IF THERE EXISTS AT LESAT ONE RECEIVER THAT THE CREATING CAPABILITY DOES NOT CONTROL...
-            // If this transaction in outgoing, i.e., we received sent some money in this transaction, then we need to grab all transparent outputs
-            // that don't belong to us as the outgoing metadata
-            // the assumption is either we already decrypted a compact output and filled in some data
-            // or transparent something
 
             // in the send case, we already know the transaction is outgoing. however, this if clause will not trigger.
             let mut outgoing_metadatas = vec![];
--- a/zingolib/src/wallet/transaction_context.rs
+++ b/zingolib/src/wallet/transaction_context.rs
@@ -88,23 +88,13 @@ pub mod decrypt_transaction {
             // Collect our t-addresses for easy checking
             let taddrs_set = self.key.get_all_taddrs(&self.config);
             // Process t-address outputs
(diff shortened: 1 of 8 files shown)


zingolib/src/wallet/transaction_context.rs
```diff
@@ -87,23 +87,6 @@ pub mod decrypt_transaction {
...
-            if self
-                .transaction_metadata_set
-                .read()
-                .await
-                .transaction_records_by_id
-                .total_funds_spent_in(&transaction.txid())
-                > 0
-            {
-                is_outgoing_transaction = true;
-            }
```
Oscar-Pepper, line 106: to save me review time as you have a better understanding of the change here, could you point me to where `is_outgoing_transaction` is set true for shielded if this is removed?

zingolib/src/wallet/transaction_context.rs
```diff
@@ -124,12 +107,14 @@ pub mod decrypt_transaction {
                             .recipient_address()
                             .map(|raw_taddr| address_from_pubkeyhash(&self.config, raw_taddr))
                         {
-                            outgoing_metadatas.push(OutgoingTxData {
-                                recipient_address: taddr,
-                                value: u64::from(vout.value),
-                                memo: Memo::Empty,
-                                recipient_ua: None,
-                            });
+                            if !taddrs_set.contains(&taddr) {
```
Oscar-Pepper, line 110: yeah this single line looks like the bug fix we needed. nice

zingolib/src/wallet/transaction_context.rs
```diff
@@ -218,6 +203,11 @@ pub mod decrypt_transaction {
                 for (n, vout) in t_bundle.vout.iter().enumerate() {
                     if let Some(taddr) = vout.recipient_address() {
                         let output_taddr = address_from_pubkeyhash(&self.config, taddr);
+                        // If there's at least one output in this bundle that does *NOT* belong to the
+                        // Creator, then the transaction is defined as "is_outgoing"
+                        if !taddrs_set.contains(&output_taddr) {
+                            *is_outgoing_transaction = true;
+                        }
```
Oscar-Pepper, line 210: hmm what about the case where a different capability has sent to us and another recipient. in that case we would call this function and think we created it?

zingolib/src/wallet/transaction_context.rs
Oscar-Pepper: https://github.com/zingolabs/zingolib/pull/1168/files#diff-c059f59e2dd106815b6b903751dd212059da1c2e8b7a0cec71db6fbcc8a48b27R536-R541
as you can see in the comment we need to be careful messing around with the definition of outgoing_tx_data as we could land a serious bug and block release.

Oscar-Pepper (changes requested):
although I think ultimately we should rewrite alot of these definitions and a lot of code this PR touches, I am in favor of the aim of this PR as a step in the right direction. however, this must be thoroughly reviewed and scroutinised to avoid introducing release-blocking bugs. I think I have found a potential bug in review which makes me think we propbably also need some suitable tests to make sure we dont break anything

PR #1172, opened 2024-05-31
do not include self sent transparent notes in outgoing tx data

zancas:
This fixes 2 concrete libtonode tests


Oscar-Pepper (approved):
great work!

PR #1179, opened 2024-06-01
Remove inconsistent method

zancas:
Builds on #1178

PR #1180, opened 2024-06-01
Two doc comments

zancas:
Builds on #1179

### DIFF
--- a/zingolib/src/wallet/transaction_context.rs
+++ b/zingolib/src/wallet/transaction_context.rs
@@ -124,12 +124,14 @@ pub mod decrypt_transaction {
                             .recipient_address()
                             .map(|raw_taddr| address_from_pubkeyhash(&self.config, raw_taddr))
                         {
-                            outgoing_metadatas.push(OutgoingTxData {
-                                recipient_address: taddr,
-                                value: u64::from(vout.value),
-                                memo: Memo::Empty,
-                                recipient_ua: None,
-                            });
+                            if !taddrs_set.contains(&taddr) {
+                                outgoing_metadatas.push(OutgoingTxData {
+                                    recipient_address: taddr,
+                                    value: u64::from(vout.value),
+                                    memo: Memo::Empty,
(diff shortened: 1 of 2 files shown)


zingolib/src/wallet/transaction_context.rs
```diff
@@ -88,12 +88,9 @@ pub mod decrypt_transaction {
             // Collect our t-addresses for easy checking
             let taddrs_set = self.key.get_all_taddrs(&self.config);
             // Process t-address outputs
-            // If this transaction in outgoing, i.e., we received sent some money in this transaction, then we need to grab all transparent outputs
-            // that don't belong to us as the outgoing metadata
-            // the assumption is either we already decrypted a compact output and filled in some data
-            // or transparent something
-
-            // in the send case, we already know the transaction is outgoing. however, this if clause will not trigger.
+            // If the there's funding known to be input to this transaction by this Capability
```
Oscar-Pepper:
```suggestion
            // If there's funding known to be input to this transaction by this Capability
```

zingolib/src/wallet/transaction_records_by_id.rs
```diff
@@ -277,6 +277,8 @@ impl TransactionRecordsById {
             .sum();
         Ok(query_record.total_transparent_value_spent + sapling_spend_value + orchard_spend_value)
     }
+    // The value that's output, but *NOT* to an explicit receiver (unless this is running on the winning validator!)
+    // is the fee.
     fn total_value_output_to_explicit_receivers(&self, query_record: &TransactionRecord) -> u64 {
```
Oscar-Pepper: as fees are defined as [total input value - total output value] I am still not really onboard with this function name and therefore desciption. I would prefer something more to the point like:

/// total sum of the value of all outgoing transaction data plus the value of all outputs received as change
fn total_output_value()

however, in the next wallet version we may not need this so lets not spend toom much time on this discussion

Oscar-Pepper (changes requested):

PR #1181, opened 2024-06-01
Refactor tx summary matcher

zancas:
Builds on #1180

### DIFF
--- a/zingolib/src/lightclient/describe.rs
+++ b/zingolib/src/lightclient/describe.rs
@@ -1,17 +1,12 @@
 //! These functions can be called by consumer to learn about the LightClient.
+use ::orchard::note_encryption::OrchardDomain;
 use json::{object, JsonValue};
-use orchard::note_encryption::OrchardDomain;
 use sapling_crypto::note_encryption::SaplingDomain;
 use std::collections::HashMap;
 use tokio::runtime::Runtime;
 
 use zcash_address::ZcashAddress;
-use zcash_client_backend::{
-    encoding::encode_payment_address,
-    fees::{orchard, TransactionBalance},
-    proto::proposal::TransactionBalance,
-    PoolType, ShieldedProtocol,
-};
+use zcash_client_backend::{encoding::encode_payment_address, PoolType, ShieldedProtocol};
 use zcash_primitives::{
     consensus::{BlockHeight, NetworkConstants},
     memo::Memo,
@@ -29,12 +24,8 @@ use crate::{
             summaries::{ValueTransfer, ValueTransferKind},
             OutgoingTxData, TransactionRecord,
         },
-        error::FeeError,
         keys::address_from_pubkeyhash,
-        notes::{
-            query::{self, OutputQuery},
-            OutputInterface,
-        },
+        notes::{query::OutputQuery, OutputInterface},
         transaction_records_by_id::TransactionRecordsById,
         LightWallet,
     },
@@ -405,7 +396,7 @@ impl LightClient {
     pub fn get_server_uri(&self) -> http::Uri {
         self.config.get_lightwalletd_uri()
     }
-
+    // Given a transaction write down ValueTransfer information in a Summary list
     fn record_value_transfers(
         summaries: &mut Vec<ValueTransfer>,
         txid: TxId,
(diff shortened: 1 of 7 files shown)


zingolib/src/lightclient/describe.rs
```diff
@@ -375,104 +396,113 @@ impl LightClient {
...
+        transaction_records: &TransactionRecordsById,
+    ) -> Result<(), ValueTransferRecordingError> {
+        let is_received = match transaction_records.transaction_is_received(transaction_record) {
+            Ok(received) => received,
+            Err(fee_error) => {
+                return Err(ValueTransferRecordingError::FeeCalculationError(
+                    fee_error.to_string(),
+                ))
+            }
+        };
```
Oscar-Pepper, line 413: let-else statements are really nice to tidy up this `Ok(r) => r` fluff
```suggestion
        let Ok(is_received) = transaction_records.transaction_is_received(transaction_record) else {
                return Err(ValueTransferRecordingError::FeeCalculationError(
                    fee_error.to_string()
                ))
        };
```

zingolib/src/lightclient/describe.rs
```diff
@@ -20,15 +20,25 @@ use crate::{
...
+        transaction_records_by_id::TransactionRecordsById,
         LightWallet,
     },
 };
 
+#[allow(missing_docs)]
+#[derive(Debug, thiserror::Error)]
+pub enum ValueTransferRecordingError {
+    #[error("Fee was not calculable because of error:  {0}")]
+    FeeCalculationError(String),
```
Oscar-Pepper: rather than strings would we benefit from holding the actual error type? this is how I have done error stacks and have seen in other codebases. not sure what the pros and cons are though but I think we only lose information or potential error handling improvements by converting to string here

zingolib/src/lightclient/describe.rs
```diff
@@ -261,7 +271,11 @@ impl LightClient {
 
     /// Provides a list of value transfers related to this capability
     pub async fn list_txsummaries(&self) -> Vec<ValueTransfer> {
+        self.list_txsummaries_and_capture_errors().await.0
+    }
+    async fn list_txsummaries_and_capture_errors(&self) -> (Vec<ValueTransfer>, Vec<String>) {
         let mut summaries: Vec<ValueTransfer> = Vec::new();
+        let mut errors: Vec<String> = Vec::new();
```
Oscar-Pepper: `Vec<ValueTransferRecordingError>` ?

zingolib/src/lightclient/describe.rs
```diff
@@ -375,104 +396,113 @@ impl LightClient {
...
-                            txid,
-                            pending,
-                        });
-                    }
-                }
+        if is_received {
+            if !transaction_record.outgoing_tx_data.is_empty() {
+                return Err(ValueTransferRecordingError::IncoherentOutgoing(
+                    "transaction_record.outgoing_tx_data".to_string(), // TODO: Make into real data.
+                ));
```
Oscar-Pepper: if you add the presence of outgoing_tx_data to the `is_received` logic we would successfully add a `Sent` value transfer even when the inputs have not been synced yet

zingolib/src/lightclient/describe.rs
```diff
@@ -481,19 +511,23 @@ impl LightClient {
...
-            // We spent funds, and received funds as non-change. This is most likely a send-to-self,
-            // TODO: Figure out what kind of special-case handling we want for these
-            (true, true) => {
+            // If the transaction is outgoing, and there are notes in:
+            //   - sapling notes
+            //   - orchard notes
+            // then they were sent to self
+            if !transaction_record.sapling_notes.is_empty()
+                || !transaction_record.orchard_notes.is_empty()
+            {
```
Oscar-Pepper: this will create a `SendToSelf` value transfer for every change note received which is not the correct behaviour we currently have or want. Currently, we have this idea of send-to-selfs not being marked as change and we only display non-change where we spent funds to create the tx as "send-to-self". I think we want to define change as ANY output we receive in a tx created by our capability, so therefore we need to define a send-to-self as an output from a transaction we created with no outgoing_tx_data

Oscar-Pepper (dismissed):
really like where this PR is going!!

zingolib/src/lightclient/describe.rs
zancas: I will look into this.   My guess is that it's OK to lose type information *ONLY* if you're about to uniquely consume the error.

zingolib/src/lightclient/describe.rs
zancas: Good point!   I hadn't thought carefully about the out-of-order case here.

zingolib/src/lightclient/describe.rs
zancas: Unless this is blocking I am leaving it with a TODO note.

zingolib/src/lightclient/describe.rs
zancas: Really good point @Oscar-Pepper!  I am glad you thought through back-compatibility more carefully than I did.
I feel like this is something to reconsider in the new architecture.

zancas: Resolved, pending a local test pass.

Oscar-Pepper (approved):

PR #1186, opened 2024-06-04
Dust selection bug

Oscar-Pepper:
ontop of #1185 to avoid conflicts

- adds a failing test when dust is selected
- comments out dust selection for now

### DIFF
--- a/libtonode-tests/tests/legacy.rs
+++ b/libtonode-tests/tests/legacy.rs
@@ -12,6 +12,7 @@ use zcash_primitives::{
     consensus::{BlockHeight, Parameters},
     transaction::fees::zip317::MINIMUM_FEE,
 };
+use zingo_testutils::lightclient::from_inputs;
 use zingo_testutils::{
     self, build_fvk_client, check_client_balances, check_transaction_equality,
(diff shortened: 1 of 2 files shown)


zancas:
I clicked update on this PR.

libtonode-tests/tests/legacy.rs
```diff
@@ -4244,3 +4245,26 @@ mod basic_transactions {
 async fn proxy_server_worky() {
     zingo_testutils::check_proxy_server_works().await
 }
+
```
zancas, line 4248: I see this test will send dust too sapling from Orchard.   I'm not clear on correct behavior.

zancas:
I vote that we accept the amount leaked (1-dust, i.e. < 5_000) and the amount demotion (again < 5_000) in order to rescue that same amount, and have a less complex policy.

zancas:
In other words, I think it's a useful test highlighting an interesting case.   Probably the best balance of concerns is to take the small demotion and information leak, for the payoff of a relatively simple sweeping policy, and rescue (albeit demoted) of the dust.

Oscar-Pepper:
closed for #1188

PR #1192, opened 2024-06-06
spendable balance command

Oscar-Pepper:
- helperises send all spendable balance calculation
- adds spendable balance command
- adds helper for total confirmed shielded balance calculation
- organises errors into more suitable places and upgrades to thiserror

### DIFF
--- a/zingolib/src/lightclient/propose.rs
+++ b/zingolib/src/lightclient/propose.rs
@@ -146,7 +146,6 @@ impl LightClient {
 
     /// Unstable function to expose the zip317 interface for development
     // TOdo: add correct functionality and doc comments / tests
-    // TODO: Add migrate_sapling_to_orchard argument
     pub async fn propose_send_all(
         &self,
         address: zcash_keys::address::Address,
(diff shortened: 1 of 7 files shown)


zingolib/src/lightclient/propose.rs
```diff
@@ -146,7 +146,6 @@ impl LightClient {
 
     /// Unstable function to expose the zip317 interface for development
     // TOdo: add correct functionality and doc comments / tests
-    // TODO: Add migrate_sapling_to_orchard argument
```
zancas, line 149: Perhaps I will learn why later in this review, but I don't currently know why this TODO is removed.

zingolib/src/lightclient/propose.rs
```diff
@@ -146,68 +140,81 @@ impl LightClient {
...
                 zcash_client_backend::data_api::error::Error::InsufficientFunds {
-                    required, ..
+                    available,
+                    required,
                 },
             )) => {
                 if let Some(shortfall) = required - confirmed_shielded_balance {
-                    (confirmed_shielded_balance - shortfall).ok_or(ProposeSendError::Proposal(
+                    Ok(shortfall)
+                } else {
```
zancas, line 196: Really clever!

zancas (approved):

PR #1193, opened 2024-06-06
fix no sent valuetransfer bug

Oscar-Pepper:
bug: do_addresses returns the full UA with all receivers where outgoing_tx_data (used by list_txsummaries) only stores the full UA with legacy send. in some tests we try to find by key where the key no longer matches in zip317 send as the UA only has orchard receiver

this PR demonstrates a fix where we no longer store the UA with all receivers in the outgoing_tx_data and add an option to do_addresses that fetches the UA with only an orchard receiver.

alternatively we could implement storing all receivers by populating the Option<Recipient_ua> in zip317 sending but i'm not sure how much work that is. probably a post-release task. I imagine it requires the change memo to retrieve the full UA which may need some thinking...

for now the test passes but needs cleaning up and removing changes to legacy.rs before merge

### DIFF
--- a/darkside-tests/tests/advanced_reorg_tests.rs
+++ b/darkside-tests/tests/advanced_reorg_tests.rs
@@ -7,14 +7,19 @@ use darkside_tests::{
         TREE_STATE_FOLDER_PATH,
     },
     darkside_types::{Empty, TreeState},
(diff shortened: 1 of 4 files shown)


Oscar-Pepper:
closing as #1196 is the correct fix. i will extract darkside upgrade in the near future

PR #1194, opened 2024-06-06
remove krufted shielded_balance parameter

zancas:
This PR started life as part of my review for #1192 but this seemed like a landable chunk of improvement along the way.

The `target_addr` parameter was being passed `None` in all cases.

I suspect it pre-dated the filter abstraction.

Also in this PR:

* doc-comment for wallet/describe.rs
* super --> crate::wallet in wallet/describe.rs

Creates a trivial conflict with #1192

PR #1195, opened 2024-06-06
Make shielded balance more private

zancas:
Continuing review, I am refining the balance checking fns a bit.

Builds on #1194

* make shield_balance a private inner method
* make some wrappers generic
* remove kruft unused address parameter
* move get_wallet_capability our of "describe"


zingolib/src/wallet/describe.rs
```diff
@@ -61,17 +66,15 @@ impl LightWallet {
                 .transaction_records_by_id
                 .values()
                 .map(|transaction| {
-                    let mut filtered_notes: Box<dyn Iterator<Item = &D::WalletNote>> = Box::new(
-                        D::WalletNote::transaction_metadata_notes(transaction)
-                            .iter()
-                            .filter(filter_notes_by_target_addr),
-                    );
+                    let mut unfiltered_notes: Box<dyn Iterator<Item = &D::WalletNote>> =
```
zancas: Should just be "notes".

zingolib/src/wallet/describe.rs
```diff
@@ -20,39 +22,42 @@ use crate::wallet::notes::ShieldedNoteInterface;
...
     {
-        let fvk = D::wc_to_fvk(&self.wallet_capability()).ok()?;
-        let filter_notes_by_target_addr = |notedata: &&D::WalletNote| match target_addr.as_ref() {
-            Some(addr) => {
-                let diversified_address =
-                    &fvk.diversified_address(*notedata.diversifier()).unwrap();
-                *addr
-                    == diversified_address
-                        .b32encode_for_network(&self.transaction_context.config.chain)
+        // For the moment we encode lack of view capability as None
```
zancas, line 46: Actually there is a relevant error enum..  will look into changing over.

zingolib/src/lightclient/send.rs
```diff
@@ -75,14 +75,10 @@ impl LightClient {
         let transaction_submission_height = self.get_submission_height().await?;
         let fee = u64::from(MINIMUM_FEE); // TODO: This can no longer be hard coded, and must be calced
                                           // as a fn of the transactions structure.
-        let tbal = self
-            .wallet
-            .tbalance(None)
-            .await
-            .expect("to receive a balance");
+        let tbal = self.wallet.tbalance().await.expect("to receive a balance");
```
zancas: This expect is probably wrong.

zingolib/src/wallet/describe.rs
```diff
@@ -124,32 +116,8 @@ impl LightWallet {
...
-        self.shielded_balance::<D>(target_addr, filters).await
-    }
-
-    /// TODO: Add Doc Comment Here!
-    pub async fn verified_balance<D: DomainWalletExt>(
-        &self,
-        target_addr: Option<String>,
-    ) -> Option<u64>
+    /// On chain balance
+    pub async fn confirmed_balance<D: DomainWalletExt>(&self) -> Option<u64>
```
Oscar-Pepper, line 117: confirmed needs to be filtered by `is_confirmed` not by at or before anchor height

zingolib/src/wallet/describe.rs
```diff
@@ -164,15 +132,39 @@ impl LightWallet {
...
-        self.shielded_balance::<D>(target_addr, filters).await
+        self.shielded_balance::<D>(filters).await
+    }
+    /// The amount in orchard notes, not yet on chain
+    pub async fn pending_balance<D: DomainWalletExt>(&self) -> Option<u64>
+    where
+        <D as Domain>::Recipient: Recipient,
+        <D as Domain>::Note: PartialEq + Clone,
+    {
+        self.shielded_balance::<D>(&[Box::new(|note, _| note.pending_receipt())])
```
Oscar-Pepper: this should be based of confirmation status, the pending reciept is based on nullifier calculation and probably less stable
```suggestion
        self.shielded_balance::<D>(&[Box::new(|_, transaction| transaction.status.is_pending())])
```

Oscar-Pepper (dismissed):

fluidvanadium:
agree with Oscarz requested changes, otherwise gtg

fluidvanadium (dismissed):

zingolib/src/lightclient/send.rs
zancas: This fn is deprecated..  probably not marked as such because it's used in test.

zancas:
I addition to the original changes I changes `legacy` to `concrete` as the complement to `generic`.

libtonode-tests/tests/concrete.rs
```diff
@@ -1604,10 +1598,7 @@ mod slow {
             - external_transparent_3
             - (5 * u64::from(MINIMUM_FEE));
         assert_eq!(
-            recipient
-                .wallet
-                .shielded_balance::<OrchardDomain>(None, &[])
-                .await,
+            recipient.wallet.spendable_balance::<OrchardDomain>().await,
```
Oscar-Pepper, line 1601: shielded_balance::orchard is replaced twice with different methods?

zingolib/src/lightclient/describe.rs
```diff
@@ -93,21 +93,21 @@ impl LightClient {
         PoolBalances {
             sapling_balance: self
                 .wallet
-                .shielded_balance::<SaplingDomain>(None, &[])
+                .pending_and_confirmed_balance::<SaplingDomain>()
```
Oscar-Pepper: does it make more sense to have pending be "total funds once pending is confirmed" and delete pending_and_confirmed

Oscar-Pepper (dismissed):

libtonode-tests/tests/concrete.rs
```diff
@@ -1409,18 +1409,12 @@ mod slow {
             - first_send_to_transparent
             - (3 * u64::from(MARGINAL_FEE));
         assert_eq!(
-            recipient
-                .wallet
-                .shielded_balance::<OrchardDomain>(None, &[])
-                .await,
+            recipient.wallet.pending_balance::<OrchardDomain>().await,
```
zancas, line 1412: I believe that this is properly pending balance because the value in question has been calculated without adding a confirming block.

That is, the terms in the `expected_funds` sum are not from confirmed transactions, rather they are from `quick_sends` that are not yet validated.

libtonode-tests/tests/concrete.rs
```diff
@@ -1409,18 +1409,12 @@ mod slow {
...
             Some(expected_funds)
         );
         //  (2) The balance is not yet verified
         assert_eq!(
-            recipient
-                .wallet
-                .verified_balance::<OrchardDomain>(None)
-                .await,
+            recipient.wallet.confirmed_balance::<OrchardDomain>().await,
```
zancas, line 1417: These funds are the complementary (0) funds that *have* been confirmed at this point.   This assertion shows that no funds have yet been confirmed.

I propose that we move away from the term `verified` in favor of `confirmed` where `confirmed` is actually short-hand for a generally correct `confirmed_by_n_blocks` where `n == 1`.   So when we say `confirmed` we are using the shorthand of `confirmed_by_1_block` and if we mean confirmed by some other number of blocks, we say that number explicitly.

libtonode-tests/tests/concrete.rs
zancas: `shielded_balance` is renamed to `get_filtered_balance` it's an internal private helper that more pub interface functions call with various filter sets.  Since it's not `pub` it's not accessible in tests, and where it was called there should now be fns that acquire balances for specific sets of Outputs.

zingolib/src/lightclient/describe.rs
zancas: I have deleted the `pending_and_confirmed` and rewritten the call-sites.

zancas:
Hi @fluidvanadium can you please re-review this?

zancas:
OK, I have fixed merge conflicts and rerun all tests locally.
@fluidvanadium

libtonode-tests/tests/concrete.rs
fluidvanadium: yeah. did this test pass before??

zingolib/src/lightclient/describe.rs
```diff
@@ -90,24 +93,27 @@ impl LightClient {
 
     /// TODO: Add Doc Comment Here!
     pub async fn do_balance(&self) -> PoolBalances {
+        let verified_sapling_balance = self.wallet.confirmed_balance::<SaplingDomain>().await;
+        let unverified_sapling_balance = self.wallet.pending_balance::<SaplingDomain>().await;
+        let spendable_sapling_balance = self.wallet.spendable_balance::<SaplingDomain>().await;
+        let sapling_balance = some_sum(verified_sapling_balance, unverified_sapling_balance);
```
fluidvanadium, line 99: still leaves me unsure what the difference is between spendable and verified

zingolib/src/wallet/describe.rs
```diff
@@ -61,17 +64,14 @@ impl LightWallet {
                 .transaction_records_by_id
                 .values()
                 .map(|transaction| {
-                    let mut filtered_notes: Box<dyn Iterator<Item = &D::WalletNote>> = Box::new(
```
fluidvanadium, line 64: this could use outputs one day

zingolib/src/lightclient/describe.rs
fluidvanadium: found it

zingolib/src/wallet/describe.rs
```diff
@@ -125,55 +113,32 @@ impl LightWallet {
...
-            Box::new(|_, transaction| {
-                transaction
-                    .status
-                    .is_confirmed_before_or_at(&BlockHeight::from_u32(anchor_height))
-            }),
-            Box::new(|nnmd, _| !nnmd.pending_receipt()),
-        ];
-        self.shielded_balance::<D>(target_addr, filters).await
+        self.get_filtered_balance::<D>(&[Box::new(|note, _| note.pending_receipt())])
+            .await
```
fluidvanadium, line 136: this looks like the sum of domain notes, whether on chain or not?? but it seems like its only supposed to note the confirmed ones

fluidvanadium (approved):
there is one issue. other wise good

PR #1199, opened 2024-06-10
Dry giskit

zancas:


Oscar-Pepper (approved):
i am approving this but im not sure i agree that shielding should allow shielding dust

zancas:
This policy doesn't shield dust, it *would* hypothetically allow the creation of a dust note, but shield should always create a single note.   It's possible that note could be dust, but given a sweeping mechanism that might be the user's intention.

PR #1201, opened 2024-06-11
name local alias to emphasize proposal type

zancas:
Clarify what this zingo type is, and how it differs from the wrapped LRZ type.


fluidvanadium (approved):
mas claro

PR #1206, opened 2024-06-12
fixed query_for_ids to return correct PoolType

fluidvanadium:


Oscar-Pepper (approved):
can we unit test query_for_ids to prevent these bugs?

PR #1211, opened 2024-06-13
Stop using is_outgoing_transaction

zancas:
`is_outgoing_transaction` definition is marked-as-deprecated:

Notes about this deprecated fn:

* this function is a known cause of bugs
* all its call-sites are removed
* except for 1 unit-test all calls were already deprecated.

### DIFF
--- a/zingolib/src/lightclient/deprecated.rs
+++ b/zingolib/src/lightclient/deprecated.rs
@@ -123,10 +123,11 @@ impl LightClient {
         let mut consumer_ui_notes = tmds.transaction_records_by_id
             .iter()
             .flat_map(|(txid, wallet_transaction)| {
+                let is_received_transaction = tmds.transaction_records_by_id.transaction_is_received(wallet_transaction).expect("To not hit a fee error!");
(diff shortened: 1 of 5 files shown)


zingolib/src/lightclient/deprecated.rs
```diff
@@ -112,19 +112,22 @@ impl LightClient {
...
+            .transaction_metadata_set
             .read()
-            .await
-            .transaction_records_by_id
+            .await;
+        // Create a list of TransactionItems from wallet transactions
+        let mut consumer_ui_notes = tmds.transaction_records_by_id
             .iter()
             .flat_map(|(txid, wallet_transaction)| {
+                let is_received_transaction = tmds.transaction_records_by_id.transaction_is_received(wallet_transaction).expect("To not hit a fee error!");
```
Oscar-Pepper, line 126: this expect will be hit, please use new function in #1215

Oscar-Pepper (changes requested):

zancas:
Superseded by #1215

zancas:
Keeping open for small bits that might not yet be handled elsewhere.

PR #1213, opened 2024-06-14
Do list txs replacement

Oscar-Pepper:
the purpose of this PR is to replace do_list_transactions with a stable and more readable alternative. datetime is human readable, fee are included and note can be more easily tracked across txs.

- adds `transactions` command
- adds `transaction_summaries` pub lightclient method
-  adds `transaction_summaries_json_string` pub lightclient method for FFIs/lib consumers
- adds both json and fmt::display impls for all the summary structs
- aligns calculate_transaction_fee with output query functions


zancas:
What's the difference between a `TransactionSummary` and a `TransactionRecord`?  Why have both things?

Oscar-Pepper:
> What's the difference between a `TransactionSummary` and a `TransactionRecord`? Why have both things?

The aim of transaction summary is to provide a struct to store a snapshot of the relevent user/consumer facing information at the time of the call. for example a transaction record may have information that clutters up the display to the user and also may calculate its status or fee etc. from foundational truths. the summary shows a snapshot of what we knew about the chain at that point in time but a record can adapt and persist as we gain more knowledge

Oscar-Pepper:
> > What's the difference between a `TransactionSummary` and a `TransactionRecord`? Why have both things?
> 
> The aim of transaction summary is to provide a struct to store a snapshot of the relevent user/consumer facing information at the time of the call. for example a transaction record may have information that clutters up the display to the user and also may calculate its status or fee etc. from foundational truths. the summary shows a snapshot of what we knew about the chain at that point in time but a record can adapt and persist as we gain more knowledge

zancas (approved):
Because it's cloning data out this summary is a snapshot.

zancas:
It passed CI before I merged in dev.

PR #1214, opened 2024-06-14
[WIP] nix flake

nejucomo:
WIP == Work In Progress

I'm in the midst of creating a `flake.nix` file so that all `zingolib` crates can be build and tested on any machine with [nix](https://nixos.org/learn/).


nejucomo:
One reason I like `nix` is that it codifies _all_ dependencies. Compare [the current `flake.nix` specification of non-cargo dependencies](https://github.com/nejucomo/zingolib/blob/nix-flake/flake.nix#L43-L57) to [`README.md` Pre-requisites](https://github.com/nejucomo/zingolib/blob/nix-flake/README.md#pre-requisites).

The nix version is automated and verified on every (`nix`) build or (`nix`) test target, and it demonstrates that `git` is also a prerequisite that's not in the `README.md`.

nejucomo:
The current blocker for `nix build` is that it appears `zcash_client_backend` has a build script which attempts to write into its source directory ([example](https://github.com/zcash/librustzcash/blob/main/zcash_client_backend/build.rs#L39-L42)).

By default in `nix` all source must be read-only because it has deterministic builds as a top-line goal. A potential work-around in _this_ branch would be to tell the nix build process to copy the `zcash_client_backend` source to a temporary read/write directory when building.

However, a cleaner solution is to fix `zcash_client_backend` not to do this anti-pattern. I'm currently investigating the latter. I consider it an antipattern to conflate the output of hopefully reproducible automation such as `protoc` with human-edited revision controlled source, but presumably `zcash_client_backend` does this as a work-around for some limitation in `protoc`.

nejucomo:
I filed https://github.com/zcash/librustzcash/issues/1420 to track my preferred upstream fix which I consider cleanest. If that remains a blocker, the work-around fix within the nix build process is possible.

nejucomo:
I implemented the nix-build work-around for the https://github.com/zcash/librustzcash/issues/1420 issue because it felt quicker.

Now the new blocker is:

```
cargo-package> error: failed to run custom build command for `darkside-tests v0.1.0 (/build/source/darkside-tests)`
cargo-package> Caused by:
cargo-package>   process didn't exit successfully: `/build/source/target/release/build/darkside-tests-e0fd5ded20d526db/build-script-build` (exit status: 1)
cargo-package>   --- stdout
cargo-package>   cargo:rerun-if-changed=../zingo-testutils/proto/compact_formats.proto
cargo-package>   cargo:rerun-if-changed=proto/darkside.proto
cargo-package>   cargo:rerun-if-changed=../zingo-testutils/proto/service.proto
cargo-package>   cargo:rerun-if-changed=proto
cargo-package>   cargo:rerun-if-changed=../zingo-testutils/proto
cargo-package>   --- stderr
cargo-package>   Error: Custom { kind: Other, error: "protoc failed: Could not make proto path relative: ../zingo-testutils/proto/compact_formats.proto: No such file or directory\n" }
cargo-package> warning: build failed, waiting for other jobs to finish...
```

I am pausing work on this and haven't diagnosed this, but at first glance it looks like a build script is assuming crates are placed identically to the git source. I do not know why this would be violated in the nix build, but I'll see if I can remove that assumption from the build script. If not, I'll see about patching the nix build somehow to meet the assumption.

zancas:
Is this still in progress?

zancas:
Hi @nejucomo is this PR still viable?

PR #1215, opened 2024-06-14
replace transaction_is_received

Oscar-Pepper:
this Pr replaces `transaction_is_received` with the new `transaction_kind` as the error handling of is_received was bugged


zingolib/src/wallet/transaction_records_by_id.rs
```diff
@@ -376,6 +365,51 @@ impl TransactionRecordsById {
...
         self.invalidate_transactions(txids_to_remove);
     }
+
+    /// Note this method is INCORRECT in the case of a 0-value, 0-fee transaction from the
+    /// Creating Capability.  Such a transaction would violate ZIP317, but could exist in
+    /// the Zcash protocol
+    ///  TODO:   Test and handle 0-value, 0-fee transaction
+    pub(crate) fn transaction_kind(&self, query_record: &TransactionRecord) -> TransactionKind {
+        let mut sapling_spends: Vec<&SaplingNote> =
+            Vec::with_capacity(query_record.spent_sapling_nullifiers.len());
```
zancas: Why are you doing this explicitly?  Isn't the capacity allocation algorithm already fairly efficient?

zingolib/src/wallet/transaction_records_by_id.rs
```diff
@@ -376,6 +365,51 @@ impl TransactionRecordsById {
...
+            .for_each(|nullifier| {
+                if let Some(spend) = self.find_orchard_spend(nullifier) {
+                    orchard_spends.push(spend);
+                }
+            });
+
+        if sapling_spends.is_empty()
+            && orchard_spends.is_empty()
+            && query_record.total_transparent_value_spent == 0
+            && query_record.outgoing_tx_data.is_empty()
```
zancas, line 396: Is there also necessarily an Output created here?

zingolib/src/wallet/transaction_records_by_id.rs
```diff
@@ -376,6 +365,51 @@ impl TransactionRecordsById {
...
+        {
+            TransactionKind::Received
+        } else if sapling_spends.is_empty()
+            && orchard_spends.is_empty()
+            && query_record.total_transparent_value_spent > 0
+            && query_record.outgoing_tx_data.is_empty()
+        {
+            // TODO: this could be improved by checking outputs recipient addr against the wallet addrs
+            TransactionKind::Sent(SendType::Shield)
```
zancas, line 405: Won't this no longer be true in ZIP320?

zingolib/src/wallet/transaction_records_by_id.rs
Oscar-Pepper: it's best practice to allocate if you know the max size right? why re-allocate for no reason?

zingolib/src/wallet/transaction_records_by_id.rs
Oscar-Pepper: sorry i dont understand your question. the logic here says "if no spends have been synced for this tx and there is no outgoing tx data, this is a received tx"

zingolib/src/wallet/transaction_records_by_id.rs
Oscar-Pepper: this still holds for zip320 because when sending to exchange with 2-step the second step will still create outgoing tx data OR the address will not match our own if we change the logic as the TODO describes

zingolib/src/wallet/transaction_records_by_id.rs
Oscar-Pepper: we will likely add TEX as a `SendType` variant

zingolib/src/wallet/transaction_records_by_id.rs
Oscar-Pepper: we will likely add TEX as a SendType variant

zingolib/src/wallet/transaction_records_by_id.rs
Oscar-Pepper: i think i understand whaty you are saying, that maybe we should explicitly check we received an output? I am happy with this implicit check as there is no way we would be aware of a transaction_record that has no spends or outputs to our capability

zingolib/src/wallet/transaction_records_by_id.rs
```diff
@@ -211,81 +211,92 @@ impl TransactionRecordsById {
         });
     }
 
+    fn find_sapling_spend(&self, nullifier: &sapling_crypto::Nullifier) -> Option<&SaplingNote> {
```
fluidvanadium, line 214: good and deserves doc-comments

zingolib/src/wallet/transaction_records_by_id.rs
```diff
@@ -211,81 +211,92 @@ impl TransactionRecordsById {
...
+            .flat_map(|wallet_transaction_record| wallet_transaction_record.sapling_notes())
+            .find(|&note| {
+                if let Some(nf) = note.nullifier() {
+                    nf == *nullifier
+                } else {
+                    false
+                }
+            })
+    }
+    fn find_orchard_spend(&self, nullifier: &orchard::note::Nullifier) -> Option<&OrchardNote> {
```
fluidvanadium, line 225: YES, this is the architecture from our plan at Zing0!

zingolib/src/wallet/transaction_records_by_id.rs
```diff
@@ -211,81 +211,92 @@ impl TransactionRecordsById {
...
-            })
-            .collect::<Result<Vec<&SaplingNote>, FeeError>>()
+        let mut sapling_spends: Vec<&SaplingNote> =
+            Vec::with_capacity(query_record.spent_sapling_nullifiers.len());
+
+        for nullifier in query_record.spent_sapling_nullifiers() {
+            if let Some(spend) = self.find_sapling_spend(nullifier) {
+                sapling_spends.push(spend);
+            } else if fail_on_miss {
+                return Err(FeeError::SaplingSpendNotFound(*nullifier));
```
fluidvanadium, line 250: not sure about this error type... nothing to do with a fee, really
perhaps make the error type into Vec<&SaplingNote>... then you can eliminate the bool and vary the return code while still representing whether the search is complete.

my understanding is that this function will fail if called with the boolean on an incomplete set:: ie it will highlight an input we havent scanned the output for yet

fluidvanadium (approved):
this functionality patches a critical blind spot in our view of transactions!!! this is the beginning of a sound perspective on inputs

zingolib/src/wallet/transaction_records_by_id.rs
fluidvanadium: ok for now

PR #1216, opened 2024-06-14
Pending QuickSend Correct Fee TEST

fluidvanadium:
Previously, fees were miscalculated before scan. FIXED

Whoohoo! This branch buffs the with_assertions module to make sure everything is right from the initial record, even before scanning the broadcast.
Previous incarnations of this branch failed. As recently as Wednesday, the following tests failed:
  
 --- STDERR:              libtonode-tests::chain_generics chain_generics::propose_and_broadcast_40_000_to_transparent ---
thread 'chain_generics::propose_and_broadcast_40_000_to_transparent' panicked at /home/cupress/ZINGO/libs/pending/zingo-testutils/src/assertions.rs:32:9:
assertion `left == right` failed
  left: 55000
 right: 15000

--- STDERR:              libtonode-tests::chain_generics chain_generics::send_shield_cycle ---
thread 'chain_generics::send_shield_cycle' panicked at /home/cupress/ZINGO/libs/pending/zingo-testutils/src/assertions.rs:32:9:
assertion `left == right` failed
  left: 124000
 right: 20000

However, merging todayz dev fixed this!


zancas:
@fluidvanadium should you click "Update Branch"?

zancas:
I clicked update branch.

zingo-testutils/src/lightclient/with_assertions.rs
```diff
@@ -29,6 +29,7 @@ where
 
     environment.bump_chain().await;
 
+    assert_send_outputs_match_sender(client, &proposal, &txids).await;
```
zancas: I am having a hard time parsing this name.

To me it seems like this is in a mod called `assertions`.. so if I know which mod it lives in, then I know it's an `assertion`.

It seems like this code shows that the LRZ calculated:

`step.balance().fee_required()`

matches the:

`records.calculate_transaction_fee(record)`

Then if there's agreement, that `fee` is returned.

So, this is a fn in a mod name assertions but it does not return a bool.   That's my intuition for a fn in an assertions mod.

I'll propose-as-code.

zancas (dismissed):

zingo-testutils/src/lightclient/with_assertions.rs
```diff
@@ -69,10 +71,11 @@ where
     environment.bump_chain().await;
 
     recipient.do_sync(false).await.unwrap();
-    assert_send_outputs_match_recipient(recipient, &proposal, &txids).await;
+    get_proposal_vs_records_matched_total_output_value(recipient, &proposal, &txids).await;
```
fluidvanadium: actually, im afraid somebody will see this line does nothing, not realize it asserts, and delete it. it needs to start with 'assert

zancas (approved):

PR #1218, opened 2024-06-14
Upgrade rescan still have outgoing metadata

zancas:
This test upgrade to two tests:

* depends on more foundational truths than do_list_transcations and do_list_notes
* removes calls to do_list_transactions
* removes calls to do_list_notes
* collects related tests in mod
* removes spurious (and time consuming chain bumps)
* tests an extra case (external send with memo)
* DRYs code with macros
* exposes a potential bug where *pending* outgoing_tx_data is lost across a rescan
* fixes a test that was failing against #1215 (by replacing `do_list_transactions` with direct references to `outgoing_tx_data`)

### DIFF
--- a/libtonode-tests/tests/legacy.rs
+++ b/libtonode-tests/tests/legacy.rs
@@ -2273,77 +2273,98 @@ mod slow {
                 .max_leaf_position(0)
         );
     }
-    #[tokio::test]
-    async fn rescan_still_have_outgoing_metadata_with_sends_to_self() {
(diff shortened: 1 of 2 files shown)


zancas:
Upon further investigation, I found that the `multiple_outgoing_metadatas_work_right_on_restore` test was redundant _except_ that it added test of list_txsummaries.   I added list_txsummaries to the validation macro, and ignored `multiple_outgoing_metadatas_work_right_on_restore` after renaming it to `check_list_txsummaries_across_rescan`

Oscar-Pepper (approved):

PR #1220, opened 2024-06-17
Do list txs replacement - preliminary

Oscar-Pepper:
this PR lays the foundations for #1213 and is on top of #1215.
- renames list_txsummaries -> list_value_transfers to avoid conflict with new transaction summaries interface
- creates to-be-used structs for displaying tx summaries to user
- creates unimplemented list_transactions fn and TransactionsCommand
- adds error type for production builders
- moves build macros to utils for use in production builders as well as mocks


fluidvanadium (approved):
reviewed!

PR #1225, opened 2024-06-17
Any pool output

zancas:
Add the `notes::AnyPoolOutput` enum.   This type provides one variant per pool.   All variants wrap a relevant type:

`Transparent(TransparentOutput)`
`Sapling(SaplingNote)`
`Orchard(OrchardNote)`

All wrapped types implement `OutputInterface`.

The `enum-dispatch` crate provides derive procedural macros that dispatch from methods of the `AnyPoolOutput` to the wrapped type `OutputInterface` trait method this provides an ergonomic interface to access pool agnostic output functionality.

At a high level this functionality will replace the older `do_list_notes` interface.

As `do_list_notes` is replaces with `AnyPoolOutputs` constructor calls in test, the `AnyPoolOutput` logic will be integrated (first) into test logic.


fluidvanadium (commented):
0a0ddfa4afe9732540c3f2f3bac1d6466a03347d: noticed that AnyPoolOutput points to a clone, not a ref. after a couple moments, i understood the advantage of being able to keep referencing it in different contexts

fluidvanadium (commented):
b3b5fd0dcd08b96efa21b3dfd350806dc9287301
these traits get to be implicitly inherited

fluidvanadium:
7823aa190116bea95511f94c103e54555908fa7b

maybe i dont truly understand why OutputInterface and OutputConstructors are different traits....

b4098cb1d4aaacf03d65fc5517d1246c76d25c4d

is there any reason to use more mutable vecs at transaction_record.rs:146? why dont we continue to just extend the one?

64319209924bf7fd2554123a1ed36c36d7fe7764

commit message is misleading

77a38d758a18daa51c4ea5e49ff6a15015d69d12

wonder if there is a way we could keep the unit test and put it to the replacement for pool_value_received

zingolib/src/wallet/transaction_record.rs
```diff
@@ -848,42 +868,34 @@ mod tests {
...
-                    unspent,
-                    pending_spent,
-                    spent,
-                    transparent,
-                    sapling,
-                    orchard,
-                ))
-                .len(),
-            expected,
+        let default_nn_transaction_record = dbg!(nine_note_transaction_record_default());
```
Oscar-Pepper, line 894: dbg! suppose to be left in here?

Oscar-Pepper (approved):
wonderful

PR #1240, opened 2024-06-21
add zennies_for_zingo bool to SendAll and SpendableBalance

### LINKED ISSUE #1233: Fee erroneous to have balance in Sapling and Orchard and Send all.
In the last tests made to the Beta RC version I could see that the fee calculation with the ZIP 317 is working quite well, but there is a problem, when you have balance in Sapling and Orchard and you try to Send everything having activated the option Donate to ZingoLabs, the fee calculation gives error.

### Requirements:
- Have balance in Sapling.
- Activate the Send All function from Settings.
- Activate the Donate to ZingoLabs option from Settings.

### Steps to reproduce the error:

1. Click on Send
2. Paste a unified address that has Orchard receiver.
3. Click on Send All

### Current behavior:

You will be able to see that tapping on Send All causes the fee calculation to error (shown in red).

### Expected behavior:

The fee calculation should be correct since when Send all the fee deduction should be made according to the inputs and outputs of the transaction.

### IMPORTANT:
This error only occurs when the address is unified with Orchard receiver (Zingo sends funds to Orchard by default), with Z (Sapling) and T (Transparent) address the fee calculation is correct.

### Demo:

https://github.com/zingolabs/zingolib/assets/37910384/e14e78e0-91a6-4e87-95df-04558ee45cc1

zancas:
to send_all and spendable_balance commands

Fixes: #1233 (with help from zingo-mobile)


zingolib/src/commands.rs
```diff
@@ -871,11 +877,13 @@ impl Command for SendAllCommand {
...
+            "zennies_for_zingo" must be specified, if set to 'true' 1_000_000 ZAT will be sent to the zingolabs
+            developer address with each transaction.
 
             Warning:
                 Does not send transparent funds. These funds must be shielded first. Type `help shield` for more information.
             Usage:
-                sendall <address> "<optional memo>"
+                sendall <address> <zennies_for_zingo> "<optional memo>"
                 OR
                 sendall '[{"address":"<address>", "memo":"<optional memo>"}]'
```
juanky201271: Are you going to add `zennies_for_zingo` here as well?

juanky201271 (dismissed):

zingolib/src/commands.rs
zancas: Updated.

juanky201271: thanks.

Oscar-Pepper (approved):
Great job

zingolib/src/commands.rs
```diff
@@ -871,13 +877,15 @@ impl Command for SendAllCommand {
...
+            developer address with each transaction.
 
             Warning:
                 Does not send transparent funds. These funds must be shielded first. Type `help shield` for more information.
             Usage:
-                sendall <address> "<optional memo>"
+                sendall <address> <zennies_for_zingo> "<optional memo>"
                 OR
-                sendall '[{"address":"<address>", "memo":"<optional memo>"}]'
+                sendall '[{"address":"<address>", "memo":"<optional memo>"}]' <zennies_for_zingo>
```
Oscar-Pepper: Zennies for zingo needs to be part of the json string here

zingolib/src/commands/utils.rs
```diff
@@ -70,20 +70,29 @@ pub(super) fn parse_send_args(args: &[&str], chain: &ChainType) -> Result<Receiv
...
     Ok(send_args)
 }
-
-// Parse the send arguments for `propose_send` when sending all funds from shielded pools.
 // The send arguments have two possible formats:
-// - 1 argument in the form of a JSON string (single address only). '[{"address":"<address>", "memo":"<optional memo>"}]'
-// - 1 (+1 optional) arguments for a single address send. &["<address>", "<optional memo>"]
+// - 2 arguments in the form of:
+//    *  a JSON string (single address only). '[{"address":"<address>", "memo":"<optional memo>"}]'
+//    * true|false
```
Oscar-Pepper: I think the point of json string is all args can be sent through FFi as a single string?

zingolib/src/commands/utils.rs
```diff
@@ -131,13 +140,16 @@ pub(super) fn parse_send_all_args(
 pub(super) fn parse_spendable_balance_args(
     args: &[&str],
     chain: &ChainType,
-) -> Result<Address, CommandError> {
```
Oscar-Pepper, line 134: Doc comment is not updated for this fn

zingolib/src/commands/utils.rs
```diff
@@ -131,13 +140,16 @@ pub(super) fn parse_send_all_args(
...
 
-    if args.len() != 1 {
+    let mut zennies_for_zingo = false;
+    if args.len() > 2 {
         return Err(CommandError::InvalidArguments);
     }
-
+    if args.len() == 2 && args[1] == "true" {
+        zennies_for_zingo = true;
+    }
```
Oscar-Pepper: Rather than set false initially and set true if true is correctly spelt etc. I think it safer to move this logic into the next block where we check if the first argument is a zcash address (not a json string) and parse true OR false there OR error

Oscar-Pepper: This also requires us to move the zennies for zingo into json which I think is correct for the json option

Oscar-Pepper (dismissed):
Noticed some potential issues

zingolib/src/commands/utils.rs
juanky201271: Yes, in the FFI for `execute` we have in zingo-mobile only 2 parameters:
`string execute_command(string cmd, string args);` (from the UDL file)
- cmd - `sendall` or `spendablebalance`
- args - with a string or a JSON string as well. For `sendall` ->`{ "address": "<address>", "memo": "<optional memo>", "zennies_for_zingo": "<true|false>" }` and for `spendablebalance` -> `{ "address": "<address>", "zennies_for_zingo": "<true|false>" }`

zancas: Ack, will fix.

zingolib/src/commands/utils.rs
zancas: Fixed.

zingolib/src/commands/utils.rs
zancas: I believe this is now ready for review.

zancas:
@fluidvanadium this fixes a release blocker, can you please review?

zancas:
@Oscar-Pepper reviewed by phone!

PR #1242, opened 2024-06-22
helperize core scan_full_tx logic

zancas:


zancas:
@AloeareV if you're around I'd appreciate you taking a look at this.   It has a significant amount of generic-bound related refactor.

zingolib/src/wallet/transaction_context.rs
```diff
@@ -174,6 +175,48 @@ pub mod decrypt_transaction {
             }
         }
 
+        /// INVARIANT: All transaction from the mempool are received.
+        /// If the transaction was created locally and broadcast then:
+        ///  * it is in the database "trbid", unless that structure was wiped
+        ///  * if memory was lost after broadcast, we'll just wait for a transaction
+        ///     that we broadcast to show up on chain, after all until it's validated..
+        ///     it's just gossip, this is an edge case we are safe ignoring
```
Oscar-Pepper: i believe that if the app is closed, a pending transaction added due to broadcasting would not be stored and when the app is restarted this tx would not appear in the wallet. when the broadcast tx is received again from the mempool it would satisfy the `is_none()` logic and then be re-added to the wallet. therefore, this comment is maybe misleading us to believe that the pending tx will not re-appear if the memory was lost

Oscar-Pepper: i have just realised that you have removed some scanning logic such as adding outgoing_tx_datas from the mempool scanning. this will appear very buggy when users close and re-open the app etc. as sending transactions will change to receiving etc.

i understand that your assumption is that mempool is only for receiving transactions, but we also use the mempool to restore pending transactions that we have broadcast. I don't see any reason why calling scan_full_tx is a problem here but calling this new function will cause bugs on reload

zingolib/src/wallet/transaction_context.rs
```diff
@@ -211,6 +254,17 @@ pub mod decrypt_transaction {
             .await;
         }
 
+        async fn _decrypt_incoming_transaction_to_record_transparent(
```
Oscar-Pepper: method is not used and is unnecessary as just a straight wrapper with no extra gumph

zingolib/src/wallet/transaction_context.rs
```diff
@@ -219,6 +273,19 @@ pub mod decrypt_transaction {
             taddrs_set: &HashSet<String>,
         ) {
             // Scan all transparent outputs to see if we received any money
+            self.account_for_transparent_receipts(transaction, status, block_time, taddrs_set)
+                .await;
+            // Scan transparent spends
+            self.account_for_transparent_spending(transaction, status, block_time)
+                .await;
```
Oscar-Pepper, line 226: i like this helperisation

zingolib/src/lightclient/sync.rs
```diff
@@ -191,24 +191,35 @@ impl LightClient {
...
-                            .await;
+                            // If the txid is already in the db, then it's already recorded
+                            // there's nothing new to do until it's read from chain.
+                            // ASSUMPTION: A transaction from the mempool_receiver is not on-chain
+                            if transaction_metadata_set
+                                .read()
+                                .await
+                                .transaction_records_by_id
+                                .get(&transaction.txid())
+                                .is_none()
```
Oscar-Pepper, line 202: yeah i like this

zingolib/src/wallet/transaction_context.rs
```diff
@@ -397,80 +468,27 @@ pub mod decrypt_transaction {
...
-        /// _IS_ 1 Spend and 1 Output.
-        #[allow(clippy::too_many_arguments)]
-        async fn decrypt_transaction_to_record_domain<D: DomainWalletExt>(
+        async fn account_for_shielded_receipts<Q>(
             &self,
-            transaction: &Transaction,
+            ivk: Ivk<Q, External>,
+            domain_tagged_outputs: &[(
+                Q,
+                <<Q as DomainWalletExt>::Bundle as wallet::traits::Bundle<Q>>::Output,
```
Oscar-Pepper: everywhere else is the codebase Domain is represented by D so i find this more confusing to read with Q, im not sure why we would change it

Oscar-Pepper (dismissed):
I like the added protection to mempool scanning and helperisation in this PR. my main issue blocking this PR is I think mempool scanning should still call `scan_full_tx` and not use this new function `receive_transaction_from_mempool`. the reason being that we are not gaining anything, but we are introducing bugs and also creating multiple places where similar actions are performed (i.e scanning).

zingolib/src/wallet/transaction_context.rs
zancas: A good illustration of why we need to separate concerns is the redefinition of `block_height` as an `Option<` in the parameters of `scan_full_tx`:

https://github.com/zingolabs/zingolib/blob/f8d1c67cc1966c68dc0eb0a6017cc96ec6ad8c33/zingolib/src/wallet/transaction_context.rs#L102

The comment shows that the code change is meant to handle an edge case that's only hit in some contexts:

`block_time: Option<u32>, // block_time should only be None when re-scanning a tx that already exists in the wallet`

This means that *all* of the scanning logic must be adapted to handle "re-scanning a tx that already exists in the wallet".

In the 4 call sites to `scan_full_tx` `block_time` is `Some(block_time)` in 3 and `None` in one.   Now, buried inside the `scan_full_tx` there must be conditional logic that does different things depending on that `Optional` parameter.

This means that a bug in the handling of that one optional parameter now has the potential to affect *all* of our scanning logic.

In my opinion separating the `None` case out into it's own concern is probably a better solution since it eliminates the need for conditional logic that affects all the scanning logic.

If that seems like a reasonable approach then the next question is...  well what do you think of that reasoning?

zancas: The first place we could benefit from such a separation of concerns would be at the call site itself.   The new method that used to be `scan_full_tx` with an optional parameter could simply have 1 fewer parameters.   Meanwhile the method (or methods) that formerly received a `Some(block_height)` would be contracted to accept a bare `block_height` instead.

zancas: And of course the conditional logic inside each of those flows could be removed.

zancas: Per your point that my assumptions about `rtransactions` from the mempool not always being received, and in the case that they're *`forgotten-sent`*, the app misbehaving:

I'm not surprised, and agree that we need to improve the code to properly handle the case.   I wasn't totally clear on correct behavior, and wrote that doc-comment explicitly calling out the edge case in the full expectation that it might well be innacurate.

zancas:
> the reason being that we are not gaining anything, but we are introducing bugs and also creating multiple places where similar actions are performed (i.e scanning).

As I have outlined, I disagree about the cost-benefit of splitting the entrypoints up, but I agree that it is best treated as a separate concern (whether or not we eventually do).

Oscar-Pepper (approved):
great job

PR #1244, opened 2024-06-24
Value transfer rewrite

Oscar-Pepper:
we are still seeing bugs associated with displaying value transfer kinds correctly in zingo mobile.
this PR aims to solve an issue that has come to light where value transfers are attempted to be used to reconstruct transactions as the value transfer abstraction was not functional or well-defined.
the major change here is that value transfers are now constructed by first calling `transaction_summaries()` which already contains stable "from foundational truth" logic that determines key building blocks for value transfers such as whether the tx was created by the capability and if so, whether this was a shielding tx or not. This helps us not only separate concerns but also not repeat logic and DRY the code.

- renames list_value_transfers() -> value_transfers()
- reworks vlaue transfer fields to have transaction information first and then specific value transfer information last. also adds fields so all info is available from only value transfers and updates fields to align with up-to-date zingolib structs and align with TransactionSummary
- found a bug in darkside re-org test! outgoing tx data is not found on re-org (there is a FIXME in the code)
- changed command name from "summaries" to "valuetransfers" and now uses fmt::Display for clean zingo-cli display
- rewrote the value_transfer() logic (coming PR with factor/helperise this code **thoughtfully and carefully**)
- value_transfers_json_string() for json consumers i.e. zingo-mobile
- moved balance_delta() to transaction summaries as it doesnt belong in value transfers due to complications with shared information like fees etc.
- value transfer now has full set of getters private fields and a builder struct
- redefined value transfer kinds to:
  > sent, per unique address in outgoing_tx_data
  > note-to-self, per sent tx with received note WITH memos (so normal change doesnt clutter up the interface)
  > shield, the whole transaction is of kind Send(SendType::Shield) with only transparent inputs and no outgoing_tx_data, per pool received to
  > receive, no inputs from this capability and received outputs, per pool received to
- full fmt::Display for printing and creating strings for json objects


zancas (approved):
I find it very concerning that no test got the duplicate check of orchard notes bug.

zingolib/src/lightclient/describe.rs
```diff
@@ -281,54 +278,251 @@ impl LightClient {
...
-                    datetime,
-                    kind: ValueTransferKind::Fee { amount: tx_fee },
-                    memos: vec![],
-                    price,
-                    txid: *txid,
-                    pending,
-                });
+                    // create 1 note-to-self if a sending transaction receives any number of memos
+                    if tx.orchard_notes().iter().any(|note| note.memo().is_some())
+                        || tx.orchard_notes().iter().any(|note| note.memo().is_some())
```
zancas, line 350: I don't know how I let this bug get in!  I fixed it somewhere.  Fixing again.

PR #1245, opened 2024-06-25
fixed transparent balance bug

### LINKED ISSUE #1236: Balance doubles while transparent balance is shielded
When you perform a transaction to shield the transparent balance, while the transaction is being processed the current balance is duplicated, thus generating a visual bug that can confuse the end user.

In the image you can see that there is a transaction in the history that is being processed and above the balance has increased to double what the wallet actually has.

image

fluidvanadium:
fixes #1236

### DIFF
--- a/zingolib/src/wallet/describe.rs
+++ b/zingolib/src/wallet/describe.rs
@@ -105,13 +105,14 @@ impl LightWallet {
         }
     }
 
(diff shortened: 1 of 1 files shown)


zancas (approved):
There's a test that will soon land, elsewhere.

zancas:
Nicely done!

PR #1248, opened 2024-06-26
Remove old send

fluidvanadium:


zancas (changes requested):
These were quite useful to test consistent behavior with the new vanilla quick_send and quick_shield..  are we sure we're ready to deprecate?

I prefer deprecation first, deletion later.  In this case the `#[deprecated, reason = "` marker will keep us aware that these are being called in test.

zancas (changes requested):
I vote we follow @Oscar-Pepper 's suggestion and deprecate-for-now...    deleting later in a coordinated way.

zancas (approved):
This is a safe change.

PR #1254, opened 2024-06-29
Require change

zancas:
Create zero-value-change note for the sake of the "Check Stub" memo that Zingo always sends to itself.

Supersedes #1253


juanky201271 (approved):
Of course I am not the person to review this, but I trust... (I approve it in case you need to merge this ASAP).

PR #1268, opened 2024-07-04
Spent in

fluidvanadium:
consolidates spent and pending_spent into a single field


zingolib/src/wallet/notes/interface.rs
```diff
@@ -42,35 +44,45 @@ pub trait OutputInterface: Sized {
...
-    /// The TxId and broadcast height of a transfer that's not known to be on-record on the chain
-    fn pending_spent(&self) -> &Option<(TxId, u32)>;
+    /// returns the id of the spending transaction, whether pending or no
+    fn spending_txid(&self) -> Option<TxId> {
+        self.spend().map(|(txid, _status)| txid)
+    }
 
-    /// TODO: Add Doc Comment Here!
-    fn pending_spent_mut(&mut self) -> &mut Option<(TxId, u32)>;
+    /// Returns true if the note has ever been spent.
```
zancas: But it actually returns a `SpendStatus`.

zingolib/src/wallet/notes/interface.rs
```diff
@@ -42,35 +44,45 @@ pub trait OutputInterface: Sized {
...
+    fn spend_status(&self) -> SpendStatus {
```
zancas: I love default methods that consume other trait methods.

zingolib/src/wallet/traits.rs
```diff
@@ -1214,12 +1214,17 @@ where
...
+            .and_then(|(txid, status)| status.get_confirmed_height().map(|height| (txid, height)));
+
         Optional::write(
             &mut writer,
-            self.spent().as_ref(),
+            confirmed_spend,
             |w, (transaction_id, height)| {
                 w.write_all(transaction_id.as_ref())?;
-                w.write_u32::<LittleEndian>(*height)
+                w.write_u32::<LittleEndian>(height.into())
```
zancas, line 1225: We need to test make sure this saves and loads correctly.

zingolib/src/wallet/notes/interface.rs
fluidvanadium: fixed

zingolib/src/wallet/describe.rs
```diff
@@ -254,7 +258,7 @@ impl LightWallet {
                 transaction
                     .transparent_outputs
                     .iter()
-                    .filter(|utxo| !utxo.is_spent())
+                    .filter(|utxo| !utxo.is_spent_confirmed())
```
Oscar-Pepper, line 257: the comment says pending spent utxos are included. it looks that is not correct anymore?

zingolib/src/wallet/describe.rs
```diff
@@ -287,6 +291,26 @@ impl LightWallet {
...
+    output
+        .spending_txid()
+        .map(|spending_txid| {
+            records
+                .get(&spending_txid)
+                .map(|transaction_record| Some(transaction_record.status))
+                .ok_or(spending_txid)
+        })
+        .unwrap_or(Ok(None))
+}
```
Oscar-Pepper: I like the idea to have a more "foundational" way to check an output is spent for future use but this one doesnt make too much sense to me. so the error variant of the Result is a TxId but the Result outputted can never reach the error variant? why not just output `Option<ConfirmationStatus>`?

zingolib/src/wallet/notes/interface.rs
```diff
@@ -42,35 +44,45 @@ pub trait OutputInterface: Sized {
     fn value(&self) -> u64;
 
     /// If the funds are spent, the TxId and Blockheight of record
-    fn spent(&self) -> &Option<(TxId, u32)>;
+    fn spend(&self) -> &Option<(TxId, ConfirmationStatus)>;
```
Oscar-Pepper: i think spend is more confusing than spent in the way it reads in the code. i feel spend should be used to referrence an actual output itself, or in some cases the nullifier in the input of a tx, or could sound like an action as if spend "spends" the note. but this is a check on whether we think it is "spent" or not

zingolib/src/wallet/notes/interface.rs
```diff
@@ -42,35 +44,45 @@ pub trait OutputInterface: Sized {
...
-    fn pending_spent(&self) -> &Option<(TxId, u32)>;
+    /// returns the id of the spending transaction, whether pending or no
+    fn spending_txid(&self) -> Option<TxId> {
+        self.spend().map(|(txid, _status)| txid)
+    }
 
-    /// TODO: Add Doc Comment Here!
-    fn pending_spent_mut(&mut self) -> &mut Option<(TxId, u32)>;
+    /// Returns Unspent if the note has never been spent. Otherwise, points to the txid.
+    fn spend_status(&self) -> SpendStatus {
```
Oscar-Pepper: this returns a SpendStatus which is part of the summaries mod. I think its great we can use stuff made for summaries in the core logic but it needs to move out of the summaries mod. I did the same for TransactionKind which now lives in transaction_record.rs. can you move SpendStatus to a logical place, maybe where the OutputInterface stuff is as its shared by multiple outputs?

zingolib/src/wallet/describe.rs
Oscar-Pepper: see my comment on spend_status below, i think we should wait to implement this and do it properly when we update the wallet version to include transparent input info in transaction records

zingolib/src/wallet/notes/interface.rs
Oscar-Pepper: i actually think that when we updat the wallet version to include transparent input info in tx's, this (and get_spend_status in describe.rs) should be removed and we only check the spend status of outputs from the transaction_records_by_id layer (similar to how fee calculation works outside of a transaction record). therefore, we can have a method on transaction_records_by_id that takes an output and then checks the tx it is "supposed" to be spent in to verify. this is the "foundational" way to check spend status and we should base all core logic on spend_status and not the option from spent(). anyways this is not for this PR!

zingolib/src/wallet/notes/interface.rs
```diff
@@ -42,35 +44,45 @@ pub trait OutputInterface: Sized {
...
+            None => SpendStatus::Unspent,
+            Some((txid, ConfirmationStatus::Pending(_))) => SpendStatus::PendingSpent(*txid),
+            Some((txid, ConfirmationStatus::Confirmed(_))) => SpendStatus::Spent(*txid),
+        }
+    }
 
     /// Returns true if the note has been presumptively spent but the spent has not been validated.
-    fn is_pending_spent(&self) -> bool {
-        self.pending_spent().is_some()
+    fn is_spent_pending(&self) -> bool {
```
Oscar-Pepper: is_pending_spent makes more sense to me than is_spent_pending

zingolib/src/wallet/notes/interface.rs
```diff
@@ -42,35 +44,45 @@ pub trait OutputInterface: Sized {
...
+        self.spend()
+            .is_some_and(|(_txid, status)| status.is_pending())
     }
 
-    /// returns true if the note is confirmed spent
-    fn is_spent(&self) -> bool {
-        self.spent().is_some()
+    /// returns true if the note is spent and the spent is validated confirmed on chain
+    fn is_spent_confirmed(&self) -> bool {
```
Oscar-Pepper, line 66: i quite like `is_pending_spent` and `is_spent` to match the spend status enum variants and makes sense to me but i think i at least prefer `is_confirmed_spent` here

zingolib/src/wallet/traits.rs
Oscar-Pepper: Yes can you verify this has been tested before merge?

zingolib/src/wallet/transaction_records_by_id.rs
```diff
@@ -188,17 +187,14 @@ impl TransactionRecordsById {
...
-                    *nd.spent_mut() = None;
-                }
-
-                // Remove pending spends too
-                if nd.pending_spent().is_some()
-                    && invalidated_txids.contains(&nd.pending_spent().unwrap().0)
+                if note
+                    .spend()
+                    .filter(|(txid, _status)| invalidated_txids.contains(txid))
+                    .is_some()
```
Oscar-Pepper, line 198: Nice!

zingolib/src/wallet/transaction_records_by_id.rs
```diff
@@ -738,29 +724,32 @@ mod tests {
             .build();
         let spending_txid = transaction_record_later.txid;
 
+        let spend_in_known_tx = Some((spending_txid, Confirmed(15.into())));
+        // let spend_in_unknown_tx = Some((spending_txid, Confirmed(15.into())));
+
```
Oscar-Pepper, line 814: Can this be deleted?

zingolib/src/wallet/transaction_records_by_id.rs
```diff
@@ -738,29 +724,32 @@ mod tests {
...
-                SaplingNoteBuilder::default()
-                    .spent(Some((random_txid(), 15)))
+                    .spend(spend_in_known_tx)
                     .clone(),
             )
+            // .sapling_notes(
+            //     SaplingNoteBuilder::default()
+            //         .spend(spend_in_unknown_tx)
+            //         .clone(),
+            // )
```
Oscar-Pepper: Delete?

Oscar-Pepper (dismissed):
All the core changes here are excellent I just have a few comments that I'd like discussing before merge. The major blocker here is whether the wallet read write has been tested

zingolib/src/wallet/describe.rs
fluidvanadium: right. that function isnt supposed to be live yet.

zingolib/src/wallet/traits.rs
fluidvanadium: PASS [   0.017s] libtonode-tests::concrete slow::load_wallet_from_v26_dat_file
        PASS [   0.026s] libtonode-tests::concrete fast::reload_wallet_from_buffer
        PASS [   0.047s] libtonode-tests::concrete fast::load_and_parse_different_wallet_versions
        SLOW [> 60.000s] libtonode-tests::concrete slow::load_old_wallet_at_reorged_height
        PASS [  64.242s] libtonode-tests::concrete slow::load_old_wallet_at_reorged_height

zingolib/src/wallet/transaction_records_by_id.rs
fluidvanadium: i think there may be a more elegant version of these 4 lines. but i dont know it exactly

zingolib/src/wallet/notes/interface.rs
fluidvanadium: i dont know why this function is in this barnch ...

fluidvanadium: okay, i know why! because there was some really awful wet logic and this fixed it. but i can fix it even better

zancas:
@fluidvanadium what's the status of this PR?

fluidvanadium:
All tests pass, including reload and save tests

zingolib/src/wallet/notes/interface.rs
fluidvanadium: renamed spending_tx_status

libtonode-tests/tests/concrete.rs
```diff
@@ -273,8 +275,8 @@ mod fast {
         );
         assert_eq!(preshield_utxos[0].value, postshield_utxos[0].value);
         assert_eq!(preshield_utxos[0].script, postshield_utxos[0].script);
-        assert!(preshield_utxos[0].pending_spent.is_none());
-        assert!(postshield_utxos[0].pending_spent.is_some());
+        assert!(preshield_utxos[0].spending_tx_status().is_none());
```
zancas, line 278: Nice readability upgrade.

libtonode-tests/tests/concrete.rs
```diff
@@ -404,7 +406,7 @@ mod fast {
     }
 
     #[tokio::test]
-    async fn unspent_notes_are_not_saved() {
```
zancas, line 407: This test actually shows that the `spend_pending_transaction` is not saved as well, does it not?   Or, at least, it strongly implies it with the popped transaction check at the end.

zingolib/src/wallet/notes/interface.rs
```diff
@@ -42,35 +43,36 @@ pub trait OutputInterface: Sized {
...
     fn is_pending_spent(&self) -> bool {
-        self.pending_spent().is_some()
+        self.spending_tx_status()
+            .is_some_and(|(_txid, status)| status.is_pending())
     }
 
-    /// returns true if the note is confirmed spent
-    fn is_spent(&self) -> bool {
-        self.spent().is_some()
+    /// returns true if the note is spent and the spent is validated confirmed on chain
```
zancas: "and the spend"

zancas (dismissed):

zancas (approved):

PR #1274, opened 2024-07-06
unify pool order in records and add deprecations

Oscar-Pepper:


fluidvanadium (approved):
Deprecation notes are golden!

PR #1275, opened 2024-07-06
Fix total transparent value spent bug

Oscar-Pepper:
ontop of #1274 

read order was not aligned with write order. therefore, each time the app was closed and opened the values were saved in opposite positions causing "on and off" bugs like wrong fees and wrong kinds


fluidvanadium (approved):
Wow, who knows how long this has been swapped! Good catch

PR #1285, opened 2024-07-11
Patch output indexes

fluidvanadium:
should allow very_experienced wallets to send


zingolib/src/blaze/full_transactions_processor.rs
```diff
@@ -21,6 +21,7 @@ use zcash_primitives::{
 
 use zingo_status::confirmation_status::ConfirmationStatus;
 
+// despite being called fetch_full_transaction, this function sends a txid somewhere else to be fetched as a Transaction. then processes it.
```
Oscar-Pepper, line 24: is this comment now wrong since you renamed it?

zingolib/src/blaze/full_transactions_processor.rs
```diff
@@ -51,8 +52,7 @@ pub async fn start(
 
     let bsync_data_i = bsync_data.clone();
 
-    let (transaction_id_transmitter, mut transaction_id_receiver) =
-        unbounded_channel::<(TxId, BlockHeight)>();
+    let (txid_sender, mut transaction_id_receiver) = unbounded_channel::<(TxId, BlockHeight)>();
```
Oscar-Pepper: why rename transaction_id to txid on only the sender and not receiver??

zingolib/src/wallet/transaction_context.rs
```diff
@@ -31,6 +31,33 @@ impl TransactionContext {
             transaction_metadata_set,
         }
     }
+
+    /// returns any outdated records that need to be rescanned for completeness..
+    /// checks that each record contains output indexes for its notes
+    /// Result::Ok means no patches were made
```
Oscar-Pepper: what result?

zingolib/src/lightclient/sync.rs
```diff
@@ -429,22 +429,25 @@ impl LightClient {
...
-            fetch_full_transaction_transmitter,
-            fetch_taddr_transactions_transmitter,
-        ) = crate::blaze::fetch_full_transaction::start(
-            transaction_context,
-            full_transaction_fetcher_transmitter.clone(),
-            bsync_data.clone(),
-        )
-        .await;
+
+        // OUTPUT_INDEX_PATCH: collect any outdated transaction record that are incomplete and missing output indexes. and simply make them up. it wont go on chain so it doesnt matter
```
Oscar-Pepper, line 433: this is a really clever patch but im not convinced. the issue with this approach is that on rescan this output index will be corrected. so it could introduce a whole class of rescan bugs. one off the top of my head is a proposal is made, then the user rescans, and then confirms the transaction. the index has moved and the wrong notes are chosen, or the proposed index doesnt exist. as output index is one of the necessary co-ordinates to uniquely identify a note there may be many other cases in the future where we depend on a "chain correct" output index.

I think we should slow is smooth a targetted rescan of transactions with missing output indices

Oscar-Pepper (changes requested):

zingolib/src/lightclient/sync.rs
fluidvanadium: if you rescan, you must re-propose

zingolib/src/blaze/full_transactions_processor.rs
fluidvanadium: oops

fluidvanadium: i guess i renamed one in a different spot and missed its pair

zingolib/src/wallet/transaction_context.rs
fluidvanadium: h

PR #1286, opened 2024-07-11
unified add pending note with add new note

Oscar-Pepper:
pending bug was caused by regular mempool checks and/or calls to `scan_full_tx` with a `confirmed` status, as `pending` status was hard coded.

this PR removes `add_pending_note` and modifies `add_new_note` to account for both confirmed and pending so confirmed scans can't set a tx back to pending status.

also adds logic to make sure a pending status cannot override a confirmed one


fluidvanadium (dismissed):
WHOOO!!! net code decrease! what does it need to get merged?

zancas:
I vote we figure out how to add an automated test to this before we merge it.

PR #1289, opened 2024-07-13
fix pending bug by disallowing pending txs to overwrite confirmed tx …

Oscar-Pepper:
…status


fluidvanadium (approved):
good logic for record merge

PR #1299, opened 2024-07-17
implement send-to-self to be created alongside zfz sent value transfer

Oscar-Pepper:


fluidvanadium (changes requested):
overall good, but the zfz test doesnt seem to actually enable ZFZ

Oscar-Pepper:
> overall good, but the zfz test doesnt seem to actually enable ZFZ

the test is not asserting anything so its actually not finished but the zfz is active, thats the `true` on line 151

Oscar-Pepper:
ok the new address is added and test is finished, ready for review!

juanky201271 (approved):

PR #1302, opened 2024-07-22
Remove zfz memo and update doc comments

Oscar-Pepper:
- removes zenny for zingo memo from send all
- updates zip317 comments

### DIFF
--- a/zingolib/src/lightclient/send.rs
+++ b/zingolib/src/lightclient/send.rs
@@ -103,8 +103,7 @@ pub mod send_with_proposal {
     }
 
     impl LightClient {
-        /// Unstable function to expose the zip317 interface for development
-        // TODO: add correct functionality and doc comments / tests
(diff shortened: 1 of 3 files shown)


zingolib/src/lightclient/send.rs
```diff
@@ -240,8 +237,7 @@ pub mod send_with_proposal {
             Ok(self.complete_and_broadcast::<NoteId>(&proposal).await?)
         }
 
-        /// Unstable function to expose the zip317 interface for development
-        // TODO: add correct functionality and doc comments / tests
+        /// Shields all transparent funds without confirmation.
```
zancas, line 240: I don't understand what the `without confirmation` part means.

zancas: Non-blocking.

zancas (approved):

PR #1304, opened 2024-07-22
Backport stable zfz fix 3

Oscar-Pepper:


zingolib/src/lightclient/send.rs
```diff
@@ -241,8 +238,7 @@ pub mod send_with_proposal {
             Ok(self.complete_and_broadcast::<NoteId>(&proposal).await?)
         }
 
-        /// Unstable function to expose the zip317 interface for development
-        // TODO: add correct functionality and doc comments / tests
+        /// Shields all transparent funds without confirmation.
```
zancas, line 241: This is still confusing.

zancas (approved):
I like the inner fn.   We could helperize the memo aggregation.  Non-Blocking.

PR #1308, opened 2024-07-23
Ulp2

fluidvanadium:
updates the LRZ dependency and makes all necessary adjustments


fluidvanadium:
very strange CI failure ... it works locally

Oscar-Pepper (changes requested):
all the changes to address parsing can be resolved by just changing the output of `address_to_str` from Address -> ZcashAddress. I moved this parsing into commands to fix a bug and it was necessary so it needs moving back. Also i think `Receivers` is fine but it we are going to change to `Destinations` then we need to do this in a separate PR and update ALL instances of receiver so we don't end up with the transaction_metadata_set issue again

PR #1309, opened 2024-07-24
derive USK from seed

fluidvanadium:
This is the simplest way to achieve this. If this works we wont have to use the next layer.


zingolib/src/lightclient/send.rs
```diff
@@ -131,9 +129,13 @@ pub mod send_with_proposal {
             let (sapling_output, sapling_spend): (Vec<u8>, Vec<u8>) =
                 read_sapling_params().map_err(CompleteAndBroadcastError::SaplingParams)?;
             let sapling_prover = LocalTxProver::from_bytes(&sapling_spend, &sapling_output);
-            let unified_spend_key =
-                UnifiedSpendingKey::try_from(self.wallet.wallet_capability().as_ref())
-                    .map_err(CompleteAndBroadcastError::UnifiedSpendKey)?;
+
+            let unified_spend_key = UnifiedSpendingKey::from_seed(
```
zancas, line 134: @Oscar-Pepper how does this interact with taddresses that we've published from transparent keys in the past?

zancas (changes requested):
I don't understand the implications for backwards compatibility, or how we're testing for maintenance of functionality.

zancas:
I notice that there's a merge conflict on this.

fluidvanadium:
this doesnt work. perhaps the hack hasnt been given enough credit

PR #1314, opened 2024-07-29
Zingo sync add client module

Oscar-Pepper:
this approach will have the efficiency gains of the client being handled by a single task but will allow multiple other tasks to make requests concurrently. this also allows request prioritisation so small data fetches are not blocked on queued larger data fetches to complete and we can compare sync performance to varied prioritisation schemes.


zingo-sync/src/client/fetcher.rs
```diff
@@ -0,0 +1,143 @@
...
+/// prioritisation and fetching from the server.
+/// Returns the data specified in the [`self::FetchRequest`] variant via the provided [`tokio::sync::mpsc::oneshot::Sender`].
+///
+/// Allows all requests to the server to be handled from a single task for efficiency and also enables
+/// request prioritisation for further performance enhancement
+pub async fn fetcher(
+    mut fetch_request_receiver: UnboundedReceiver<FetchRequest>,
+    mut client: CompactTxStreamerClient<zingo_netutils::UnderlyingService>,
+) -> Result<(), ()> {
+    let mut fetch_request_queue: Vec<FetchRequest> = Vec::new();
```
idky137, line 27: Have you thought about using a mcsp::channel directly as the queue? It acts as a fifo list by default so actually works quite well as a queue. Sender threads push directly onto the channel and the fetcher pops them. I think there is some added complexity using a vec as a fifo list (better for filo)..

Oscar-Pepper: good point, i think we can use a more optimal queue for this reason but it is not designed to be FIFO, we want to prioritise requests. I think your idea about Ordering may be a good solution but as the backpressure should be very low I dont think we need to optimise this yet

zingo-netutils/src/lib.rs
```diff
@@ -16,7 +16,8 @@ use tonic::Status;
 use tower::util::BoxCloneService;
 use zcash_client_backend::proto::service::compact_tx_streamer_client::CompactTxStreamerClient;
 
-type UnderlyingService = BoxCloneService<
+/// TODO: add doc-comment
```
zancas, line 19: Why is this type being defined?

zingo-sync/src/client.rs
```diff
@@ -0,0 +1,49 @@
+//! Module for handling all connections to the server
+
+use std::ops::Range;
+
+use zcash_client_backend::proto::{compact_formats::CompactBlock, service::BlockId};
```
zancas, line 5: It's important to double check with zingolib to see if there's another definition of `CompactBlock` dangling around.

Oscar-Pepper: yeah good idea. i have checked and this is the same one used in zingolib currently and in LRZ etc. as defined in the proto files

zingo-netutils/src/lib.rs
Oscar-Pepper: well, you may have a better idea here but I think it's a lot to define without this type alias. any suggestions welcome

zancas:
> this approach will have the efficiency gains of the client being handled by a single task

What efficiency gain is this?

zingo-netutils/src/lib.rs
zancas: Please replace the TODO: with something else.

Oscar-Pepper: Yeah sure

Oscar-Pepper:
> > this approach will have the efficiency gains of the client being handled by a single task
> 
> What efficiency gain is this?

Me and @idky137 conducted a test on codeshare where two tasks were requesting blocks simultaneously and compared this to two tasks where only one was requesting the same total number of blocks. When only one task was connecting to the server the blocks came in faster

zingo-netutils/src/lib.rs
zancas: Oh I see it's just being made `pub`...

zingo-sync/src/client.rs
```diff
@@ -0,0 +1,49 @@
...
+) -> Result<BlockHeight, ()> {
+    let (sender, receiver) = oneshot::channel::<BlockId>();
+    fetch_request_sender
+        .send(FetchRequest::ChainTip(sender))
+        .unwrap();
+    let chain_tip = receiver.await.unwrap();
+
+    Ok(BlockHeight::from_u32(chain_tip.height as u32))
+}
+/// Gets the height of the blockchain from the server.
```
zancas, line 34: This doc-comment seems out of date... or associated with the wrong fn.

Oscar-Pepper: Yeah I noticed that, it's fixed in the following PR

zancas (approved):

PR #1321, opened 2024-08-07
LightWallet::load_example_wallet : consolidate example wallets

fluidvanadium:
The titular method `load_example_wallet` is a streamlined way to load an example wallet file for testing purposes.  The next step (I almost labeled it as a draft, but merging this could be good) is to move the rest of the test wallets into this new system and then create a _comprehensive_ test system.


zancas (approved):
Great Work @fluidvanadium !!!!  I love how you think!

PR #1322, opened 2024-08-07
Mv zingo testutils into zingolib

zancas:
This refactor pushes testutils, config, and testvectors into `zingolib` behind feature gates.

Reducing the number of crates in favor of conditional compilation makes the project less ergonomic for consumption by other crates, but it provides test-utilities private access to abstractions within `zingolib`.


fluidvanadium (approved):
bueno

PR #1319, opened 2024-08-06
Zingo sync calculate nullifiers and positions

Oscar-Pepper:
ontop of #1317

PR #1323, opened 2024-08-08
Zingo sync note commitments and retention

Oscar-Pepper:
ontop of #1319


zingo-sync/src/scan.rs
```diff
@@ -293,30 +277,120 @@ fn check_continuity(
...
                 .expect("key should be available as it was used to decrypt output");
             let nullifier = key
-                .nf(note, position)
+                .nf(&incoming_output.note, position)
                 .expect("only fvks currently supported");
             nullifiers_and_positions.insert(*output_id, (nullifier, position));
         });
 }
+
+// TODO: unify sapling and orchard leaf and retention fns
```
fluidvanadium, line 308: this function is begging for a unit test. if it is possible to create example versions of these structs

fluidvanadium (approved):

PR #1328, opened 2024-08-15
Scan cleanup

Oscar-Pepper:
ontop of #1327


zancas:
I am seeing CI failures, and I clicked update branch.

fluidvanadium:
i am going to review this set of PRs as a whole starting from here

zingo-sync/src/sync.rs
```diff
@@ -27,62 +31,107 @@ pub async fn sync<P, W>(
...
     tracing::info!("Syncing wallet...");
 
-    // TODO: add trait methods to read/write wallet data to/from sync engine
-    // this is where sync state would be read from wallet data
-    let sync_state = SyncState::new(); // placeholder
+    let mut handles = Vec::new();
 
     // create channel for sending fetch requests and launch fetcher task
-    let (fetch_request_sender, fetch_request_receiver) = unbounded_channel();
+    let (fetch_request_sender, fetch_request_receiver) = mpsc::unbounded_channel();
```
fluidvanadium, line 42: because, only because i have semantic saturation on the word 'fetch', this bit could use commentary

zingo-sync/src/scan.rs
```diff
@@ -101,27 +98,65 @@ impl InitialScanData<'_> {
...
+#[allow(dead_code)]
+struct ScanData {
+    pub(crate) nullifiers: NullifierMap,
+    pub(crate) wallet_blocks: BTreeMap<BlockHeight, WalletBlock>,
+    pub(crate) relevent_txids: HashSet<TxId>,
+    pub(crate) decrypted_note_data: DecryptedNoteData,
+    pub(crate) shard_tree_data: ShardTreeData,
+}
+
+pub(crate) struct ScanResults {
```
fluidvanadium, line 123: please include comments explanating the methodology and significance of these results

zingo-sync/src/scan/workers.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    pub(crate) fn is_worker_idle(&self) -> bool {
+        self.workers.iter().any(|worker| !worker.is_scanning())
+    }
+
+    pub(crate) fn add_scan_task(&self, scan_task: ScanTask) -> Result<(), ()> {
+        if let Some(worker) = self.workers.iter().find(|worker| !worker.is_scanning()) {
+            worker.add_scan_task(scan_task);
+        } else {
+            panic!("no idle workers!")
```
fluidvanadium, line 79: to be replaced with err handling

zingo-sync/src/sync.rs
```diff
@@ -127,14 +176,84 @@ fn prepare_next_scan_range(sync_state: &SyncState) -> Option<ScanRange> {
     if let Some((lower_range, higher_range)) = selected_scan_range
         .split_at(selected_scan_range.block_range().start + BlockHeight::from_u32(BATCH_SIZE))
     {
-        scan_ranges.splice(index..=index, vec![lower_range.clone(), higher_range]);
-
-        // TODO: set selected scan range to `ignored` so the same range is not selected twice when multiple tasks call this fn
+        let lower_range_ignored =
```
fluidvanadium, line 179: okay, i am understanding ScanPriority::Ignored here means [currently in scan]

zingo-sync/src/sync.rs
```diff
@@ -127,14 +176,84 @@ fn prepare_next_scan_range(sync_state: &SyncState) -> Option<ScanRange> {
...
+        scan_ranges.splice(index..=index, vec![selected_range_ignored]);
 
         Some(selected_scan_range.clone())
     }
 }
+
+fn mark_scanned(scan_range: ScanRange, sync_state: &mut SyncState) -> Result<(), ()> {
+    let scan_ranges = sync_state.scan_ranges_mut();
+
+    if let Some((index, range)) = scan_ranges
```
fluidvanadium, line 199: there has gotta be a better way to do this!

fluidvanadium: (eventually)

zingo-sync/src/sync.rs
```diff
@@ -27,62 +31,107 @@ pub async fn sync<P, W>(
...
-        )
-        .await
-        .unwrap();
+    drop(scanner);
+    while let Some((scan_range, scan_results)) = scan_results_receiver.recv().await {
+        update_wallet_data(wallet, scan_results).unwrap();
+        // TODO: link nullifiers and scan linked transactions
+        remove_irrelevant_data(wallet, &scan_range).unwrap();
+        mark_scanned(scan_range, wallet.get_sync_state_mut().unwrap()).unwrap();
     }
```
fluidvanadium, line 105: DRY this block?

zingo-sync/src/sync.rs
```diff
@@ -27,62 +31,107 @@ pub async fn sync<P, W>(
...
+            Ok((scan_range, scan_results)) => {
+                update_wallet_data(wallet, scan_results).unwrap();
+                // TODO: link nullifiers and scan linked transactions
+                remove_irrelevant_data(wallet, &scan_range).unwrap();
+                mark_scanned(scan_range, wallet.get_sync_state_mut().unwrap()).unwrap();
+            }
+            Err(TryRecvError::Empty) => (),
+            Err(TryRecvError::Disconnected) => break,
+        }
+    }
```
fluidvanadium, line 97: at this point IIUC, either the connection broke or all ranges are scanned.
are we now making assumptions about how many tuples are left in the receiver? couldnt it still be more than one? or shouldnt it be 0?

zingo-sync/src/sync.rs
fluidvanadium, line 42: because, only because i have semantic saturation on the word 'fetch', this bit could use commentary explaining how the channels interface

zingo-sync/src/scan/workers.rs
fluidvanadium, line 79: to be replaced with err handling

zingo-sync/src/sync.rs
fluidvanadium, line 97: at this point IIUC, either the connection broke or all ranges are scanned.
are we now making assumptions about how many tuples are left in the receiver? couldnt it still be more than one? or shouldnt it be 0?

zingo-sync/src/sync.rs
fluidvanadium, line 105: DRY this block?

zingo-sync/src/sync.rs
fluidvanadium, line 179: okay, i am understanding ScanPriority::Ignored here means [currently in scan]

zingo-sync/src/sync.rs
fluidvanadium, line 199: there has gotta be a better way to do this! eventually

not just because of mark. but because == comparison on the range seems weird, and because rebuilding it instead of changing its priority seems weird, and because panicking is not great, and because what if the ranges overlappp?

zingo-sync/src/scan.rs
fluidvanadium, line 123: this could use some comments explaning the methodology and significance of these ScanResults

fluidvanadium (dismissed):
overall, extremely well organized

zingo-sync/src/sync.rs
Oscar-Pepper: ok thanks i will try to add clarity in the following PR. this channel can essentially be abstracted as how the sync engine connects to the server

zingo-sync/src/scan/workers.rs
Oscar-Pepper: yes! all panics and asserts and unwraps will definately be replaced after PoC

zingo-sync/src/sync.rs
Oscar-Pepper: yes if we keep this repeated part for shutdown we can DRY, i havent yet because i dont know if this architecture will stay and I also dont know what this block really is yet

zingo-sync/src/sync.rs
Oscar-Pepper: thats right you are following this really well, so LRZ probably had a slightly different vision for scan ranges but in the way we are using them `ignored` is not used and `scanning` does not exist. for now, ignored means scanning until we maybe PR into LRZ or create our own copy

zingo-sync/src/sync.rs
Oscar-Pepper: yes thats right. if all senders are closed, `try_recv` will still return Ok until the channel is empty and then it will hit the Err(Disconnected) and break the loop. although this will not happen the way the code is currently written as in the happy path the break on line 83 will be hit when all ranges are scanned (except the ones currently scanning). in that case we break and the while loop will process that remaining results being scanned. this break on line 95 will still be useful if we decide to close the workers with the scan_results_senders but if its just an error case then we will change this to return an error

zingo-sync/src/sync.rs
Oscar-Pepper: i am open to suggestions about the comparison but i dont see an issue, it is essentially just two numbers that we must make sure are identical.

i agree rebuilding to change priority could be improved, again we will have to open a PR into LRZ or copy it into our lib.

in terms of overlapping. only one scan range is ever created, from birthday to chain height. then this scan range is only split so no overlapping or gaps are possible, as long as we unit test the split and splicing functions properly. the only exception is when ranges are rebuilt to change its priority but the ranges are copied so we can assume this will never be bugged and should also be tested eventually

zingo-sync/src/scan.rs
Oscar-Pepper: sure, when the scan code is finished i will make sure the comments are thorough. i dont want to waste too much time yet as its still ion flux so may change

PR #1330, opened 2024-08-19
Owls 0.2

fluidvanadium:
creates a new schema for loading and testing example wallets. adds two tests. so far, they verify example wallets
1} have the expected chain
2} deserialize correctly


Oscar-Pepper (approved):
this looks like great back compatibility framework! good job!

PR #1331, opened 2024-08-20
Zingo sync scan transaction

Oscar-Pepper:
ontop of #1328


zingo-sync/src/scan.rs
```diff
@@ -524,3 +545,143 @@ fn collect_nullifiers(
...
+    parameters: &P,
+    scanning_keys: &ScanningKeys,
+    relevant_txids: HashSet<TxId>,
+    decrypted_note_data: DecryptedNoteData,
+    wallet_blocks: &BTreeMap<BlockHeight, WalletBlock>,
+) -> Result<HashMap<TxId, WalletTransaction>, ()>
+where
+    P: Parameters,
+{
+    let mut wallet_transactions = HashMap::with_capacity(relevant_txids.len());
```
fluidvanadium, line 568: because a txid may include multiple transactions, this here must be a HashSet rather than a Hashmap

zingo-sync/src/scan.rs
```diff
@@ -524,3 +545,143 @@ fn collect_nullifiers(
...
+        scan_notes::<OrchardDomain, Action<Signature<SpendAuth>>, OrchardNote>(
+            &mut orchard_notes,
+            transaction.txid(),
+            &orchard_keys,
+            &orchard_actions,
+            &decrypted_note_data.orchard_nullifiers_and_positions,
+        )
+        .unwrap();
+    }
+
```
fluidvanadium, line 661: // TODO: scan transparent?

zingo-sync/src/scan.rs
```diff
@@ -524,3 +545,143 @@ fn collect_nullifiers(
...
+    Ok(WalletTransaction::from_parts(
+        transaction.txid(),
+        block_height,
+        sapling_notes,
+        orchard_notes,
+    ))
+}
+
+fn scan_notes<D, Op, N>(
+    wallet_notes: &mut Vec<N::WalletNote>,
```
fluidvanadium, line 671: why do we need to pass a mutable vector here? I see that the input is empty. instead of the extra parameter just initialize an empty vector in the function

fluidvanadium (dismissed):
I think it needs a couple edits. i like how its moving.

zingo-sync/src/scan.rs
Oscar-Pepper: a txid is a unique id for a single transaction. wallet_transactions is the equivalent of transaction_records_by_id in zingolib which is also a hashmap

zingo-sync/src/scan.rs
Oscar-Pepper: indeed! i have added a TODO to my latest PR

zingo-sync/src/scan.rs
Oscar-Pepper: this way is the simplest and cleanest because the vec is initialised in the right scope. if the vec was created by the scan_notes fn then we would have to return it out of the if let block. this gets more complicated as there will also be outgoing notes and encoded memos that also need returning from the block

zingo-sync/src/scan.rs
Oscar-Pepper: just so you are aware hashsets also are unique lists with no duplicates

zancas:
I pushed a resolution of merge conflicts with latest dev.

zancas (dismissed):

zingo-sync/src/scan.rs
zancas: I share @fluidvanadium 's concern here.   The return of this fn isn't very useful.

This C-style pattern of passing in mutable references and then messing with the state outside the fn doesn't leverage the type system to constrain what can happen.

zancas (approved):
This PR is the first part of a new feature.  The zingo-sync feature is not yet integrated into the codebase, so it's safe to merge this code into dev.   Review has shown some interesting directions we might take, but we know there's more code to review in subsequent PRs, which are also chunks of the feature.

PR #1334, opened 2024-08-21
Zingo sync parse memos

Oscar-Pepper:
ontop of #1331


zingo-sync/src/scan.rs
```diff
@@ -676,20 +682,143 @@ fn scan_notes<D, Op, N>(
...
+
             let wallet_note = N::from_parts(output_id, note, *nullifier, *position, memo);
             wallet_notes.push(wallet_note);
         }
     }
 
     Ok(())
 }
+
+fn parse_encoded_memos<P, N>(
```
zancas, line 708: This belongs in zingo-memo.  I'll look into migrations.

zancas (approved):
This version of the code does not properly handle zingo-memo::ParsedMemo::Version1.  We'll need to fully test and handle that before this feature can get into production.

zancas:
I am concerned about dependent PRs not being landable because of this, so I am landing it with a todo which is to upgrade to modern memo handling.

PR #1335, opened 2024-08-21
Zingo sync scan modules

Oscar-Pepper:
on top of  #1334


zingo-sync/src/client.rs
```diff
@@ -49,22 +54,22 @@ pub async fn get_compact_block_range(
     fetch_request_sender: UnboundedSender<FetchRequest>,
     block_range: Range<BlockHeight>,
 ) -> Result<Vec<CompactBlock>, ()> {
-    let (sender, receiver) = oneshot::channel::<Vec<CompactBlock>>();
+    let (sender, receiver) = oneshot::channel();
```
zancas: What (if anything) is the benefit of removing the turbo-fish?  Does clippy prefer it?

zancas: I suppose it's pretty obvious.

Oscar-Pepper: its not necessary and improved my workflow as the same line can be used for all types but im not against adding turbofish back in if you prefer

zancas (approved):

PR #1339, opened 2024-08-23
Zingo sync retain relevant blocks

Oscar-Pepper:
on top of #1338


zingo-sync/src/traits.rs
```diff
@@ -57,6 +57,9 @@ pub trait SyncBlocks: SyncWallet {
 
 /// Trait for interfacing [`crate::primitives::WalletTransaction`]s with wallet data
 pub trait SyncTransactions: SyncWallet {
+    /// Get mutable reference to wallet transactions
```
AloeareV:
```suggestion
    /// Get reference to wallet transactions
```

zancas (approved):

PR #1342, opened 2024-08-27
simplified parameters

fluidvanadium:
eventually, if we remove every reference to ZingoConfig, it will be possible to create a wallet without specifying a single server


Oscar-Pepper (approved):
lavly

PR #1345, opened 2024-08-28
Owls a.1

fluidvanadium:


zingolib/src/lightclient/send.rs
```diff
@@ -275,5 +285,18 @@ pub mod send_with_proposal {
                     .to_string(),
             );
         }
+
+        #[ignore]
```
zancas, line 289: What's the point of adding an ignored test?

zancas (approved):

PR #1347, opened 2024-08-28
LightWallet can propose

fluidvanadium:
moves methods from LightClient that oughta be in LightWallet


zancas (approved):
I reformatted some conflicting use statements to the standard formats, and `use`d errors from a common origin.

zancas:
Merging based on local test assurance.

PR #1348, opened 2024-08-29
Zingo sync reorg logic

Oscar-Pepper:
atop #1343


zancas:
I clicked update branch.

zancas:
I noticed that CI was failing, and that this branch was out of date, so I clicked "Update Branch".

Oscar-Pepper:
> I noticed that CI was failing, and that this branch was out of date, so I clicked "Update Branch".

Oh I forgot to run clippy! Ok fixing now

libtonode-tests/tests/sync.rs
```diff
@@ -31,7 +31,7 @@ async fn sync_mainnet_test() {
     let mut lightclient = LightClient::create_from_wallet_base_async(
         WalletBase::from_string(HOSPITAL_MUSEUM_SEED.to_string()),
         &config,
-        2_611_700,
+        2_715_150,
```
zancas, line 34: If this is, for example a NU height, or protocol epoch, I think it would be best to import the most general relevant constant and use it explicitly.

If it's an HMS-specific birthday, then maybe  we should associate it with that seed in some ergonomic way, so that e.g. other tests can get the seed-associated bday? 

If it's not, for example if it's a birthday that's idiosyncratic to this test, then let's throw some kind of comment on it, specifying that.

libtonode-tests/tests/sync.rs
```diff
@@ -77,8 +77,8 @@ async fn sync_test() {
     .await
     .unwrap();
```
zancas, line 79: I thought we wanted to cut dbgs out?  I guess I am ok with commented out dbgs in test code.  Is that our practice?

zingo-sync/src/primitives.rs
```diff
@@ -35,6 +35,15 @@ impl SyncState {
             spend_locations: Vec::new(),
         }
     }
+
```
zancas, line 38: From local checks this fn could currently be `pub(crate)`..  should we promote it to fully `pub` "on demand"?

Do we already have a consumer using it?

zingo-sync/src/scan.rs
```diff
@@ -166,8 +169,7 @@ where
     .await
     .unwrap();
 
-    let scan_data =
-        scan_compact_blocks(compact_blocks, parameters, ufvks, initial_scan_data).unwrap();
+    let scan_data = scan_compact_blocks(compact_blocks, parameters, ufvks, initial_scan_data)?;
```
zancas, line 172: Nice.

zancas (approved):
There are a number of private methods introduced in this PR, for example the "truncate" family.

What is our strategy for adding focused single-fn unit tests?

What about test-coverage information for the zingo-sync crate?

libtonode-tests/tests/sync.rs
Oscar-Pepper: This test will be removed eventually it is ignored and I only use it sometimes when a larger chain is needed. This birthday is just a block height near to the server tip at time of writing.

libtonode-tests/tests/sync.rs
Oscar-Pepper: For sync development its good to have these dbgs ready to go when needed. I don't think we should normally allow debug but that's why I created a separate sync.rs test suite

zingo-sync/src/primitives.rs
Oscar-Pepper: I thought this could be useful API but Im happy with making it pub(crate), i will make this change on the following PR. I think it's a good policy to keep things private unless there is a known use case for sure

Oscar-Pepper:
> There are a number of private methods introduced in this PR, for example the "truncate" family.
> 
> What is our strategy for adding focused single-fn unit tests?
> 
> What about test-coverage information for the zingo-sync crate?

before zingo sync is released all error handling and a large % test coverage will be added.

as unit tests have to change when a fn changes I want to write a PoC first to be more time efficient

PR #1362, opened 2024-09-08
set nextest install action to v2

zancas:
I explicitly set nextest to be the version that I am running locally.
Running CI to check for non-breaking update.


fluidvanadium (approved):
good upgrade

PR #1367, opened 2024-09-08
Mini dep updates 3: Update Network Stack

fluidvanadium:


zancas:
It seems likely to me that the dependency version conflict currently blocking CI would be resolved by depending on `zip339`.

fluidvanadium:
> It seems likely to me that the dependency version conflict currently blocking CI would be resolved by depending on `zip339`.

zcash_primitives::...::zip339 which used to be a reexport of bip0039 no longer exists

fluidvanadium:
force push reverted Cargo.lock accidental update

fluidvanadium:
reopen

fluidvanadium:
reopen and comment

PR #1380, opened 2024-09-10
TxMap may contain SpendingData

fluidvanadium:
this is a rename and then a change to struct hierarchy. no effects yet


zingolib/src/wallet/tx_map/spending_data.rs
```diff
@@ -0,0 +1,20 @@
+//! the subsection of TxMap that only applies to spending wallets
+
+use crate::data::witness_trees::WitnessTrees;
+
+/// the subsection of TxMap that only applies to spending wallets
+pub(crate) struct SpendingData {
+    witness_trees: WitnessTrees,
+}
+
+impl SpendingData {
```
AloeareV, line 10: Why add an additional layer of type wrapper? Why disambiguate between a WitnessTrees, and a SpendingData?

zingolib/src/wallet/tx_map.rs
```diff
@@ -1,25 +1,26 @@
...
+use spending_data::SpendingData;
 use std::{fmt::Debug, sync::Arc};
 use thiserror::Error;
 use zcash_primitives::legacy::TransparentAddress;
 
 /// HashMap of all transactions in a wallet, keyed by txid.
 /// Note that the parent is expected to hold a RwLock, so we will assume that all accesses to
 /// this struct are threadsafe/locked properly.
-pub struct TxMapAndMaybeTrees {
+pub struct TxMap {
```
AloeareV, line 20: I think this name is less clear than TxMapAndMaybeTrees. What's the reasoning here?

AloeareV (commented):
I don't understand the intent behind this PR.

zingolib/src/wallet/tx_map.rs
fluidvanadium: is Map the same as sById?

AloeareV: I think so, yes

fluidvanadium:
the goal here is to prepare the ground for #1382 that adds another Spend-only field to SpendingData, behind the same option

AloeareV (approved):

zingolib/src/wallet/tx_map.rs
fluidvanadium: okay, darn it, im sorry, i guess i picked the wrong name again. i regret it, darn oops. sorry. maybe we can revert it later

PR #1381, opened 2024-09-11
simplified broadcast fn

fluidvanadium:


AloeareV (approved):
This is a removal of a feature (a consumer being able to specify a custom broadcast function) that I'm not aware of us ever having used, since we inherited it from aditya...and in fact I think the feature was 'removed' before now in that it stopped being part of our public interface, and this is just deleting the artifact left behind.

PR #1384, opened 2024-09-12
LRZ update: clarify LRZ rust version 1.77.0

fluidvanadium:


Oscar-Pepper (changes requested):
i suggest we close this PR and start working from an updated LRZ fork. i have proposed `main_zingolib` in element. process still up for discussion

PR #1400, opened 2024-09-17
Lrz update address from str

zancas:


Oscar-Pepper:
apologies this was already fixed but pushed to my branch instead of zingolabs!

Issue #1389, opened 2024-09-13
Run-time problem in Android - `rustls` upgrade related.


juanky201271:
History:
1. in some near past someone upgraded `rustls` to `0.23.12` if I recall correctly.
This update have two new crates: `aws-ls-rs` & `aws-ls-sys`, AFAIK this crate `rustls` have 2 different cryptographic backends:
- the old one: `ring`.
- the new one: `aws`.

This upgrade introduce something new, When you use: `ClientConfig::builder()` -> https://github.com/zingolabs/zingolib/blob/dev/zingo-netutils/src/lib.rs#L109 for instance... before you have to choose one backend between `ring` and `aws` or select the backend by default `CryptoProvider::install_default()` which I believe is `aws`.

Nobody knows about this until now... I guess.

2. I had some issues cross-compiling zingolib, commit: `https://github.com/zingolabs/zingolib/commit/6454707abf12671f8935c5937ba25e51ee2bace9` because of these libs from `aws`. I wondered what really will bring to us, and I find out...

3. I had the same error in both Platforms: 
![image](https://github.com/user-attachments/assets/84f6c5b2-cb54-44ec-b9bd-1579f165e3c9)
![Image 9-13-24 at 8 32 AM](https://github.com/user-attachments/assets/09156759-8fe0-4fa4-8824-1deb0a5abe27)

4. some research:
I found this:
```
Support for process-wide selection of CryptoProviders. See [the documentation](javascript:;). Note that callers of ClientConfig::builder(), ServerConfig::builder(), WebPkiServerVerifier::builder() and WebPkiClientVerifier::builder() must now ensure that the crate's features are unambiguous or explicitly select a process-level provider using CryptoProvider::install_default(). Otherwise, these calls will panic with:

    no process-level CryptoProvider available -- call CryptoProvider::install_default() before this point

We recommend that libraries rely on the process-level provider by default, and that applications use this new API to select the provider they wish to use.
```
Ref: https://mygit.osfipin.com/release/144217327

and this doc: https://rustls.github.io/rustls/prerelease/crypto/struct.CryptoProvider.html#using-the-per-process-default-cryptoprovider

and a poor example that how to solve this issue: https://github.com/zydou/arti/blob/c21370400bd52d670dcfac79b18901467d6ffa36/crates/tor-rtcompat/src/impls/rustls.rs#L103C9-L115C10

[...]

PR #1401, opened 2024-09-17
Mempool result

nachog00:
Got 32 failed tests while running `cargo nextest`.
``` bash
------------
     Summary [ 791.275s] 309 tests run: 277 passed (59 slow), 32 failed, 21 skipped
        FAIL [ 107.824s] libtonode-tests::chain_generics chain_generics::simpool_change_10_000_orchard_to_orchard
        FAIL [ 110.085s] libtonode-tests::chain_generics chain_generics::simpool_change_10_000_orchard_to_sapling
        FAIL [ 132.373s] libtonode-tests::chain_generics chain_generics::simpool_change_10_000_orchard_to_transparent
        FAIL [ 107.074s] libtonode-tests::chain_generics chain_generics::simpool_change_10_000_sapling_to_orchard
        FAIL [  85.808s] libtonode-tests::chain_generics chain_generics::simpool_change_10_000_sapling_to_sapling
        FAIL [  85.983s] libtonode-tests::chain_generics chain_generics::simpool_change_10_000_sapling_to_transparent
        FAIL [ 116.079s] libtonode-tests::chain_generics chain_generics::simpool_change_50_000_orchard_to_orchard
        FAIL [  95.346s] libtonode-tests::chain_generics chain_generics::simpool_change_50_000_orchard_to_sapling
        FAIL [  95.879s] libtonode-tests::chain_generics chain_generics::simpool_change_50_000_orchard_to_transparent
        FAIL [  99.074s] libtonode-tests::chain_generics chain_generics::simpool_change_50_000_sapling_to_orchard
        FAIL [  91.380s] libtonode-tests::chain_generics chain_generics::simpool_change_50_000_sapling_to_sapling
        FAIL [ 112.936s] libtonode-tests::chain_generics chain_generics::simpool_change_50_000_sapling_to_transparent
        FAIL [  98.107s] libtonode-tests::chain_generics chain_generics::simpool_change_50_orchard_to_orchard
        FAIL [  99.500s] libtonode-tests::chain_generics chain_generics::simpool_change_50_orchard_to_sapling
        FAIL [ 121.109s] libtonode-tests::chain_generics chain_generics::simpool_change_50_orchard_to_transparent
        FAIL [  97.342s] libtonode-tests::chain_generics chain_generics::simpool_change_50_sapling_to_orchard
        FAIL [  87.854s] libtonode-tests::chain_generics chain_generics::simpool_change_50_sapling_to_sapling
[...]

PR #1403, opened 2024-09-18
Rustls upgrade fix

nachog00:
This is the simplest solution that gets rid of the error:  #1389 

```bash
thread 'main' panicked at /home/chona/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rustls-0.23.12/src/crypto/mod.rs:259:14:
no process-level CryptoProvider available -- call CryptoProvider::install_default() before this point
```

on `cargo run`.

I'm not sure if `ring` is the option we want nor if `run_cli` is the right place to call that install.


zancas:
Have you run the full test suite against this feature locally?

If yes, did all the tests pass?

zancas:
This is blocked on #1401

zingocli/src/lib.rs
```diff
@@ -513,6 +514,11 @@ fn dispatch_command_or_start_interactive(cli_config: &ConfigTemplate) {
 
 /// TODO: Add Doc Comment Here!
```
zancas, line 515: Upgrade the doc-comment, to something better.

Please add a doc-test.

zingocli/src/lib.rs
```diff
@@ -513,6 +514,11 @@ fn dispatch_command_or_start_interactive(cli_config: &ConfigTemplate) {
...
 pub fn run_cli() {
+    // install default crypto provider (ring)
+    rustls::crypto::ring::default_provider()
+        .install_default()
+        .expect("Failed to install rustls crypto provider");
```
zancas: In general `expect` and `unwrap` are not allowed in our codebase.  Should this function return Result?

zancas (dismissed):

fluidvanadium:
> This is blocked on #1401

not exactly. however
allowing Nacho to run the full test suite is maximum priority and will have to happen before we can expect him to perfect this fix

zingocli/src/lib.rs
Oscar-Pepper: this is the entrypoint to the CLI so i'm not sure we can add a doc-test for this fn

zingocli/src/lib.rs
```diff
@@ -11,6 +11,7 @@ use std::sync::Arc;
 use log::{error, info};
 
 use clap::{self, Arg};
+use rustls;
```
Oscar-Pepper, line 14: this line should not be necessary as you are calling `rustls::crypto::ring::default_provider().install_default()` with its full path

Oscar-Pepper (changes requested):
Sorry I was not aware of the state of this PR when I wrote #1426. This looks good but we will also have to delete the crypto provider install from netutils as this is called multiple times which causes an error. Would you mind if we closed this for #1426 ?

nachog00:
> Sorry I was not aware of the state of this PR when I wrote #1426. This looks good but we will also have to delete the crypto provider install from netutils as this is called multiple times which causes an error. Would you mind if we closed this for #1426 ?

I'm fine with closing this one. Didn't do it already yesterday only cause @zancas  asked me to add a test, documenting and some other stuff on #1426 and I wasnt sure how to collaborate on your PR. What is our flow in those cases? commit directly on it? PR on it? So far i got working on a local branch and was planning on PRing into yours.

Oscar-Pepper:
> > Sorry I was not aware of the state of this PR when I wrote #1426. This looks good but we will also have to delete the crypto provider install from netutils as this is called multiple times which causes an error. Would you mind if we closed this for #1426 ?
> 
> I'm fine with closing this one. Didn't do it already yesterday only cause @zancas asked me to add a test, documenting and some other stuff on #1426 and I wasnt sure how to collaborate on your PR. What is our flow in those cases? commit directly on it? PR on it? So far i got working on a local branch and was planning on PRing into yours.

I commented on your PR as I think a doc test is difficult in this case but good practice for functions that can be unit tested. So in most cases the flow would be to PR into my branch. However, in this case I am happy for you to just PR into dev and I'll close mine or push directly to my branch if your have collaborator rights (I'm not sure if you being part of the organisation allows this by default). It's up to you really

Oscar-Pepper:
closed for #1426

PR #1406, opened 2024-09-18
cargo update

fluidvanadium:
Cargo update allows us to get newer versions of dependencies. They might be safer or faster.

Unfortunately this breaks nextest CI. trying to fix this. This PR is not ready until it does, but marking it as ready is necessary to trigger CI run to test if it works.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,6 +33,8 @@ jobs:
         uses: actions/checkout@v3
 
       - uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: 1.81
 
       - name: Install protoc
         run: sudo apt-get install protobuf-compiler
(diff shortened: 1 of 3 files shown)


Oscar-Pepper (changes requested):
@zancas recently had a PR that removed this so there is only one source of truth for version, the rust-toolchain file. could you write in the PR description what issue this is solving so we can have a dsicussion?

fluidvanadium:
> @zancas recently had a PR that removed this so there is only one source of truth for version, the rust-toolchain file. could you write in the PR description what issue this is solving so we can have a dsicussion?

description improved

fluidvanadium:
when i run cargo install cargo-nextest locally, i get the following line:
`Compiling guppy v0.17.7`
in contrast to this here error
`Location: /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/guppy-0.17.1/src/graph/build.rs:749`

Oscar-Pepper:
CI is working fine now

PR #1407, opened 2024-09-18
re-add comment about spend_status

fluidvanadium:


Oscar-Pepper:
I commented on this in your previous PR. this is incorrect, the transaction is also infact removed

PR #1415, opened 2024-09-23
Remove assumptions that caused a thread panic

fluidvanadium:
minimal sane fix


zancas (commented):
Should ***something*** explicit happen in the Err case?

Oscar-Pepper:
> Should _**something**_ explicit happen in the Err case?

No, there are cases where the tx will not be found in the wallet so we should not error here

Oscar-Pepper (approved):

PR #1420, opened 2024-09-23
added LightWallet::get_first_address

fluidvanadium:
before this PR merges, all methods to receive an address are methods of LightClient. even though a LightWallet contains an address. 

fixes above.


zingolib/src/wallet/describe.rs
```diff
@@ -302,6 +302,14 @@ impl LightWallet {
             .map(|(_index, sk)| *sk)
             .collect::<Vec<_>>()
     }
+
+    /// gets the basic receiver for the wallet. this is the only receiver implemented as 2024-09-22
```
pacu: What does "basic receiver" mean?

Note that this will always return the first element of the iterator. One could assume that it has some kind of order, but it may not be the case, so the returned address might not be the one intended.

pacu (changes requested):

zingolib/src/wallet/describe.rs
fluidvanadium: thanks for the review! I will adjust.

Oscar-Pepper (commented):
i left comments on parent PR #1424

zingolib/src/wallet/describe.rs
zancas: This is resolved, in the future let's say Option<foo> or optionally returns (instead of Some(foo) and None).

zingolib/src/utils.rs
```diff
@@ -31,10 +32,11 @@ macro_rules! build_push_list {
     };
 }
 
+#[cfg(any(test, feature = "test-elevation"))]
```
zancas, line 35: At some point can we combine these in way that permits a single `cfg` attribute to gate all of them?

zancas (approved):

PR #1421, opened 2024-09-23
Added Tests of Addresses in each pool for existing Test Example Wallets

fluidvanadium:
adds tests confirming that addresses generated in each example wallet do not change


Oscar-Pepper (commented):
i left comments on parent PR #1424

zancas (approved):

PR #1423, opened 2024-09-23
Test that sending with insufficient funds returns a useful error.

fluidvanadium:
child of #1420 

these tests are rare and necessary

PR #1424, opened 2024-09-24
MainNet hot wallet send to send Generic test

fluidvanadium:
this PR is a child of #1421


Oscar-Pepper (changes requested):
see comments on #1423

PR #1426, opened 2024-09-24
move crypto provider installer to be called once

Oscar-Pepper:
this PR must be followed by an addition to zingo-mobile rust to also call this installer on app launch

### DIFF
--- a/libtonode-tests/tests/concrete.rs
+++ b/libtonode-tests/tests/concrete.rs
@@ -3466,7 +3466,7 @@ mod slow {
         let (regtest_manager, _cph, faucet, recipient) =
             scenarios::faucet_recipient_default().await;
         for i in 1..4 {
-            let _ = faucet.do_sync(false).await;
+            faucet.do_sync(false).await.unwrap();
             from_inputs::quick_send(
                 &faucet,
                 vec![(&get_base_address_macro!(recipient, "sapling"), 10_100, None)],
@@ -3476,7 +3476,7 @@ mod slow {
             let chainwait: u32 = 6;
             let amount: u64 = u64::from(chainwait * i);
             zingolib::testutils::increase_server_height(&regtest_manager, chainwait).await;
-            let _ = recipient.do_sync(false).await;
+            recipient.do_sync(false).await.unwrap();
             from_inputs::quick_send(
                 &recipient,
                 vec![(&get_base_address_macro!(recipient, "unified"), amount, None)],
--- a/zingocli/src/main.rs
+++ b/zingocli/src/main.rs
@@ -1,4 +1,8 @@
 #![forbid(unsafe_code)]
 pub fn main() {
+    // install default crypto provider (ring)
+    if let Err(e) = rustls::crypto::ring::default_provider().install_default() {
+        eprintln!("Error installing crypto provider: {:?}", e)
+    };
     zingo_cli::run_cli();
 }
--- a/zingolib/src/lightclient/sync.rs
+++ b/zingolib/src/lightclient/sync.rs
@@ -681,11 +681,12 @@ pub mod test {
         std::env::set_var("RUST_BACKTRACE", "1");
         let wallet = wallet_case.load_example_wallet().await;
         let lc = LightClient::create_from_wallet_async(wallet).await.unwrap();
-        let _ = lc.do_sync(true).await;
+        lc.do_sync(true).await.unwrap();
         lc
     }
 
     /// this is a live sync test. its execution time scales linearly since last updated
+    #[ignore = "testnet and mainnet tests should be ignored due to increasingly large execution times"]
     #[tokio::test]
     async fn testnet_sync_mskmgdbhotbpetcjwcspgopp_latest() {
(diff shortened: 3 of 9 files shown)


juanky201271:
I added the FFI function in zingo-mobile:
https://github.com/zingolabs/zingo-mobile/pull/696/commits/62d68d765e948e1810af02f25d79b3ad980404bf
And it's working, no run-time error about the Crypto Provider.

fluidvanadium:
i am proud of these sync tests. i understand that they have variable run time, but i do not understand why they would be ignored

nachog00 (commented):
Whats our stance on expects and unwraps? Do we allow them in tests only?
@zancas @Oscar-Pepper

Oscar-Pepper:
> i am proud of these sync tests. i understand that they have variable run time, but i do not understand why they would be ignored

as you said, we cant have tests unignored that are guaranteed to slow down test run time, they will just get ignored some time in the near future  anyway and cost a load of CI / local test developer time. this is what darkside is for, we have chain load framework. the only limitation I am aware of is transparent

Oscar-Pepper:
> Whats our stance on expects and unwraps? Do we allow them in tests only? @zancas @Oscar-Pepper

unwraps are test-only. expects can be acceptable in production with a message that eludes to why it should NEVER panic.

zancas:
In order to guarantee non-panic on an `expect` in production, we'd need to add wrapping correctness guarantees.   That's not a proper effort for a developer unless they're already **expert** in the code base.

Please do **NOT** add `expect`s to production code.

zancas:
> > i am proud of these sync tests. i understand that they have variable run time, but i do not understand why they would be ignored
> 
> as you said, we cant have tests unignored that are guaranteed to slow down test run time, they will just get ignored some time in the near future anyway and cost a load of CI / local test developer time. this is what darkside is for, we have chain load framework. the only limitation I am aware of is transparent

I disagree.   This is clear evidence that we do not have sufficient resources allocated to CI.   We should *not* degrade our test quality because the free tier of github runners is too slow.   That's the wrong direction.   We need to invest in test infrastructure.

zingocli/src/main.rs
```diff
@@ -1,4 +1,8 @@
 #![forbid(unsafe_code)]
 pub fn main() {
+    // install default crypto provider (ring)
+    if let Err(e) = rustls::crypto::ring::default_provider().install_default() {
+        eprintln!("Error installing crypto provider: {:?}", e)
```
sebastianlujan, line 5: You could use `std::process:exit(1) ` to exit if the default crypto provider fails.

libtonode-tests/tests/concrete.rs
```diff
@@ -3452,7 +3452,7 @@ mod slow {
             let chainwait: u32 = 6;
             let amount: u64 = u64::from(chainwait * i);
             zingolib::testutils::increase_server_height(&regtest_manager, chainwait).await;
-            let _ = recipient.do_sync(false).await;
+            recipient.do_sync(false).await.unwrap();
```
sebastianlujan, line 3455: Unwap is generally discouraged in production, because could generate panics.

zancas: This is a good point!  And you're correct for production code in `lib` crates.   However we generally permit `unwrap` in test code, so it's ok in the `libtonode-tests` crate.

zingocli/src/main.rs
zancas: This is an interesting option.   What are the trade-offs between these approaches?

zingolib/src/lightclient/sync.rs
```diff
@@ -681,11 +681,12 @@ pub mod test {
         std::env::set_var("RUST_BACKTRACE", "1");
         let wallet = wallet_case.load_example_wallet().await;
         let lc = LightClient::create_from_wallet_async(wallet).await.unwrap();
-        let _ = lc.do_sync(true).await;
+        lc.do_sync(true).await.unwrap();
         lc
     }
 
     /// this is a live sync test. its execution time scales linearly since last updated
+    #[ignore = "testnet and mainnet tests should be ignored due to increasingly large execution times"]
```
zancas, line 689: I am OK with this as a StopGap while we invest in test infrastructure.

zingolib/src/lightclient/sync.rs
```diff
@@ -681,11 +681,12 @@ pub mod test {
         std::env::set_var("RUST_BACKTRACE", "1");
         let wallet = wallet_case.load_example_wallet().await;
         let lc = LightClient::create_from_wallet_async(wallet).await.unwrap();
-        let _ = lc.do_sync(true).await;
+        lc.do_sync(true).await.unwrap();
```
zancas, line 684: May as well make it an:

`expect("If sync panics here, then the test is already interesting.")`

Or perhaps there's a better string.   I don't see why we wouldn't use expect.

zancas:
Alright  I added some non-blocking comments.

zancas (approved):
Given @juanky201271 's test, and caveats about upgrading infrastructure..  I approve.

PR #1428, opened 2024-09-27
Move transparent receiver generation to a separate function

dorianvp:


zingolib/src/wallet/keys/unified.rs
```diff
@@ -331,6 +298,57 @@ impl WalletCapability {
...
     }
 
+    /// Generates a transparent receiver if the wallet is capable of it.
+    ///
+    /// If the wallet is not capable of generating a transparent receiver,
+    /// `None` is returned.
+    ///
+    /// The generated receiver is stored in the wallet if the wallet lock is
+    /// acquired.
+    pub fn generate_transparent_receiver(&self, desired_receivers: ReceiverSelection) -> Result<Option<Secp256k1PublicKey>, String> {
```
Oscar-Pepper: non-blocking comment: I would prefer to use `secp256k1::PublicKey` I don't see the need to alias this as it gives the same information to the reader. it's your call here

Oscar-Pepper:
```suggestion
    pub fn generate_transparent_receiver(&self, desired_receivers: ReceiverSelection) -> Result<Option<secp256k1::PublicKey>, String> {
```

zingolib/src/wallet/keys/unified.rs
```diff
@@ -331,6 +298,57 @@ impl WalletCapability {
         Ok(ua)
     }
 
+    /// Generates a transparent receiver if the wallet is capable of it.
+    ///
+    /// If the wallet is not capable of generating a transparent receiver,
+    /// `None` is returned.
+    ///
+    /// The generated receiver is stored in the wallet if the wallet lock is
+    /// acquired.
```
Oscar-Pepper: good comment. well organised and formatted, explains returns with no mistakes

Oscar-Pepper: wallet lock comment I think may be unneccessry but I have no issue here

zingolib/src/wallet/keys/unified.rs
```diff
@@ -15,7 +15,7 @@ use sapling_crypto::note_encryption::SaplingDomain;
 use zcash_primitives::consensus::{BranchId, NetworkConstants, Parameters};
 
 use crate::config::{ChainType, ZingoConfig};
-use secp256k1::SecretKey;
+use secp256k1::{PublicKey as Secp256k1PublicKey, SecretKey};
```
Oscar-Pepper: if my suggestion below is followed, this alias should be removed. in fact its best to remove this whole line and also call SecretKey as `secp256k1::SecretKey` everywhere it's called.

Oscar-Pepper (dismissed):
I have approved but I have non-blocking suggestions. You can merge as is, or apply the suggestions or reply to my comments and re-send for review.

nachog00 (approved):

PR #1431, opened 2024-10-01
update deps to fix zaino dep conflicts

idky137:
restricted bip0039 dep, moved rustls dep.

This is required to update the zingolib/zingo-testutils dep in zaino-testutils


idky137:
all tests pass locally

PR #1434, opened 2024-10-01
simplify load_example_wallet match

zancas:
This successive match makes the code slightly less repetitive and is a bit easier to read I think since it separates the match of the network from the match of the seed, and handles the seed match inside the network match as a separate *and subsequent* concern.


zancas:
`example_wallet_base` and `example_wallet_address` could be similarly refactored into a cleaner hierarchical separation of concerns.   In fact, this pattern would *extend* the existing pattern for the `pool` match in `example_wallet_address`.

nachog00 (commented):
I notice there's quite a bit of repeated code here. Is there any reason why we wouldn't define a static struct with the needed properties for all of this calls and implement a trait that does the actual matching and calling?

dorianvp (approved):

PR #1435, opened 2024-10-01
update tonic-build version to pass cargo checkmate in CI

zancas:


zancas:
Scooped by dependabot...

PR #1443, opened 2024-10-07
Multistep proposals

AloeareV:


libtonode-tests/tests/concrete.rs
```diff
@@ -159,6 +167,24 @@ mod fast {
                 && vt.recipient_address() == Some(ZENNIES_FOR_ZINGO_REGTEST_ADDRESS)));
     }
 
+    #[tokio::test]
```
AloeareV: Here

zingolib/src/wallet/keys/legacy.rs
```diff
@@ -169,13 +169,12 @@ pub(crate) fn legacy_sks_to_usk(
 pub(crate) fn generate_transparent_address_from_legacy_key(
     external_pubkey: &AccountPubKey,
     address_index: NonHardenedChildIndex,
-) -> Result<TransparentAddress, String> {
+) -> Result<TransparentAddress, bip32::Error> {
```
zancas, line 172: I am cautious around Endianness flips between [B/Z]IP32s.

AloeareV:
I would like to change the test to confirm that the receiver actually receives the funds sent to the tex.

zancas:
Running a local sanity check.

zingolib/src/wallet/transaction_context.rs
```diff
@@ -216,40 +209,69 @@ mod decrypt_transaction {
...
                 .await;
             // Scan transparent spends
             self.account_for_transparent_spending(transaction, status, block_time)
                 .await;
         }
+
+        /// A receipt of funds has been detected at a ZIP320 "ephemeral" return
+        /// address for a Transparent-Source-Only encoded "TEX" address.
+        /// This method records that receipt in therelevant receiving
+        /// TransactionRecord in the TransactionRecordsById database.
```
Oscar-Pepper, line 214: this comment seems to only refer to ephemeral but it is also called when the output_taddr is found in the taddrs_set (external)

zingolib/src/wallet/tx_map/spending_data.rs
```diff
@@ -15,13 +15,16 @@ pub(crate) struct SpendingData {
     witness_trees: WitnessTrees,
     #[getset(get = "pub(crate)", get_mut = "pub(crate)")]
     cached_raw_transactions: HashMap<TxId, Vec<u8>>,
+    #[getset(get = "pub(crate)", get_mut = "pub(crate)")]
+    transparent_ephemeral_ivk: EphemeralIvk,
```
Oscar-Pepper, line 17: can you give some context to what problem are we trying to solve by having keys held anywhere other than UnifiedKeyStore?

zingolib/src/wallet/send.rs
```diff
@@ -143,11 +144,22 @@ impl LightWallet {
...
-        // We don't support zip320 yet. Only one step.
-        if proposal.steps().len() != 1 {
-            return Err(BuildTransactionError::ExchangeAddressesNotSupported);
+        match proposal.steps().len() {
+            1 => {
+                self.create_transaction_helper(sapling_prover, unified_spend_key, proposal)
+                    .await
+            }
+            2 /*TODO: check that this is actually tex*/ => self.create_transaction_helper(sapling_prover, unified_spend_key, proposal).await,
+            _ => panic!("TODO: error handle"),
```
Oscar-Pepper: can we change the BuildTransactionError::ExchangeAddressesNotSupported variant to something else that we put here?

zingolib/src/wallet/keys/unified.rs
```diff
@@ -695,6 +627,15 @@ impl WalletCapability {
             },
         }
     }
+
+    pub(crate) fn ephemeral_ivk(
```
Oscar-Pepper: I think its a mistake to have key related helper functions attached to WalletCapability, I think all key related stuff should stay contained in UnifiedKeyStore. UnifiedKeyStore has a TryFrom for AccountPubKey so I dont see why this helper is neccessary but I think it should at least be on UnifiedKeyStore

zingolib/src/wallet/keys/unified.rs
```diff
@@ -421,31 +400,9 @@ impl WalletCapability {
...
-                Err(e) => {
-                    self.addresses_write_lock
-                        .swap(false, atomic::Ordering::Release);
-                    return Err(e.to_string());
-                }
-            }
-        } else {
-            None
-        };
+        let transparent_receiver = self
```
Oscar-Pepper: why have we removed the desired_receivers check and also the fvk generation so its unified with the logic of sapling and orchard? I did this work on my last PR and its much cleaner and important to prevent bugs with UFVKs without transparent viewing keys

Oscar-Pepper (changes requested):

zingolib/src/wallet/tx_map/spending_data.rs
AloeareV: We don't want to our WalletRead/WalletWrite implementers to need spending keys. One reason is because we want a hardware wallet to be able to call the transaction creation step, without zingo holding the keys, and then passing the transaction to the hardware wallet to sign.

There are a couple places where we need viewing keys for walletread/walletwrite, so we copy those keys into the TxMap.

zingolib/src/wallet/transaction_context.rs
zancas: Ready for resolution.

zingolib/src/wallet/keys/unified.rs
AloeareV: Oh lord...I think this may have been lost when I was trying to fix merge conflicts.

zingolib/src/wallet/send.rs
zancas: Ready for resolution.

zingolib/src/wallet/keys/unified.rs
zancas: Resolved by preceding comment.

AloeareV: This function goes WalletCapability->AccountPubKey->EphemeralIvk. I was calling `AccountPubKey::try_from(wallet_capability().unified_key_store()).ephemeral_ivk()` in enough places I wanted to make a helper function for it.

Turning it into a unified_key_store method means that in all the places it's used, we'd need to add `.unified_key_store()` first...which isn't a very large cost, but this function exists to combine 3-ish lines of code to 1, and going down to 2 lines of code defeats half the point. 

What makes a helper 'necessary'? If you're worried about cluttering the `unified_key_store` impls, would moving it to be a free function instead of a method solve that?

Oscar-Pepper: ok thats fine. I think i am anticipating that the keystore might move eventually so its best to keep it contained but we can cross that bridge later.

zingolib/src/wallet/tx_map/spending_data.rs
Oscar-Pepper: this leads me to believe that walletwrite/ walletread is implemented on the wrong struct, which was my suspicion in the past. we can fix this later!

zingolib/src/wallet/keys/unified.rs
zancas: Is this resolved to your satisfaction @Oscar-Pepper ?

AloeareV: Yes, I think on one version it was moved, and when I resolved conflicts it was now in neither location...added back, good catch.

zancas: @AloeareV has re-added it.

PR #1445, opened 2024-10-08
implement read/write for ephemeral addresses

Oscar-Pepper:


zancas:
Endianess is not consistent between BIP32 and ZIP32:

https://zips.z.cash/zip-0032#conventions

Why are we *now* specifying a BigEndian order for read/write?    Given that ZIP32 specifies LittleEndian throughout, are we compliant?

pacu (commented):
utACK (untested but acknowledged)

zingolib/src/wallet/keys/unified.rs
```diff
@@ -11,6 +11,7 @@ use std::{marker::PhantomData, sync::Arc};
 use append_only_vec::AppendOnlyVec;
 use bip0039::Mnemonic;
 use byteorder::{ReadBytesExt, WriteBytesExt};
+use bytes::BigEndian;
```
zancas: Why are using BigEndian?  ZIP32 specifies LittleEndian.

zancas (changes requested):

zingolib/src/wallet/keys/unified.rs
Oscar-Pepper: AFAIK bip32 is bigendian and more suitable for multiplatform support as oppose to performance for mobile OSs. The data being stored in bigendian is only a u32 relating to a number of transparent addresses so im not sure why this is related to ZIP32? I dont claim to be well-versed on bigendian vs. littleendian but this is why I made this decision. I will wait for your response and am happy to change to littleendian if you feel its more appropriate

zancas: BIP32 *is* BigEndian but ZIP32 is LittleEndian.

Oscar-Pepper: yes i said that in my comment.... my point is if bip32 or zip32 was related to this (which it's not) it would be closer to bip32 (BigEndian) because this is to do with transparent addresses not shielded. However, this is not related to key derivation, it's just an extra piece of data to do with number of addresses we are adding to the wallet, so it's our call on which endianness, and from my research I decided BigEndian made more sense. If you disagree i will change it so we can land this fix

Oscar-Pepper: i have looked an `read_internal` and it seems we prefer little endian for most things so i will change this now!

zancas (approved):

zancas (approved):

PR #1451, opened 2024-10-13
fix zingocli viewkey

Oscar-Pepper:
this fixes an error when creating a wallet from UFVK in zingocli due to the viewing key not having 24 words (it was passed through seed phrase checks)


pacu (approved):
LGTM!

zancas (approved):
UtAck

PR #1453, opened 2024-10-14
Add invalid value for parse_address command

dorianvp:
This PR adds a return value to indicate when an invalid address is encountered in the `parse_address` command.


juanky201271 (approved):
looks good to me... maybe @AloeareV can take a look as well...

PR #1464, opened 2024-10-21
Add Spacers and Points in zingocli balance command

fluidvanadium:
Adding a decimal point allows easy estimation of zec per zatoshi.

### DIFF
--- a/zingolib/src/lightclient.rs
+++ b/zingolib/src/lightclient.rs
@@ -131,17 +131,19 @@ pub struct PoolBalances {
 fn format_option_zatoshis(ioz: &Option<u64>) -> String {
     ioz.map(|ioz_num| {
         if ioz_num == 0 {
-            return "0".to_string();
+            "0".to_string()
         } else {
             let ioz_string = format!("{:0>8}", ioz_num);
             let digits = ioz_string.as_bytes();
-            digits
+            let mut ioz_string = digits
                 .rchunks(3)
                 .rev()
                 .map(std::str::from_utf8)
                 .collect::<Result<Vec<&str>, _>>()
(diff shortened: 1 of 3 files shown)


zancas:
Is this moving us towards reproducing the vaporizing transaction bug?

zingolib/src/lightclient.rs
```diff
@@ -128,6 +128,63 @@ pub struct PoolBalances {
     /// TODO: Add Doc Comment Here!
     pub transparent_balance: Option<u64>,
 }
+fn format_option_zatoshis(ioz: &Option<u64>) -> String {
```
zancas, line 131: I don't know what "`ioz`" means.

zancas:
It looks like this crate has been downloaded aroun 15_000_000 times:

https://crates.io/crates/num-format

Why don't we us it?

zancas (changes requested):
Let's use num-format.

zancas:
I really like the idea of a more useful/readable zingocli interface.   I don't like the idea of writing our own reimplementation of standard logic.   Let's use the thing that's "off-the-shelf".

fluidvanadium:
> It looks like this crate has been downloaded aroun 15_000_000 times:
> 
> https://crates.io/crates/num-format
> 
> Why don't we us it?

the answer is: num-format doesnt specifically make use of the 10,000,000 mark for Satoshis

zingolib/src/lightclient.rs
zancas: Is it like "input_of_zatoshis"?    Could we name this parameter:  "zats"?

zancas (approved):

PR #1474, opened 2024-10-28
Explicit Invocation of thiserror::Error macro

fluidvanadium:
- increases explicitness: Error is a crate, not to be confused with a standard derive trait, like Debug.
    - increased locality: rather than having a use statement in an arbitrary place, this groups error-related code together
    - standardization


zancas (approved):
Nice hack.  I like this a lot.

fluidvanadium:
FYI somebody else did it first, but i didnt bother to blame them

PR #1475, opened 2024-10-28
Handle start mempool monitor

fluidvanadium:
This is a best-practice improvement that helps  development


zingolib/src/lightclient/sync.rs
```diff
@@ -1,4 +1,4 @@
-//! TODO: Add Mod Description Here!
+//! TODO: Add Mod Description HeStartMempoolMonitorError::Disabled!
```
zancas: This looks like a search-and-replace error.

zancas (changes requested):
Please fix the mysterious doc-comment.

zancas (approved):

PR #1482, opened 2024-10-30
Lrz oct 28 - Upgrade shardtree

fluidvanadium:
Trying to keep the LRZ dependency close to updated here.


fluidvanadium:
requesting review from @AloeareV because of her insight as to shardtree truncation heights

AloeareV (approved):

PR #1496, opened 2024-11-04
For host x86_64 do not use arm64 android AVD in ci

juanky201271:
logs:
```
INFO    | Found systemPath /usr/local/lib/android/sdk/system-images/android-29/default/arm64-v8a/
PANIC: Avd's CPU Architecture 'arm64' is not supported by the QEMU2 emulator on x86_64 host.
```

### DIFF
--- a/.github/workflows/ci-nightly.yaml
+++ b/.github/workflows/ci-nightly.yaml
@@ -42,7 +42,7 @@ jobs:
   android-ubuntu-integration-test-actions:
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


fluidvanadium (approved):
any particular reason?

juanky201271:
> any particular reason?

The reason is: the nightly-ci flow is going to fail all the time for the arm64 architectures... and for me, this everyday CI flow is interesting because we can check every night if zingolib is building in zingo-mobile correctly...

PR #1497, opened 2024-11-04
retry test to handle flakes

fluidvanadium:
this should save on developer time spent hitting the github retry button. should leave a record in place of how many tries it took


fluidvanadium:
@zancas do you have a strong opinion about this?

AloeareV (approved):

PR #1505, opened 2024-11-07
fix shield when wallet contains more than one transparent address

AloeareV:
There was a bug preventing propose_shield from working when there was more than one transparent address in the wallet.


fluidvanadium (approved):
tested! success!

PR #1510, opened 2024-11-13
try not to shield for nothing

fluidvanadium:
This fixes an issue in which 0-value shields might not be caught instantly.


AloeareV (dismissed):
Great!

AloeareV (approved):

PR #1513, opened 2024-11-14
[DRAFT] Implement Ledger Support 

pacu:
@Oscar-Pepper and I debated a bit on how to integrate ledger under the new UnifiedKeyStore architecture.

This PR will be implementing those changes gradually. There are a lot of build problems that are blocking the full integration and need to be solved but all the structural code here will work around that 

Kickstarting ideas from Pepper

> so i had a look through the the zondax PR and my opinion is that we shouldnt be duplicating the WalletCapability, it has many methods that just re-implements code thats already in the codebase which is harder to maintain. a lot of the code around derivation paths and address generation is standardised so I think its best we try to use current zingolib code as much as possible and integrate the ledger instead of tacking it on. My approach would be to have a feature gated struct in WalletCapability that holds any novel ledger information such as the ZcashApp and LedgerId. the current wallet capability can be used to hold the addresses and keys etc.

> I think we can have the WalletBase::Ledger be similar to WalletBase::Ufvk except it connects to the ledger to retrieve the keys instead of a user inputted encoded UFVK string on startup. we can also feature gate the write impl for keys to stop the UFVK being written to file, so its loaded to memory on startup only

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,6 @@
 {
     "rust-analyzer.diagnostics.disabled": [
         "unresolved-macro-call"
-    ]
+    ],
+    "rust-analyzer.cargo.features": "all"
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,7 +33,7 @@ version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
- "cfg-if 1.0.0",
+ "cfg-if",
  "cipher",
  "cpufeatures",
 ]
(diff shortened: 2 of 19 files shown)


pacu:
@Oscar-Pepper here's the point where the single WalletCapability falls short.

I think I can apply a refactor where we extract the inner workings of WalletCapability into an internal trait that implements a visitor pattern that calls the corresponding variant without exposing that there is a variant WalletCapability

zancas:
How is this related to #1352 ?   Does this supersede that PR?  Should that PR still be open?

pacu:
> How is this related to #1352 ? Does this supersede that PR? Should that PR still be open?

That one unfortunately has been made obsolete by the new UnifiedKeystore work. I'm using it as reference only.

pacu:
@zancas @Oscar-Pepper I rebased and updated this PR

zancas:
Hi @pacu should we update this PR?

pacu:
@zancas this work can't be continued until we don't have a proper Zcash Shielded App API that works with current key material.  I will close this

PR #1521, opened 2024-11-15
Remove least semver numbers

zancas:
Removes least semver numbers below x: "0.x.n" and "1.x"

Builds on previous updates.


zancas:
I would like @idky137 's opinion on this.   This proposes a permissive dependency policy, that still forbids dependency changes that would break public APIs.   I think it's the most permissive policy that has this constraint.

zancas:
I ran a full sanity test locally.

dorianvp (dismissed):

PR #1526, opened 2024-11-16
Test of pending load

zancas:
I think this test is showing that pending transactions are *not* properly recorded in the wallet.  Maybe the test doesn't save the new wallet.dat?


zancas:
Superceded by new chain_generic test.

PR #1537, opened 2024-11-19
don't touch send flow this time

AloeareV:
Instead of messing up send flow when this rescan doesn't update obsolete metadata, only add the check to the rescan flow and not the send flow.

### DIFF
--- a/zingolib/src/wallet/transaction_context.rs
+++ b/zingolib/src/wallet/transaction_context.rs
@@ -40,16 +40,22 @@ impl TransactionContext {
         &self,
         wallet_height: BlockHeight,
     ) -> Result<(), Vec<(TxId, BlockHeight)>> {
-        self.transaction_metadata_set
+        let tmdsrl = &self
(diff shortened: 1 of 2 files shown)


zancas:
I see that this constrains extension of output indexes to only be from confirmed transactions.   I understand this to exclude data from "the send flow" is that how to interpret the title of this commit?

zancas (approved):

AloeareV:
> I see that this constrains extension of output indexes to only be from confirmed transactions. I understand this to exclude data from "the send flow" is that how to interpret the title of this commit?

No, the original attempt at this added the new code inside `get_spendable_note_ids_and_values`, which was called during the send flow as well as inside `unindexed_records`. The reason we only check confirmed notes, is that this code is only called when we read old transactions from the wallet, that are missing information. Unconfimed notes don't need to be rescanned (they're too new to be able to hit this clause, and even if they could they're also gonna be rescanned on confirmation anyway.)

PR #1540, opened 2024-11-19
Offline Mode for zingo-mobile

dorianvp:


zancas:
@dorianvp can this be tested with Kotlin in zingo-mobile?  Or detox?

zingolib/src/config.rs
```diff
@@ -98,6 +98,45 @@ pub fn load_clientconfig(
...
+pub fn load_clientconfig_serverless(
+    lightwallet_uri: http::Uri,
+    data_dir: Option<PathBuf>,
+    chain: ChainType,
+    monitor_mempool: bool,
+) -> std::io::Result<ZingoConfig> {
+    use std::net::ToSocketAddrs;
+    match format!(
+        "{}:{}",
+        lightwallet_uri.host().unwrap(),
```
zancas, line 76: Who's passing this uri to this function?   Should we add a error enum to handle the failures that can occur inside this fn?

Probably using all the functionality that `thiserror` offers to be ergonomic in the definition?

zancas: Given that the parameter is a Uri, how can host() and port() fail?

Can we write unit-tests that force the exercise of these failure modes?

dorianvp: The ideal thing would be to alter the `construct_lightwalletd_uri` method to return a `Result<http::Uri, http::uri::InvalidUri>`. We could use `panic::catch_unwind` meanwhile in tests, and then have another PR with the refactor in place.

zancas: I agree let's change the return of `construct_lightwalletd_uri` to the `Result` suggested.

zancas: We can do this later.

zancas (approved):

zingolib/src/config.rs
```diff
@@ -726,3 +730,42 @@ impl ActivationHeights {
...
+    #[tokio::test]
+    async fn test_load_clientconfig_serverless() {
+        rustls::crypto::ring::default_provider()
+            .install_default()
+            .expect("Ring to work as a default");
+        tracing_subscriber::fmt().init();
+
+        let valid_uri = crate::config::construct_lightwalletd_uri(Some(
+            crate::config::DEFAULT_LIGHTWALLETD_SERVER.to_string(),
+        ));
```
juanky201271, line 745: I don't really think this test is correct, I see `valid_uri` with a valid uri, hehehehe, obviously, and IMO this test needs a different value here, IDK like `http::Uri::default()` which for me is a `empty` server uri.

zingolib/src/config.rs
```diff
@@ -726,3 +730,42 @@ impl ActivationHeights {
...
+
+        let valid_config = crate::config::load_clientconfig(
+            valid_uri.clone(),
+            Some(temp_path),
+            crate::config::ChainType::Mainnet,
+            true,
+        )
+        .unwrap();
+
+        assert_eq!(valid_config.get_lightwalletd_uri(), valid_uri);
```
juanky201271, line 760: this make no sense to me... in a serverless scenario... in fact the normal situation for `serverless` is:
- server uri: empty.
- chain type: empty.

PR #1541, opened 2024-11-20
propose distinct 40 block ZIP410 eviction threshold

zancas:
I suspect that these constants should be almost completely decoupled.   To promote this into a review-ready PR we need to add a test that covers the logical change, and makes appropriate assertions.


fluidvanadium:
oh yike! this is a legacy function that probably does nothing ATM!

zancas:
> oh yike! this is a legacy function that probably does nothing ATM!

https://github.com/zingolabs/zingolib/blob/7bc2d8b79230e650c7241ae736fc44a4a271f017/zingolib/src/lightclient/sync.rs#L661

PR #1542, opened 2024-11-20
Filter empty messages in messages_containing function

dorianvp:


fluidvanadium (approved):
upgraded test confirms

PR #1551, opened 2024-11-27
Zingo sync transparent with get taddress txids

Oscar-Pepper:
following PR will send the locators in the scan tasks to fetch and scan transparent bundles of relevant txs.
there is a TODO for implementing memo encoded address ids also


zancas:
CI fail

PR #1554, opened 2024-11-28
Parallelize shardtree

AloeareV:
This is in _dire_ need of some renames, but this ideally makes the batch trait generic enough to implement a `ShardTreeUpdateBatch`. I plan to add that in this PR, I'm opening it now mostly to get help with the naming, and in case I've done something horribly wrong that @Oscar-Pepper can save me from!


zingo-sync/src/scan/compact_blocks/runners.rs
```diff
@@ -96,42 +104,42 @@ where
         for tx in block.vtx.into_iter() {
```
Oscar-Pepper, line 134: sorry that this might get more complicated but this add_block method seems specific to trial decryption. in ShardTreeData which is returned in ScanResults of a scan, we have the initial leaf position and a vec of Nodes and Retentions. I think we want to massage the way that batchrunners add widgets to allow to add all outputs from a compact block for trial decryption but for shard tree batchs we want to add an initial position for the Batch and a bunch of Nodes and Retentions. maybe the simple way to do this would be to move the pre-processing to outside of the batchrunner completely so for example in the trial decrpytion case we gather all the outputs (batch inputs) and feed directly to batch runners? this way the batch runners doesnt need additional traits for pre-processing batch inputs

Oscar-Pepper: i think its worth also looking into the best format for inserting into ShardStore in a parallel manner, as initial positon, Nodes and Retention is specific for `batch_insert` and it might be better to include the position with each node/retention for example

zingo-sync/src/scan/compact_blocks/runners.rs
```diff
@@ -306,47 +290,108 @@ pub(crate) struct Batch<D: BatchDomain, Output, Dec: Decryptor<D, Output>> {
...
+    /// `replier` will be called with the result of every output.
+    fn add_widgets(
+        &mut self,
+        widgets: impl ExactSizeIterator<Item = Self::Input>,
+        replier: channel::Sender<(usize, Self::ResultVal)>,
+    ) {
+        let widget_len = widgets.len();
+        self.inputs_mut().extend(widgets);
+        self.repliers_mut()
+            .extend((0..widget_len).map(|output_index| (output_index, replier.clone())));
```
Oscar-Pepper: i am making an assumption here but i imagine there is a distinct difference between trial decrpytion and shardtree where trial decrypting and returning a result for each individual output is neccessary but for shardtree we would probably have a result for the insertion of the whole batch? would we need a replier for each node? it likely this is just easier for now, maybe an optimisation we can work on later

zingo-sync/src/scan/compact_blocks/runners.rs
```diff
@@ -453,28 +480,29 @@ impl DynamicUsage for ResultKey {
 }
 
 /// Logic to run batches of trial decryptions on the global threadpool.
-pub(crate) struct BatchRunner<D, Output, Dec, T>
+pub(crate) struct BatchRunner<B, D, T>
 where
-    D: BatchDomain,
-    Dec: Decryptor<D, Output>,
-    T: Tasks<Batch<D, Output, Dec>>,
+    D: BatchDomain<IncomingViewingKey: Send, Recipient: Send, Memo: Send>,
```
Oscar-Pepper, line 441: IVK are not related to shardstore update but im happy to untangle this after we have a PoC

zingo-sync/src/scan/compact_blocks/runners.rs
AloeareV: > i think its worth also looking into the best format for inserting into ShardStore in a parallel manner, as initial positon, Nodes and Retention is specific for `batch_insert` and it might be better to include the position with each node/retention for example

The best method of parallelizing I believe is to create `LocalizedPrunableTree`s, which can be created in parallel (each one taking an initial position (well, a range, but the upper bound is unnecessary), as well as Nodes and Retentions), and then acquiring the shardtree lock at the end in order to insert the localizedprunabletrees. The trees will have been constructed and maximally pruned in parallel, and then insertion is still unparallelizable, but after pruning there will be probably at least an order of magnitude fewer nodes to insert
.

AloeareV: > sorry that this might get more complicated but this add_block method seems specific to trial decryption. in ShardTreeData which is returned in ScanResults of a scan, we have the initial leaf position and a vec of Nodes and Retentions. I think we want to massage the way that batchrunners add widgets to allow to add all outputs from a compact block for trial decryption but for shard tree batchs we want to add an initial position for the Batch and a bunch of Nodes and Retentions. maybe the simple way to do this would be to move the pre-processing to outside of the batchrunner completely so for example in the trial decrpytion case we gather all the outputs (batch inputs) and feed directly to batch runners? this way the batch runners doesnt need additional traits for pre-processing batch inputs

Well, we could do a similar thing where we add the batch inputs as we acquire them. It might be simpler in the concrete case to build up a full batch's worth of nodes instead of having the add_widgets(name pending) method, but I think using the same interface for both will be simpler than trying to make the interfaces diverge further.

zingo-sync/src/scan/compact_blocks/runners.rs
AloeareV: Agreed! The whole batch should resolve to a single `LocatedPrunableTree`, the exact details will depend on how the implementation of the `ShardTreeUpdateBatch` shakes out

zingo-sync/src/scan/compact_blocks/runners.rs
AloeareV: Yeah, the shardstore stuff will still be generic for a domain, so these are unnecessary constraints for the shardtree case, but as the domains remain the same two concrete types for both batch kinds, these constraints shouldn't actually hurt anything. It would be nice to untangle them nonetheless!

PR #1558, opened 2024-11-29
Zingo sync transparent bundle scanning

Oscar-Pepper:


zingo-sync/src/primitives.rs
```diff
@@ -92,6 +115,30 @@ impl Default for NullifierMap {
     }
 }
 
+/// Binary tree map of out points (transparent spends)
+#[derive(Debug)]
+pub struct OutPointMap(BTreeMap<OutputId, Locator>);
```
AloeareV, line 120: Why is this a new type? This feels like an unnecessary layer of indirection. Non-blocking

AloeareV (approved):

PR #1561, opened 2024-11-29
Fix off-by-2! Darkside Reloaded

fluidvanadium:
So far:
-darkside bump_chain works with multiple transactions
-darkside generic works with more fixtures
-FIXED off-by-2 production error!

Future:
-feature-gate the accept server txids block.
-Enough features in darkside work that the next eviction test will be writable


fluidvanadium:
superceded by 1577

PR #1564, opened 2024-11-30
Make value transfers more idiomatic

zancas:
This makes `ValueTransfers` implement trait to support an idiomatic interface.

Apart from readability in general it's possible poor ergonomics on this type confused us into writing a bad test:

`fast::tex::send_to_tex`


zancas:
Confirm CLT passes.

dorianvp (approved):

PR #1565, opened 2024-11-30
Investigate send to tex

zancas:
Builds on #1564 which makes value_transfers easier to use.

Why does the first value transfer of sender (which should be a receipt) have a `recipient_address`?

Should it?   Has this already been fixed on @fluidvanadium 's feature branch?


zancas:
OK, mystery solved, the value transfers are sorted by newer first.  No obvious bug.

PR #1575, opened 2024-12-03
Zingo sync mempool

Oscar-Pepper:


zancas:
These changes are incorporated in #1579 .

fluidvanadium:
Why is there not OutgoingTransparentCoin?

Oscar-Pepper:
> Why is there not OutgoingTransparentCoin?

OutgoingSaplingNote and OutgoingOrchardNote come from decrypting the outgoing notes in the transaction but there is no such transparent equivalent.

Oscar-Pepper:
> These changes are incorporated in #1579 .

this PR should be reviewed and merged separately to the removal of getset to keep this PR concise. also there is still ongoing discussion about getset removal

libtonode-tests/tests/sync.rs
```diff
@@ -75,10 +75,10 @@ async fn sync_test() {
     // .await
     // .unwrap();
 
-    increase_server_height(&regtest_manager, 1).await;
-    recipient.do_sync(false).await.unwrap();
-    recipient.quick_shield().await.unwrap();
-    increase_server_height(&regtest_manager, 1).await;
+    // increase_server_height(&regtest_manager, 1).await;
```
AloeareV, line 78: Why is the shield test getting commented out? Is this restored by the verification fix followup pr?

zingo-sync/src/primitives.rs
```diff
@@ -284,7 +290,7 @@ pub struct WalletNote<N, Nf: Copy> {
     nullifier: Option<Nf>, //TODO: syncing without nullfiier deriving key
     /// Commitment tree leaf position
```
AloeareV, line 291: Position is none for mempool and some for confirmed, is this the only source of truth for whether a walletnote is confirmed? If so this needs to be clearer, a comment will probably do the trick,,
```suggestion
    /// Commitment tree leaf position if confirmed, or None for an unconfirmed note
```

zingo-sync/src/sync.rs
```diff
@@ -202,6 +228,102 @@ where
...
+            if mempool_transaction.transparent_coins().is_empty()
+                && mempool_transaction.sapling_notes().is_empty()
+                && mempool_transaction.orchard_notes().is_empty()
+                && mempool_transaction.outgoing_orchard_notes().is_empty()
+                && mempool_transaction.outgoing_sapling_notes().is_empty()
+            {
+                return;
+            }
+
+            // TODO: there is an edge case where txs sent to the mempool that do not return change or create outgoing notes
```
AloeareV, line 298: About this TODO: I think the second step of a send-to-tex doesn't return change, and sends only to transparent. Non-blocking, but worth thinking about.

AloeareV (approved):

libtonode-tests/tests/sync.rs
Oscar-Pepper: this test is just a template test i use to verify things as im developping the PoC. its not a fixed test that will stay in the codebase. integration tests are on the todo lists after PoC is complete

zingo-sync/src/primitives.rs
Oscar-Pepper: well... the single source of truth for whether a note is confirmed is whether the transaction it belongs to is confirmed. I imagine the position will be None for notes in confirmed transactions when we add support for viewing key only wallets for extra speed up right?

zingo-sync/src/sync.rs
Oscar-Pepper: yes I thought of this and i am already working on this edge case in a following PR. im glad you are thinking deeply about the code in various scenarios though!

PR #1576, opened 2024-12-04
Simpler parallelize shardtree

AloeareV:
The runners were definitely overkill! This is so much simpler.


Oscar-Pepper (approved):
wonderful work!

PR #1577, opened 2024-12-05
Darkside reloaded 3

fluidvanadium:
Test upgrades:
  - feature-gated txid replacement logic
  - darkside bump chain now works with multiple transactions
  - with_assertions has more checks
  - specific values replaced with single_sufficient_send proptest
  - DRY and refactors


darkside-tests/src/chain_generics.rs
```diff
@@ -82,41 +103,105 @@ pub(crate) mod conduct_chain {
...
                             .unwrap();
-                        self.tree_state = update_tree_states_for_transaction(
-                            &self.darkside_connector.0,
-                            raw_tx.clone(),
-                            u64::from(self.staged_blockheight),
+
+                        //trees
+                        let transaction = zcash_primitives::transaction::Transaction::read(
+                            raw_tx.data.as_slice(),
+                            zcash_primitives::consensus::BranchId::Nu5,
```
AloeareV: Hardcoded branch ID of nu5 concerns me, as we just hit nu6! I intend to turn this into a more meaningful comment before I submit the review, once I understand how this effects things.

darkside-tests/src/darkside_connector.rs
```diff
@@ -51,7 +47,7 @@ impl DarksideConnector {
                     let uri = Uri::builder()
                         .scheme(uri.scheme().unwrap().clone())
                         .authority(uri.authority().unwrap().clone())
-                        //here. The Request's uri contains the path to the GRPC server and
+                        //here. The Request's uri contains the path to the GRPC sever and
```
AloeareV: Revert typo!
```suggestion
                        //here. The Request's uri contains the path to the GRPC server and
```

libtonode-tests/tests/chain_generics.rs
AloeareV, line 1: I need to think through this set of proptest tests, to confirm that all those removed tests don't remove any cases not covered by the new proptest invocation, this comment is therefore a breadcrumb for myself! This is so much more succinct, big fan.

AloeareV (commented):
This is great! I've left myself a couple comments of things I want to understand better before either approving them or requesting changes.

libtonode-tests/tests/chain_generics.rs
AloeareV: > I need to think through this set of proptest tests, to confirm that all those removed tests don't remove any cases not covered by the new proptest invocation, this comment is therefore a breadcrumb for myself! This is so much more succinct, big fan.

I have confirmed to my satisfaction that these two tests do the job of the roughly 35(!!) tests they replace!

darkside-tests/src/chain_generics.rs
fluidvanadium: i think this wont effect the transaction, but it should be fixed
it would be good to make this a Const

AloeareV: Looks like this arg is only read in the pre-zip255 case!

AloeareV (approved):

PR #1579, opened 2024-12-06
Zingo sync mempool getsetless

zancas:
Builds on #1575 and #1573 which outlines the argument for a getsetless approach.

#1580 builds on this but controversially removes probably future-proofing work. (In an easily reapplied commit.)

Probably the right pattern is to keep dead code on feature branches until it's ready to LLIiiiVE!


fluidvanadium:
In my experience, during an involved feature is not the best time to refactor.

Issue #1557, opened 2024-11-29
Tasks remaining -> Offline mode (serverless)


juanky201271:
We need to use this command to change to `no server` -> with the parameter empty, something like:

```
commands::do_user_command("changeserver", "", lightclient.as_ref())
```

We need to start zingo-cli with `no server` -> with this parameter empty. And we need all the commands like `balance` or `valuetransfers` working with the information stored in the file `wallet.dat`.

We need this: `LightClient::read_wallet_from_buffer_runtime` working, using an instance of `config` that was created with `no server` -> `http::Uri::default()`.

juanky201271:
would be nice if the user can start zingo-cli with `no server`. This is the Error:
![image](https://github.com/user-attachments/assets/f337d595-8aba-4994-a477-70b352c920d9)

juanky201271:
Thanks @dorianvp , closing...

PR #1582, opened 2024-12-06
Add support for empty server in `changeserver` command

dorianvp:
Ref: #1557


zancas (commented):
LGTM   @juanky201271 what do you think?

juanky201271 (approved):
As you know, I'm not the best to review this, but for me looks good.

PR #1583, opened 2024-12-06
Handle case where args len is empty in `changeserver` command

dorianvp:


juanky201271 (approved):
nice.

dorianvp:
The CLT suite passed.

dorianvp:
The CLT suite passed.

dorianvp:
The CLT suite passed.

PR #1586, opened 2024-12-11
fix set_verify_scan_range

Oscar-Pepper:


AloeareV (approved):
Missed this one, good catch!

PR #1588, opened 2024-12-14
Fix minor bugs

Oscar-Pepper:
Sync should be much more finely tuned with this PR

fixes:
1. to fix tokio polling being CPU-starved. moves located tree construction to scanning and uses spawn_blocking to keep rayon tasks in a separate thread pool to the main tokio thread
2. fixes a bug where workers may be shutdown before scan is complete
3. moves the point where a worker is a set to be scanning to fix a bug where workers are assigned multiple tasks
4. fixes a mempool panic, more work to be done on mempool in following PR

### DIFF
--- a/zingo-sync/src/scan.rs
+++ b/zingo-sync/src/scan.rs
@@ -173,7 +173,19 @@ where
     .await
     .unwrap();
 
-    let scan_data = scan_compact_blocks(compact_blocks, parameters, ufvks, initial_scan_data)?;
+    let consensus_parameters_clone = parameters.clone();
+    let ufvks_clone = ufvks.clone();
+    let scan_data = tokio::task::spawn_blocking(move || {
+        scan_compact_blocks(
+            compact_blocks,
+            &consensus_parameters_clone,
+            &ufvks_clone,
+            initial_scan_data,
+        )
+    })
+    .await
+    .unwrap()?;
+    // let scan_data = scan_compact_blocks(compact_blocks, parameters, ufvks, initial_scan_data)?;
 
     let ScanData {
         nullifiers,
--- a/zingo-sync/src/scan.rs
+++ b/zingo-sync/src/scan.rs
@@ -40,7 +40,7 @@ struct InitialScanData {
 impl InitialScanData {
     async fn new<P>(
         fetch_request_sender: mpsc::UnboundedSender<FetchRequest>,
-        parameters: &P,
+        consensus_parameters: &P,
         first_block: &CompactBlock,
         previous_wallet_block: Option<WalletBlock>,
     ) -> Result<Self, ()>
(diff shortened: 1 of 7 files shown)


zancas:
I'm confused..  is there a test?   How can we know whether the sync is tuned, without observations?

Oscar-Pepper:
> I'm confused.. is there a test? How can we know whether the sync is tuned, without observations?

we dont have the benchmarking framework yet to show the performance improvements in the form fo a test.
i will edit the PR description to include more information about the changes / fixes

zancas:
Are these bugs really "minor"?  It seems like at least some are **critical** even if the effort to fix them is low.  I am concerned that this code should be reviewed carefully, and "minor" might give the PR the wrong "flavor".

zingo-sync/src/scan.rs
```diff
@@ -201,11 +214,32 @@ where
...
     .unwrap();
 
+    let WitnessData {
+        sapling_initial_position,
+        orchard_initial_position,
+        sapling_leaves_and_retentions,
+        orchard_leaves_and_retentions,
+    } = witness_data;
+
+    let sapling_located_trees = tokio::task::spawn_blocking(move || {
```
AloeareV, line 224: Is there any reason not to join these two futures instead of awaiting them individually, so that both tasks can spawn and run concurrently on different threads? EDIT: I thought this was how I originally did it, but it turns out that's how I originally did subtree_root addition, which is unpushed.

zingo-sync/src/witness.rs
```diff
@@ -41,21 +43,76 @@ impl Default for ShardTrees {
...
             sapling_initial_position,
             orchard_initial_position,
             sapling_leaves_and_retentions: Vec::new(),
             orchard_leaves_and_retentions: Vec::new(),
         }
     }
 }
+
+/// Located prunable tree data built from nodes and retentions during scanning for insertion into the shard store.
+pub struct LocatedTreeData<H> {
```
AloeareV, line 66: Why use this new type? BatchInsertionResult contains these two fields, plus extras that may-or-may-not be useful later...but adding this transformation that only strips data from the result before returning it feels like unneeded ceremony.

AloeareV (approved):
I've requested 2 changes (well, one of them isn't phrased like a request), neither are blocking! Feel free to apply them here, or merge and either add them in later or argue against them!

zingo-sync/src/scan.rs
Oscar-Pepper: as each of these tasks create concurrent synchronous tasks across multiple threads I didn't think this would improve performance but now you say it there may be some downtime where sapling_trees is almost complete and the threads can be utilised for orchard_trees so will make this change in a future PR. good find!

zingo-sync/src/witness.rs
Oscar-Pepper: I actually tried this approach but I had issues with the fact that BatchInsertionResult's remainder field is an Iterator. I spent quite a bit of time trying to get it to work but settled for creating a new struct with all the currently relevant data.

PR #1593, opened 2024-12-17
get subtree roots to allow for spend-before-sync

AloeareV:


AloeareV:
Ohh.... [5bd083a](https://github.com/zingolabs/zingolib/pull/1593/commits/5bd083abfaafcbe5212e144824e3c593599bbf1d) breaks load_old_wallet_at_reorged_height, as adding that flag to the zcash config is incompatible with the old chain being loaded that didn't have that flag. Either that test needs special handling to not use the new flag, or we need to create a new chain for that test.

Oscar-Pepper:
> Ohh.... [5bd083a](https://github.com/zingolabs/zingolib/pull/1593/commits/5bd083abfaafcbe5212e144824e3c593599bbf1d) breaks load_old_wallet_at_reorged_height, as adding that flag to the zcash config is incompatible with the old chain being loaded that didn't have that flag. Either that test needs special handling to not use the new flag, or we need to create a new chain for that test.

ok i will take a look at fixing this. having zcash-local-net integrated would make this much easier!

Oscar-Pepper (approved):
splendid!

PR #1596, opened 2024-12-19
add file of tests extracted from zcash_local_net

zancas:
I've [removed `zinglib` as a dependency of `zcash-local-net`](https://github.com/zingolabs/zcash-local-net/pull/35) and made `libtonode-tests` depend on `zcash-local-net`.

I've migrated tests that required `zingolib` from `zcash-local-net` into a file in `libtonode-tests`.  That file starts with:

```Rust
#![cfg(feature = "zln_tests")] 
```

`zln_tests` includes `portpicker` and `tonic` which are optional and therefore not built when this feature is not selected.

The upshot is that the 304 integrations tests that were running before this PR remain the same and will trigger with `cargo nextest run`.  The additional tests from zln run with the feature and mostly fail even if binaries are reachable from the expected env variables, due to the sufficient shards requirement for most of these tests.

Things to consider before merging:

   * zaino-testutils depends on zcash-local-net and I haven't verified correct function against this feature
   * the `clent` and `test_fixtures` mods were copied over without consideration for whether the were repeating more ergonomic functionality available in zingolib
   * the LocalNet fixture isn't yet leveraged to upgrade the integration tests

Obviously this PR is complimentary to the PR in `zcash-local-net`.

Probably @Oscar-Pepper and @idky137 should have a look at it before we merge it (if we ever do).

It *does* have some nice properties...

`zcash-local-net` is vastly simpler.

For example to test `zcash-local-net` I run:

```BASH
 cargo check && cargo check --tests && cargo fmt && cargo clippy && ./utils/trailing-whitespace.sh reject && cargo nextest run
```

`zcash-local-net` has 0 features (conditional compilation rules) and 0 optional dependencies.

The `tokei` stats for `zcash-local-net` show about 1559 lines of Rust.


Oscar-Pepper (changes requested):
I would iike some context to the issue we are trying to solve by removing zingolib from zcash-local-net test dependencies. the tests obviously do not live in zingolib, most of these dont even use the zingolib lightlclient, they generate the RPC requests directly

PR #1598, opened 2024-12-20
define a public testvectors crate.

zancas:
Necessary for `zcash-local-net` port has additional benefits:

* reduces dependency of libtonode-tests on zingolib
* reduces dependency of darkside-tests on zingolib
* removes separate testvectors feature gate from zingolib
* privatizes config_templaters inside testutils/scenarios
* makes testutils/scenarios represented on fs
etc.


zancas:
I recall @idky137 mentioning that zaino testutils depends on zcash-local-net in a way that we want to be careful to not break.  There's a draft PR in `zcash-local-net` that we'll need to ensure doesn't break `zaino-testutils` before we merge it:

https://github.com/zingolabs/zcash-local-net/pull/35

nachog00:
Summary [ 921.536s] 304 tests run: 304 passed (72 slow), 28 skipped

zingolib/src/testutils/scenarios.rs
```diff
@@ -0,0 +1,965 @@
...
+//! require some funds, the simplest way to access funds is to use a "faucet".
+//! A "faucet" is a client that receives mining rewards (because its spend capability
+//! generated the address registered as the `minetoaddress` in the zcash.conf that's
+//! used by the 'regetst mode' zcashs backing these tests.).
+//! HELPERS:
+//! If you just need a faucet, use the "faucet" helper.
+//! If you need a faucet, and a single recipient, use 'faucet_recipient`
+//! For less common client configurations use the client builder directly with
+//! custom_clients
+//! All scenarios have a default (i.e. faucet_default) which take minimal parameters and
```
Oscar-Pepper, line 12: this was true until someone renamed `faucet_funded_recipient_default` to `orchard_funded_recipient`! non-blocking suggestion to put that name back so it matches this comment

Oscar-Pepper (approved):
great! i left a non-blocking comment that we should consider but i will merge as is

PR #1601, opened 2024-12-25
Darkside eviction simulator

fluidvanadium:
/// this test demonstrates that
/// when a transaction is dropped by lightwallet, zingolib does not proactively respond
/// the test is set up to have a different result if proactive rebroadcast happens
/// there is a further detail to understand: zingolib currently considers all recorded transactions to be un-contradictable. it cant imagine an alternate history without a transaction until the transaction is deleted. this can cause it to become deeply confused until it forgets about the contradicted transaction VIA a reboot.
/// despite this test, there may be unexpected failure modes in network conditions


darkside-tests/tests/tests.rs
```diff
@@ -245,92 +246,124 @@ async fn sent_transaction_reorged_into_mempool() {
...
         .await
-        .unwrap();
-    environment.bump_chain().await;
+        .into_iter()
+        .map(|stepwise_result| {
+            stepwise_result
+                .map(|(fee_comparison_result, others)| (fee_comparison_result.unwrap(), others))
+                .unwrap()
+        })
+        .unzip();
```
Oscar-Pepper, line 311: this could be refactored to be a lot more readable. the idea here is very simple but there is a lot of mapping and nested results that is cumbersome to follow. non-blocking

Oscar-Pepper (approved):

PR #1592, opened 2024-12-16
Sync status

Oscar-Pepper:
ontop of #1597

adds a sync status that allows a user / consumer to get the status of sync while sync is running.

example test showing sync status:
`cargo nextest run --run-ignored all sync_status --nocapture --features sync`

PR #1603, opened 2024-12-28
Shard ranges

Oscar-Pepper:
ontpo of #1592

this completes and cleans up the sync state logic.
- when spends are detected, the block range containing all note commitments to the shard this note belongs to is scanned.
- chain tip is now correctly a block range that contains the incomplete shards of all shielded protocols
- fixes punch_scan_priority bug where a scan range that contains the given block range is not punched (split and updated)
- rewrite/clean-up of state fns


Oscar-Pepper:
closed for #1613

PR #1604, opened 2024-12-31
Fix reorg and improve continuity checks

Oscar-Pepper:
ontop of #1603 

fix re-org bug where is was assumed the chain tip would be scanned but not always the case when sync is interupted right at the beginning.

adds continuity checks to the end of a scan range to improve out of order sync safety

revisits wallet data cleanup after historic scans, now correctly uses `highest_scanned_height`


Oscar-Pepper:
closed for #1614

PR #1611, opened 2025-01-04
changed all runners to ziserv-2 (self-hosted)

nachog00:
We got a new server and a nice Kubernetes managed actions runner.
This should be pretty fast... I intend on bench marking this.

This PR isn't intended to be merged as of now, but I didn't mark it as a draft cause that would prevent the actions to trigger.


nachog00:
Having all jobs run on the self hosted runner was pretty ambitious.
I have to review/remember all the little caveats that made each job to run in this fashion possible/hard/convenient.

zancas:
Hi @nachog00 is this PR still relevant, or has it staled?

PR #1612, opened 2025-01-07
Sync status

Oscar-Pepper:
adds a sync status that allows a user / consumer to get the status of sync while sync is running.

example test showing sync status:
cargo nextest run --run-ignored all sync_status --nocapture


zingo-sync/src/primitives.rs
```diff
@@ -76,6 +88,15 @@ impl Default for SyncState {
     }
 }
 
+/// A snapshot of the current state of sync. Useful for displaying the status of sync to a user / consumer.
```
AArnott, line 132: Reporting % progress based on blocks is a straightforward way to do it. But it produces rather inconsistent perceived speed to the user when some blocks are nearly empty while others are full (e.g. the sandblasting period). From prior conversations with the ECC folks, it sounds like counting shielded outputs would produce a smoother progress reporting experience. This is more complicated though, and in particular if you're not scanning from start to finish but rather than several ranges to scan, figuring out how many outputs are in each of those ranges to be scanned will take some extra code and a few extra network calls. 
Something to consider.

Oscar-Pepper: Yes good comment. I have already considered this and all the framework is already there in the latest PR to implement this quite easily. The only issue is how would you solve the issue of the unknown number of outputs in the incomplete shards at the tip and also when starting sync from mid-shard?

AArnott: That's great to hear. The way I've done it (in a similar but not identical situation) is to download the tree state at each critical point (current scan position, and scan range endpoints). Since each tree state includes data about the size of the trees (i.e. the number of outputs), it's trivial to do (endpoint - startpoint) to calculate how many outputs are between any two points before scanning or downloading compact blocks begins.

Oscar-Pepper: Ah yes now I think of it, compact blocks actually contain chain metadata with the tree sizes as well

Oscar-Pepper: Even if this is None I fetch the tree states to make sure all wallet blocks have this info so this is actually even easier than I first thought!

Oscar-Pepper: Yeah this is a great idea thanks

Oscar-Pepper: ok I have implemented this suggestion

dorianvp (approved):

PR #1613, opened 2025-01-07
Shard ranges

Oscar-Pepper:
ontop of #1612 

this completes and cleans up the sync state logic.

- when spends are detected, the block range containing all note commitments to the shard this note belongs to is scanned.
- chain tip is now correctly scan a block range that contains the incomplete shards of all shielded protocols
- fixes punch_scan_priority bug where a scan range that contains the given block range is not punched (split and updated)
- rewrite/clean-up of state fns


zingo-sync/src/sync/state.rs
```diff
@@ -602,3 +733,89 @@ where
         orchard_final_tree_size: end_block.tree_boundaries().orchard_final_tree_size,
     }
 }
+
+/// Creates block ranges that contain all outputs for the shards associated with `subtree_roots` and adds the to
```
dorianvp:
```suggestion
/// Creates block ranges that contain all outputs for the shards associated with `subtree_roots` and adds them to
```

zingo-sync/src/primitives.rs
```diff
@@ -21,8 +24,8 @@ use crate::{
     utils,
 };
 
-/// Block height and txid of relevant transactions that have yet to be scanned. These may be added due to spend
-/// detections or transparent output discovery.
+/// Block height and txid of relevant transactions that have yet to be scanned. These may be added due transparent
```
dorianvp:
```suggestion
/// Block height and txid of relevant transactions that have yet to be scanned. These may be added due to transparent
```

dorianvp (changes requested):
A few typos in doc comments, the rest looks great!

PR #1615, opened 2025-01-07
Fixed output batching

Oscar-Pepper:
on top of #1613

this PR introduces the Batcher as a new task in the Scanner that solves the following issues:
- once a ScanWorker receives a task of X blocks, it must scan this range even if there is an unusually large amounts of outputs as it does not (and should not) have access to the wallet to update its now-smaller scan range for other workers to pick-up the rest
- due to this, an arbitrarily large amount of memory might be needed to scan some ranges i.e. in the sandblast regions
-  when scanning linearly, the workers are both given scan tasks (which contain the optional seam blocks [or adjacent blocks to the scan range]) so the workers will not have access to these for continuity checks between the "seams" of scan ranges as the first worker will not have finished and therefore added the seam block to the wallet

the Batcher solves these issues by:
- receiving the initial scan task and filling it with the compact blocks needed for scanning that range
- when the outputs count exceeds a maximum, splits the scan task into two and sends the full batch to a ScanWorker for scanning
- as its the only part of the sync engine that fetches compact blocks, it can manage the seam blocks by sending the highest block of the full batch to the new batch etc.
- also saves the first and last blocks of the previous scan task so it can manage seam blocks when scanning linearly in all cases for full continuity check coverage


zingo-sync/src/primitives.rs
```diff
@@ -21,27 +24,84 @@ use crate::{
     utils,
 };
 
-/// Block height and txid of relevant transactions that have yet to be scanned. These may be added due to spend
-/// detections or transparent output discovery.
+/// Block height and txid of relevant transactions that have yet to be scanned. These may be added due transparent
```
dorianvp:
```suggestion
/// Block height and txid of relevant transactions that have yet to be scanned. These may be added due to transparent
```

zingo-sync/src/scan/task.rs
```diff
@@ -139,15 +170,24 @@ where
 
     /// Updates the scanner.
     ///
-    /// If verification is still in progress, do not create scan tasks.
-    /// If there is an idle worker, create a new scan task and add to worker.
-    /// If there are no more range available to scan, shutdown the idle workers.
+    /// Creates a new scan task and sends to batcher if it's idle.
+    /// The batcher will stream compact blocks into the scan task, splitting the scan task when the maximum number of
+    /// outputs is reached. When a scan task is ready is it stored in the batcher ready to be taken by an idle scan
```
dorianvp:
```suggestion
    /// outputs is reached. When a scan task is ready it is stored in the batcher ready to be taken by an idle scan
```

dorianvp (changes requested):

dorianvp (approved):

PR #1619, opened 2025-01-15
add bool to set do_addresses to shielded_only

zancas:


zingolib/src/commands.rs
```diff
@@ -721,9 +721,9 @@ struct AddressCommand {}
 impl Command for AddressCommand {
     fn help(&self) -> &'static str {
         indoc! {r#"
-            List current addresses in the wallet
+            List current addresses in the wallet, shielded exludes t-addresses.
             Usage:
-            address
+            address true|false
```
juanky201271: there is a typo here, the command is: `addresses`.

zingolib/src/commands.rs
```diff
@@ -732,8 +732,20 @@ impl Command for AddressCommand {
...
+                "true" => {
+                    RT.block_on(async move { lightclient.do_addresses(true).await.pretty(2) })
+                }
+                "false" => {
+                    RT.block_on(async move { lightclient.do_addresses(false).await.pretty(2) })
+                }
+                _ => self.help().to_string(),
+            },
+            _ => self.help().to_string(),
+        }
```
juanky201271, line 745: Maybe can be better if `args.len === 0` the command assume `false` (it would work as before... no breaking changes...) IMO.

zancas: I changed it, so that you *must* say "truie" or "false"..  which is a little bit more work, but forces the reader to understand more.

zingolib/src/commands.rs
```diff
@@ -721,9 +721,9 @@ struct AddressCommand {}
 impl Command for AddressCommand {
     fn help(&self) -> &'static str {
         indoc! {r#"
-            List current addresses in the wallet
+            List current addresses in the wallet, shielded exludes t-addresses.
```
dorianvp:
```suggestion
            List current addresses in the wallet, shielded excludes t-addresses.
```

dorianvp (changes requested):

dorianvp (dismissed):

zancas:
Per @juanky201271 's suggestion I will preserve back compatibility!
I'm also going to add a test.

zancas:
I am running a CLT, there's a new test since this PR passed review.

juanky201271 (approved):
I tested and worked as expected. Good job.

PR #1620, opened 2025-01-15
Make replyto orchard only

zancas:


zancas:
There are `expects` that could panic in `command.rs`..  I will look into defining a proper error soon.

juanky201271 (approved):
I tested and works well for the three scenarios...

PR #1623, opened 2025-01-17
orchard only ua

zancas:


zingolib/src/commands.rs
```diff
@@ -228,71 +228,91 @@ impl Command for ParseAddressCommand {
...
-                            Address::Tex(_) => {
-                                object! {
-                                    "status" => "success",
-                                    "chain_name" => chain_name_string,
-                                    "address_kind" => "tex",
-                                }
-                            }
-                        }
-                    },
+        if args.len() > 2 {
```
juanky201271: you need to change this 2 by a 1, since there is only one parameter `address`.

juanky201271 (changes requested):

juanky201271 (approved):

PR #1629, opened 2025-01-22
Backported sub ua

zancas:
Also includes code to grab the first N VTs in the JSON String.

In CLT of both this feature and `dev_2-0` 206 tests passed.

### DIFF
--- a/zingolib/src/commands.rs
+++ b/zingolib/src/commands.rs
@@ -242,7 +242,7 @@ impl Command for ParseAddressCommand {
     }
 
     fn exec(&self, args: &[&str], _lightclient: &LightClient) -> String {
-        if args.len() > 1 || args.len() == 0 {
+        if args.len() > 1 || args.is_empty() {
             return self.help().to_string();
         }
         fn make_decoded_chain_pair(
--- a/zingolib/src/commands.rs
+++ b/zingolib/src/commands.rs
@@ -230,7 +230,7 @@ impl Command for ParseAddressCommand {
         indoc! {r#"
             Parse an address
             Usage:
-            parse_address [address]
+            parse_address <address> [orchard]
 
             Example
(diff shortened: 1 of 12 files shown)


zingolib/src/commands.rs
```diff
@@ -242,71 +242,91 @@ impl Command for ParseAddressCommand {
...
-                            Address::Tex(_) => {
-                                object! {
-                                    "status" => "success",
-                                    "chain_name" => chain_name_string,
-                                    "address_kind" => "tex",
-                                }
-                            }
-                        }
-                    },
+        if args.len() > 1 {
```
Oscar-Pepper: the change from only 1 to >1 implies there is now a way to use parse address with 0 args?

zingolib/src/commands.rs
```diff
@@ -242,71 +242,91 @@ impl Command for ParseAddressCommand {
...
+                    }
+                    if ua.sapling().is_some() {
+                        receivers_available.push("sapling")
+                    }
+                    if ua.transparent().is_some() {
+                        receivers_available.push("transparent")
+                    }
+                    if ua.orchard().is_some()
+                        && ua.sapling().is_some()
+                        && ua.transparent().is_some()
```
Oscar-Pepper: do we want to return orchard only in the case there is orchard and transparent receivers?

Oscar-Pepper (approved):
i have approved with comments to consider before merging

zingolib/src/commands.rs
zancas: Fixed

zingolib/src/commands.rs
zancas: Sure, why not always offer the orchard_only_ua.. if there's an orchard receiver available?

PR #1634, opened 2025-01-27
Deprecate status get height

fluidvanadium:
This function used to have a warning comment. Now it has a deprecation. It recently caused a bug for a member of the team that other people have used this function.

This PR cant merge as is because clippy wont like it.


zancas:
Why can't this merge?

Oscar-Pepper:
`get_height` has very useful utility, I don't think we should remove functions that serve a unique purpose because people are not to careful enough to misuse them. I will take a look at how I have used in sync

fluidvanadium:
get height returns a value that either represents when the transaction was Calculated, Transmitted, Mempool, or Confirmed. These numbers are typically different by a couple of blocks.
then it forgets which case

it has been useful as a blunt instrument, for example in reorgs. but it has caused Problems

Oscar-Pepper:
> get height returns a value that either represents when the transaction was Calculated, Transmitted, Mempool, or Confirmed. These numbers are typically different by a couple of blocks. then it forgets which case
> 
> it has been useful as a blunt instrument, for example in reorgs. but it has caused Problems

ok could we replace it with a "get_pending_height" then so we can access the height of unconfirmed txs also then?

Oscar-Pepper:
well, there is a couple of times that I will need to write uneccessary logic to retrieve the height if this method didnt exist. I think its best we add the "pending height" method and just push to use confirmed_height and pending_height where possible and only get_height if its necessary

zancas:
Hi @fluidvanadium is this still relevant now that we've upgrade dev to pepper-sync?

Oscar-Pepper:
im closing this as it would make a lot of logic unnecessarily complicated and difficult to manage. i am for adding an additional "unconfirmed_height" and advise consumers to use it where possible though

PR #1635, opened 2025-01-27
Deprecate status get height

fluidvanadium:
This should be able to go. I will cherry pick it.


zancas:
#1634 has the same name, should we close one of these?

PR #1637, opened 2025-01-31
fix spending of immature coinbase utxos

Oscar-Pepper:


zancas:
I'm not convinced that we should add this workaround, if it's just to support the edge case of receiving transparent coinbase transactions.

I appreciate the effort, but maybe we should prioritize mining-to-orchard in zebrad *instead*.

zancas:
I stand corrected.  I misunderstood the purpose.

Would this change be stable across the pepper-sync integration, if it landed?

Oscar-Pepper:
yes but we actually will not need to port this to 2.0 as it will have a better solution to this issue and will not be using this transparent output type anyway

pacu (approved):
Untested but looks good to me.

Non-blocking: add a test to verify that output that are not coinbase don't get flagged as such

PR #1644, opened 2025-02-03
fix RUSTSEC-2025-0004

rex4539:
https://rustsec.org/advisories/RUSTSEC-2025-0004


fluidvanadium (approved):
essential! good eye

fluidvanadium:
Thank you @rex4539

PR #1646, opened 2025-02-03
Broadcast insistor

fluidvanadium:
This is nearly ready. 

The only thing holding me back from asking merge right now is
I feel we need more multi-step transaction tests.


zancas:
Is this relevant or stale?

PR #1659, opened 2025-02-16
handle invalid txid errors

Oscar-Pepper:
handles user facing panic where server rejects a transaction and returns error string that is not hex decodable


fluidvanadium (approved):
This is good. I don't know what the deal is with the features, maybe its worth me looking into.

PR #1665, opened 2025-02-18
Fix sync hang

Oscar-Pepper:
fixes issues with mempool where under certain conditions it gets stuck awaiting a mempool stream

### DIFF
--- a/pepper-sync/src/scan/error.rs
+++ b/pepper-sync/src/scan/error.rs
@@ -8,7 +8,7 @@ pub enum ScanError {
     ContinuityError(#[from] ContinuityError),
 }
 
-/// Block continuuity errors.
+/// Block continuity errors.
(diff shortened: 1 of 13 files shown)


pepper-sync/src/scan/error.rs
```diff
@@ -1,18 +1,23 @@
...
 
+/// Scan errors.
 #[derive(Debug, thiserror::Error)]
 pub enum ScanError {
+    /// Continuity error.
     #[error("Continuity error. {0}")]
     ContinuityError(#[from] ContinuityError),
 }
 
+/// Block continuuity errors.
```
dorianvp:
```suggestion
/// Block continuity errors.
```

dorianvp (changes requested):

dorianvp (approved):

Issue #1674, opened 2025-02-26
Pepper-sync: locator API and targetted rescan


Oscar-Pepper:
there needs to be an API for consumers of pepper-sync to add locators to the SyncState and for the sync engine to prioritise the scan ranges accordingly. this is needed for wallet data migration from zingo version 1.X to zingo 2.0.

PR #1689, opened 2025-03-06
Load from previous wallet versions

Oscar-Pepper:
adds API for targetted scan which is leveraged to re-scan zingo2 wallets very quickly by extracting all relevant transaction height and txids from zingo 1.x versions without having to parse between incompatible old and new wallet types. due to spend-before-sync and batching by shards, the notes in transactions that are targetted for scanning are almost immediately spendable so users will not have to wait when upgrading from zingo 1 to 2


Oscar-Pepper:
Fixes #1674

PR #1697, opened 2025-03-09
Fix incomplete tree bug

Oscar-Pepper:
adds frontiers to new wallets to complete the tree for witness construction

can be tested by running the ignored "initial_frontier_test". it will fail if the new "add_initial_frontiers" fn is commented out and pass when uncommented


zancas:
This seems like an issue that belongs on shardtree?!

Oscar-Pepper:
> This seems like an issue that belongs on shardtree?!

no shard tree is just a store, it requires the consumer to add any relevant data. when scanning from a birthday, the frontier is needed to fill in all the relevant nodes related to note commitments from blocks below the birthday height

dorianvp (approved):
utACK

dorianvp (approved):
ACK

PR #1704, opened 2025-03-15
Fix chain generics

Oscar-Pepper:
fixed failing chain generic tests and also set transparent address discovery to enabled by default for testing


dorianvp (approved):
tACK

PR #1705, opened 2025-03-15
Fix encoded memo parsing

Oscar-Pepper:
test "diversified_addresses_receive_funds_in_best_pool" is now passing and pepper sync's encoded memo parsing has been improved with unwraps replaced


dorianvp (approved):
tACK

PR #1725, opened 2025-04-05
cargo update, to stabilize deps before removing openssl-sys dep

zancas:


zancas:
I ran the full test suite locally.

PR #1739, opened 2025-04-09
fix intra-doc links

zancas:


ala-mode:
Builds locally!

ala-mode (approved):
Builds with `cargo doc --document-private-items --all-features`

PR #1743, opened 2025-04-16
remove unused zingomobile fn

Oscar-Pepper:


fluidvanadium:
Noticing that integration tests ran and passed successfully

juanky201271 (approved):
We don't use it...

Issue #1753, opened 2025-04-20
Cannot load testnet wallet


fluidvanadium:
More info here:

https://github.com/zingolabs/zingolib/pull/1744#issuecomment-2814522582

fluidvanadium:
The wallet file is corrupted by opening in dev-2.0

fluidvanadium:
>             scope: external
>         }
>         {
>             value: 4998525000
>             memo:
>             recipient: utest1dmu08vg5wt5m9w0ejwgeqdndzzkfka7c94heuz5llxv5vx4lhcethmm6p5ean3wcj8l0m6tf8k8cau9r636gq7sxq3wa6zey2sysfteh
>             recipient unified address: utest1dfj7pz3v2hewxk5nt36h9mu9g5h9cuc3064h5jlzgnq9f9f74fv4vr4vz4yvp2mq36ave8y8ggghm4g0e2nw9fg7qewuflpxxtvq4dzpscws4eq0hlh09shhj83a5yuzm9p3hsrtd028xk37gyf6403g90jwvtxkgyqv25kzzafyjfl7rvexzvrjx24akdr83qzkssyg22jm5cgcxc9
>             output index: 1
>             account id: 0
>             scope: internal
>         }
>         {
>             value: 10000
>             memo:
>             recipient: utest1f7hprvzygjeq6gtyh2splf0qq3ug70krvq44z40d7lg590u2dwg59l8we9v2mdpk72llg5wd9cat68ur0pdrmyx8fs54x6asy5z86elr
>             recipient unified address: utest17wwv8nuvdnpjsxtu6ndz6grys5x8wphcwtzmg75wkx607c7cue9qz5kfraqzc7k9dfscmylazj4nkwazjj26s9rhyjxm0dcqm837ykgh2suv0at9eegndh3kvtfjwp3hhhcgk55y9d2ys56zkw8aaamcrv9cy0alj0ndvd0wll4gxhrk9y4yy9q9yg8yssrencl63uznqnkv7mk3w05
>             output index: 2
>             account id: 0
>             scope: external
>         }
>     outgoing sapling notes:
>         {
>             value: 10000
>             memo:
>             recipient: ztestsapling17d9jswjas4zt852el3au4fpnvelc5k76d05fga2rdm9zttac29m8g202dkz8wsfcd7r4c8q540n
[...]

PR #1757, opened 2025-04-26
Fix v32 read write

Oscar-Pepper:
on top of #1755

fixes critical read write error in v32 (zingo 2.0 wallet version)

also removes legacy unused price and option


fluidvanadium (approved):
fixes #1753

dorianvp (approved):

PR #1759, opened 2025-04-28
add percentage total outputs scanned field to sync status

Oscar-Pepper:
- rewrites sync status
- merges scanned ranges during sync
- quarters batch size to make UI more responsive


juanky201271 (approved):
tested! the sync process stuck issue is fixed & the total percentage is working...

PR #1762, opened 2025-04-30
fixed sync handling and added sync stop

Oscar-Pepper:


juanky201271 (approved):
All the test passed... good enough to me.

PR #1774, opened 2025-05-01
added sync config to optimise sync setup times

Oscar-Pepper:


juanky201271 (approved):
I trust you @Oscar-Pepper ... nice job.

PR #1777, opened 2025-05-02
Improve rescan pt2

Oscar-Pepper:
- fixes bug where shard tree is not cleared on rescan
- adds locators for targetted rescan


juanky201271 (approved):
perfect.

Issue #1714, opened 2025-03-22
Zingo 2: Price list


Oscar-Pepper:
implement fiat price list for zingo 2. this should be a wallet feature and not part of pepper-sync. I imagine a price list is gathered for 00.00 of each day since birthday height for the fiat currency the user selects. `transaction_summaries` will then query the price list against the tx's datetime / height and include this price in the summary. it would be more efficient to gather a price list for only days where a user has a transaction but we should consider the privacy impliciations

PR #1788, opened 2025-05-21
current price fetch reimpl with tor

Oscar-Pepper:


zancas:
I think we're considering caching historical data in the release.

Oscar-Pepper:
> I think we're considering caching historical data in the release.

ok i have removed the historical data fetching, we can reuse the price list logic for however we implement getting this data later on

Oscar-Pepper:
ok ready to go. it now has zingo1 non-tor price fetch option also

docker-ci/Dockerfile
```diff
@@ -56,6 +56,7 @@ RUN apt update \
     libssl-dev \
     build-essential \
     protobuf-compiler \
+    libsqlite3-dev \
```
juanky201271: I think with the `bundled` feature in `rusqlite` maybe you can remove this safely from the Docker build... I'm not sure 100% but is wothy to try...

zancas (changes requested):
LGTM, pending resolution of @juanky201271 's comment.

docker-ci/Dockerfile
Oscar-Pepper: ok yes you were right I have reverted the CI version and passing CI

zancas (approved):

Oscar-Pepper:
Fixes: #1714

PR #1789, opened 2025-05-21
Multi account support

Oscar-Pepper:


juanky201271 (approved):
I made it work with only one account (default)... in zingo-mobile.

PR #1806, opened 2025-05-26
`zingolib`: Reintroduce `zcash-params` for mobile devices

dorianvp:
This PR fixes an issue where the sapling params where not being included for mobile bundles. 
Now they are copied from the default params location to the previous location.


Oscar-Pepper (approved):
ideally wouldnt atleast replace unwraps with expects and eprintlns for errors (or put message inside panic) but non-blocking as this fix needs to land!

PR #1807, opened 2025-05-26
remove price from summaries

Oscar-Pepper:


juanky201271 (approved):
I'm sure it is perfect.

PR #1819, opened 2025-06-02
fix bugs with address mismatches

Oscar-Pepper:
i realised there were some edge cases that i didnt consider in address mismatching when reloading from seed with unified address discovery and also when generating addresses in other wallets

this PR fixes these issues and also is more efficient. using encoded memos to restore UAs to their exact receiver selections is a very desirable feature now due to the limitations here.


juanky201271 (approved):
I don't see any problems... nice job.

PR #1820, opened 2025-06-02
fix batcher timeout bug

Oscar-Pepper:
mitigates reaching timeout when scanning on slow devices. when timeout is detected, instead of returning an error it calls the rpc again from the updated block height


juanky201271 (approved):
nice improvement.

PR #1823, opened 2025-06-03
fix transparent receiver rw

Oscar-Pepper:


Oscar-Pepper:
closed for future PR which optionally enforces no gap transparent address generation

PR #1824, opened 2025-06-04
Safe taddr generation

Oscar-Pepper:
- add no gap enforcement option to keep sync initialisation very efficient
- remove transparent receiver support for UAs due to an issue with potentially missing transparent funds and also general support from community due to privacy concerns
- remove a bunch of kruft no longer needed in legacy read/write


juanky201271 (approved):
nicely done!

PR #1825, opened 2025-06-04
Reset addresses from test version

Oscar-Pepper:
fixes transparent address missing when loading from previous zingo mobile test version


juanky201271 (approved):
better & better...

PR #1827, opened 2025-06-05
change mutex wallet lock for rwlock

Oscar-Pepper:


juanky201271 (approved):
tests passed... that is a good sign...

PR #1828, opened 2025-06-06
Add initial `tracing` support

dorianvp:
This PR adds tracing. In order to see the events, prefix the run command with `RUST_LOG=info`.


zancas (dismissed):
We need to be clear that logging isn't leaking in release!

zancas (approved):

PR #1831, opened 2025-06-09
rework spendable balance

Oscar-Pepper:
- updates spendable balance to check witness can be constructed
- changes `spendablebalance` command to return orchard and sapling spendable balances
- adds `max_send_value` method and command to show maximum value that can be sent to a given address (previously `spendablebalance`)
- adds `account_balance` wrapper to lightclient
- removes async from balance fns
- adds minimum confirmations to wallet settings to be used in spendable notes logic
- removes dust from confirmed and unconfirmed balances in `account_balance` wallet method


juanky201271 (approved):
Another step. Thanks.

PR #1834, opened 2025-06-11
stabilise spendable balance and spend before sync

Oscar-Pepper:
- changes note selection to prioritise notes that are guaranteed spent based on the current state of the nullifier map
- add bool parameter to `spendable_balance` wallet method to return either spendable balance or potentially spendable balance (useful for sending funds out of old wallets which would have to sync for along time before the spendable funds show, making them recoverable much earlier to the user)
- update spendable balance command to return both balances

### DIFF
--- a/zingolib/src/commands.rs
+++ b/zingolib/src/commands.rs
@@ -658,7 +658,10 @@ struct SpendableBalanceCommand {}
 impl Command for SpendableBalanceCommand {
     fn help(&self) -> &'static str {
         indoc! {r#"
-            Display the wallet's spendable balance for each shielded pool.
+            Display the wallet's spendable balance.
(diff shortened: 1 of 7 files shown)


zancas:
I don't think we can get this new feature into this release.   Shall we update stable, **before** merging this new development feature?

Oscar-Pepper:
> I don't think we can get this new feature into this release. Shall we update stable, **before** merging this new development feature?

this is not a new feature, this is a huge bug fix. without it we would be disabling all spend before sync and causing the user confusing UX and send failures

juanky201271 (approved):
I need to merge this in order to test the other PR... I didn't see here any problem.

PR #1836, opened 2025-06-14
Add performance level to sync config

Oscar-Pepper:


juanky201271 (approved):
I don't see any problems... good job.

PR #1838, opened 2025-06-18
add spendable balance for wallet read lock

Oscar-Pepper:


juanky201271 (approved):
We need to test manually with all the PR's. Thanks @Oscar-Pepper

PR #1850, opened 2025-07-02
fix missing block bounds for nullifier refetching

Oscar-Pepper:


Oscar-Pepper:
i have a better solution in following PR

PR #1851, opened 2025-07-02
improve block retention

Oscar-Pepper:


Oscar-Pepper:
better solution in following PR (3rd attempt)

PR #1854, opened 2025-07-07
invalid txs restore balance

Oscar-Pepper:
- txs in mempool for 3 blocks or calculated/transmitted txs that have expired (40 blocks) now automatically reset spent notes to restore balance and allow to be respent
- fixes re-org bug where transparent spends are not reset


juanky201271 (approved):
looks good.

PR #1874, opened 2025-08-10
deps(json): replace with `serde_json`

dorianvp:


zancas:
@dorianvp is this still something we intend to land?

PR #1878, opened 2025-08-16
Add --waitsync parameter to zingo-cli to run the one-shot commands after background sync completes

Tomas-M:
**Problem**

zingo-cli launches a background sync unless --nosync is passed, but it executes the requested command immediately. For one-shot invocations like:

```zingo-cli --data-dir=/path balance```

the command often runs before sync finishes, returning stale or partial data.

**Solution**

Introduce a new CLI flag, --waitsync, that makes zingo-cli wait for the background sync to complete before executing the requested command. This only affects non-interactive mode.

**Example**

```
# Wait for sync to finish, then print the balance
zingo-cli --data-dir=/abs/path --waitsync balance

# No effect if sync is disabled
zingo-cli --data-dir=/abs/path --nosync --waitsync balance
```

**User-visible behavior**

When --waitsync is present and sync is enabled, zingo-cli:

- Starts or detects the background sync
- Polls sync status until completion (or an error)
- Executes the requested command and exits
- When --nosync is present, --waitsync is ignored.

Interactive mode is unchanged.

### DIFF
--- a/zingocli/src/lib.rs
+++ b/zingocli/src/lib.rs
@@ -40,6 +40,10 @@ pub fn build_clap_app() -> clap::ArgMatches {
                 .long("nosync")
                 .short('n')
                 .action(clap::ArgAction::SetTrue))
+            .arg(Arg::new("waitsync")
+                .help("Block execution of the specified command until the background sync completes. Has no effect if --nosync is set.")
+                .long("waitsync")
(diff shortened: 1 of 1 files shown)


zancas:
Thanks for contributing!

dorianvp (approved):

PR #1894, opened 2025-08-18
Manual note selection for quicksend command by transaction id

Tomas-M:
Zingo-cli manages note selection automatically.
In order to allow the user to spend note from specific transaction id (txid) only, this patch adds quick support for that.
It accepts optional txid argument and filters the automatically-selected notes by this txid.

Currently for quicksend only. This could be perhaps useful for "send" command too, but I don't use it so I skipped it due to additional complexity. I am not sure if you even like to include such filter in zingolib. I will be happy for feedback if you have any.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,12 +18,16 @@ zingodisclosure@proton.me
 * Price information is retrieved from Gemini exchange.
 
 ### Note Management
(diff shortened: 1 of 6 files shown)


Tomas-M:
The automatic jobs are failing for reason beyond my control. Not my fault.
action requires bash, exit 1

PR #1895, opened 2025-08-21
bugfix import unified full viewing key ufvk

Tomas-M:
When trying to start new wallet by importing unified viewing key to zingolib, it fails because it considers the ufvk as a seed word. It was practically impossible to import the viewing key. This patch fixes this.

The key is imported and wallet is started successfully.
When I export ufvk again from the newly created wallet, it matches the imported key byte by byte.

However, when I try "notes" or "transactions" command, I do not see any notes or transactions, even if there were many incomming and outgoing payments, even if I rescan and sync since the birthday. So I am not sure if the whole thing is correct, may need further fixing or improvements.


Oscar-Pepper (changes requested):
Thanks for your fix! sorry for the inconvenience, zingo cli is not as easy to intetgration test atleast in its current form. Please can you look at fixing the fmt to pass CI? you should be able to just run `cargo fmt`

Tomas-M:
tests passed now, thanks for pointing out what to do.

Still, my patch adds the ufvk import, but I am not sure if this is everything which needs to be done, since I cannot see any transactions or notes on the wallet with imported ufvk.

Tomas-M:
I think there was probably something misplaced on my side when I couldnt see the transactions or notes, I tried again and it works fine now. So we can consider this patch fully functional to add working viewkey support.

Oscar-Pepper:
> I think there was probably something misplaced on my side when I couldnt see the transactions or notes, I tried again and it works fine now. So we can consider this patch fully functional to add working viewkey support.

yes i tested it and my txs appeared so I consider this fully functional

Oscar-Pepper (dismissed):

Tomas-M:
Good. Would you please merge it to main dev?

Oscar-Pepper (approved):

PR #1900, opened 2025-08-27
Lightwallet protocol rust

zancas:
This makes zingo-netutils independent of zcash_client_backend.

It also makes pepper-sync independent of zingo-netutils, though that's not known to be valuable.

Once this merges we can update infrastructure to use the new zcash_client_backend independent zingo-netutils.


AloeareV (approved):
Nice separation!

PR #1901, opened 2025-08-27
Bump `zcash_client_backend` to 0.19.1

dorianvp:


juanky201271:
Just a note: We need to avoid using `ring` 0.16.20 because we can't build zingo-pc for Windows arm64. This issue is fixed for `ring` 0.17.X...

zancas:
Is this superseded by #1933 ?

PR #1905, opened 2025-08-28
Merge stable into dev

dorianvp:
Stable received some direct commits that need to be merged into dev.


dorianvp:
Moved into #1906 so that the merge conflict resolution does not involve pushing directly into dev.

PR #1917, opened 2025-09-02
generic local net test upgrade

Oscar-Pepper:
this is the first part of generalising the test infrastructure to also test zainod and zebrad.

the next step will be to work on the issue that zebrad does not mine to shielded pools and therefore cannot be a 1-to-1 drop-in for zcashd. the most likely solutions in further work will be to migrate all tests to *not* use a faucet directly so all assertions can be uniform across validators OR for zebrad to support mining to shielded pools


zancas:
We should open a PR against Zebra that enables mining to shielded pools.

zingolib/src/testutils/scenarios.rs
```diff
@@ -19,26 +19,232 @@ use std::sync::LazyLock;
...
+                chain_cache: None,
+                network: Network::Regtest,
+            },
+            ZcashdConfig {
+                zcashd_bin: ZCASHD_BIN.clone(),
+                zcash_cli_bin: ZCASH_CLI_BIN.clone(),
+                rpc_listen_port: None,
+                activation_heights,
+                miner_address: Some(miner_address),
+                chain_cache,
```
fluidvanadium, line 165: Maybe this default config instantiation could be a method. I have concerns about code duplication.

fluidvanadium: Check this out! https://github.com/zingolabs/infrastructure/pull/108

Cargo.toml
```diff
@@ -109,7 +109,7 @@ crossbeam-channel = "0.5"
 rayon = "1"
 
 # Zingolabs
-zingo-infra-services = { git = "https://github.com/zingolabs/infrastructure.git", branch = "dev" }
+zingo-infra-services = { git = "https://github.com/zingolabs/infrastructure.git", tag = "zingo-infra-services-0.2.0" }
```
fluidvanadium: In order to prevent circular dependencies, zingolib could build without this as an optional or dev- dependency.

zancas:
Let's see what needs to be done to land this, as soon as these conflicts are resolved.

zingolib/src/testutils/scenarios.rs
Oscar-Pepper: this is a good idea, i will take a note for a future PR so we can land this as is if thats ok

dorianvp (dismissed):
Looks good! For some weird reason the `cargo-hack` workflow failed, but I am approving this anyway as it happened in an unmodified file.

PR #1925, opened 2025-09-05
pepper-sync: pre-sapling_protection

Oscar-Pepper:
fix potential pre-sapling issues in the edge case that the wallet has transparent data just before the sapling epoch or when the test chain only contains the genesis block


idky137 (approved):
Looks good!

I haven't yesterday tested that this does fix the error in zaino tests but this looks like a good robustness improvement either way!

PR #1943, opened 2025-09-15
ConfirmationStatus unit test and debug

fluidvanadium:


zancas (approved):
This LGTM...   I'd like to get @Oscar-Pepper 's input.. if he has time..  but I marked this approved.

zingo-status/src/confirmation_status.rs
```diff
@@ -250,6 +250,7 @@ impl ConfirmationStatus {
     }
 }
 
+/// a public interface, writ in stone
```
zancas, line 253: Requires updating the semver and CHANGELOG..  "stones"...

zancas: Our policy is to precede deprecation with this:

https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute

PR #1944, opened 2025-09-15
updated dependency to match zaino

fluidvanadium:


zancas (commented):
If zaino and zingolib were in the same workspace, and the lock file were checked in.  And all version policies were set in the workspace manifest (i.e. cargo autoinherit), then Rust tools would enforce matching dependencies across projects.

fluidvanadium:
> If zaino and zingolib were in the same workspace, and the lock file were checked in. And all version policies were set in the workspace manifest (i.e. cargo autoinherit), then Rust tools would enforce matching dependencies across projects.

After today, I am particularly convinced we must not have cross dependencies in this way.

PR #1946, opened 2025-09-15
Proptest too slow

zancas:


fluidvanadium:
According to this CI run: https://github.com/zingolabs/zingolib/actions/runs/17743251997/job/50422837580
The proptest runs in 125 seconds. Not the slowest test.

zancas:
Is the most likely explanation that these tests depend on testnet, and network weather caused them to hang?

If that is what happened..   how should we handle that?   Do we want a clean CI run to be defined to include fair network weather?

fluidvanadium:
> Is the most likely explanation that these tests depend on testnet, and network weather caused them to hang?
> 
> If that is what happened..   how should we handle that?   Do we want a clean CI run to be defined to include fair network weather?

They do not depend on network weather, being libtonode tests

zancas:
Well...   sometimes they don't hang...

PR #1949, opened 2025-09-17
update ring crate dependency version

ala-mode:
Simple `Cargo.toml` dependency update, requiring a higher version.

see https://github.com/zingolabs/zingolib/security/dependabot/38
and
https://github.com/zingolabs/zingolib/security/dependabot/35


ala-mode:
All tests pass for me locally.

juanky201271:
This is a good idea, but if you look into your Cargo.lock you can still find a ring version 0.16.x

https://github.com/ala-mode/zingolib/blob/b30b2b30b33b8be81274532609d4a7052b17813e/Cargo.lock#L4943

And if you research that dependency, you can see the problem is in zcash_client_backend with this dependency: arti_client 0.28 (tor-rtcompat)
For `arti_client` version 0.29 -> no ring 0.16.x as dependency (because not use x509-signature anymore...)

tip: x509-signature have ring 0.16.20 as dependency.

Oscar-Pepper (changes requested):
I'm in favor of closing this PR on the grounds that I think it's probably better to allow our library to be compatible with earlier patches of ring v0.17 than enforcing consumers must use the latest version

juanky201271:
> I'm in favor of closing this PR on the grounds that I think it's probably better to allow our library to be compatible with earlier patches of ring v0.17 than enforcing consumers must use the latest version

For me, this is an old big problem because meanwhile we have ring v.0.16.X as a dependency, we cannot build zingo-pc for Windows arm64. The next version v0.17.X fixed that situation...

ala-mode:
@juanky201271 @Oscar-Pepper  I don't have a strong insight or position on this, but I noticed this issue, so I am glad we can discuss it.

I think this particular 'vuln' is more in a DoS category...

5.3 severity:
"Scope: Availability"
https://cwe.mitre.org/data/definitions/770.html
DoS for CPU, memory and 'other'

So if this is going to cause problems, maybe it would be best to either leave it open, or close and open another issue to track it?

Oscar-Pepper:
> > I'm in favor of closing this PR on the grounds that I think it's probably better to allow our library to be compatible with earlier patches of ring v0.17 than enforcing consumers must use the latest version
> 
> For me, this is an old big problem because meanwhile we have ring v.0.16.X as a dependency, we cannot build zingo-pc for Windows arm64. The next version v0.17.X fixed that situation...

ah ok, so the issue is that we depend on crates using ring 0.16 and they have not updated their ring dependency? we should open an issue for this but i dont think this is related to this PR as there will be two versions on ring being built, where this PR enforces our miniumum version of our direct dependency and the other ring verson is due to downstream deps

juanky201271:
> > > I'm in favor of closing this PR on the grounds that I think it's probably better to allow our library to be compatible with earlier patches of ring v0.17 than enforcing consumers must use the latest version
> > 
> > 
> > For me, this is an old big problem because meanwhile we have ring v.0.16.X as a dependency, we cannot build zingo-pc for Windows arm64. The next version v0.17.X fixed that situation...
> 
> ah ok, so the issue is that we depend on crates using ring 0.16 and they have not updated their ring dependency? we should open an issue for this but i dont think this is related to this PR as there will be two versions on ring being built, where this PR enforces our miniumum version of our direct dependency and the other ring verson is due to downstream deps

You are right man... I agree... we can merge this PR right now & if someone wants to solve this issue in zcash_client_backend, that is another topic.

juanky201271 (approved):
I don't see any problem merging this...

Oscar-Pepper (approved):

PR #1957, opened 2025-09-20
Add command new_taddress_allow_gap

Tomas-M:
This change adds new command new_taddress_allow_gap, to create a new transparent address even if the current one has not received funds.

There are legitimate situations where the user may need to generate multiple addresses in advance for organizational or privacy reasons without immediately using each one.

This new command bypasses the built-in "no-gap" rule for new_taddress, and provides detailed explanation in help, so the user can take full responsibility to choose what (s)he needs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ This will launch the interactive prompt. Type `help` to get a list of commands.
 
 ## Running in non-interactive mode:
 You can also run `zingo-cli` in non-interactive mode by passing the command you want to run as an argument. For example, `zingo-cli addresses` will list all wallet addresses and exit.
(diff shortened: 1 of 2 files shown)


Oscar-Pepper (approved):
I agree this is a useful command. Thanks for providing a good warning and description of why it would degrade the sync performance to use this command and increase the gap limit during scanning. We actually don't currently support syncing with a gap limit for zingi-cli, see here:
https://github.com/zingolabs/zingolib/blob/dev/zingo-cli/src/lib.rs#L407

i will merge this PR but i will also create an issue to add a --gap-limit flag to zingo-cli so you can recover funds when rescanning after using this command. I will try to land this ASAP

PR #1972, opened 2025-10-02
cargo update

zancas:
Running tests locally in parallel.


fluidvanadium (approved):
This may interact with dependencies. Hopefully in a positive way.

PR #1978, opened 2025-10-08
Interface Change: Added an error type, replacing a string, in a pub fn

fluidvanadium:
Also removed conditional compilation that was causing issues.
Now on top of #1981


zingo-cli/src/regtest.rs
```diff
@@ -2,32 +2,16 @@
...
-            configured_activation_heights: for_test::all_height_one_nus(),
-            miner_address: Some(REG_O_ADDR_FROM_ABANDONART),
-            chain_cache: None,
-        },
-    )
-    .await
+pub(crate) async fn launch_local_net() -> LocalNet<Zcashd, Lightwalletd> {
+    LocalNet::launch_default()
+        .await
+        .expect("ing to launch a Regtest LocalNet")
```
dorianvp:
```suggestion
        .expect("a Regtest LocalNet should've launched")
```

dorianvp (dismissed):

zingo-cli/src/regtest.rs
fluidvanadium: I usually get confusing when I hit expect panics, and I am using expect("ing to give a hint that I need to process the error on a different level.

dorianvp: I read an "expect" as "expect that".

fluidvanadium: Fixed.

dorianvp (approved):

PR #1983, opened 2025-10-09
zwi syncs, benchmarks

fluidvanadium:
Why:

The interface to zingolib is scattered over many modules. I wanted to re-imagine what the interface could look like.

What:

The sum of assumptions necessary to create a wallet with zingolib is contained in code here.
The only change made to the zingolib package so far is the correction of a chain_from_str function.

Tests:

So far, this creates a ZingoWallet.
And tests that it reports syncking a chain.
Notably, this test does not involve customized test behavior. It only uses the interface methods presented in zcash_wallet_interface.

Testnet Benchmarks are dependent on the network, so this leaves something to be desired for benchmarks.

TODO: nextest profile


zingo_wallet/src/interface.rs
```diff
@@ -0,0 +1,191 @@
...
+#[derive(thiserror::Error, Debug)]
+pub enum AddKeyError {
+    #[error("Todo")]
+    AlreadyHasKey, //TODO
+}
+
+impl zcash_wallet_interface::Wallet for ZingoWallet {
+    fn user_agent_id() -> zcash_wallet_interface::UserAgentId {
+        const AGENT_NAME: &str = "zingo_wallet";
+        const VERSION: &str = "0.0.1";
```
zancas, line 57: I advocated the removal of the _foo.0_ accessor.

fluidvanadium: Good idea. Done!

zingo_wallet/src/interface.rs
```diff
@@ -0,0 +1,191 @@
...
+                .map_err(|invalid_uri| AddServerError::CantParseUri(server_address, invalid_uri))?;
+
+            let lightwalletd_uri: Arc<RwLock<Uri>> = Arc::new(RwLock::new(server_uri.clone()));
+
+            let (chain_type, birthday) = {
+                // we need to ask the indexer for this information
+
+                let mut client = {
+                    // global configuration must be manually set *somewhere*
+                    rustls::crypto::ring::default_provider().install_default();
```
zancas: Where should this be called?  Why?

Is this called on the last possible line before use?

zancas: I'm referring to the previous line of code.

zingo_wallet/src/interface.rs
```diff
@@ -0,0 +1,191 @@
...
+                    zingolib::grpc_client::get_zcb_client(server_uri.clone())
+                        .await
+                        .map_err(|e| AddServerError::CantCreateClient(server_uri.clone(), e))?
+                };
+
+                let lightd_info = client
+                    .get_lightd_info(tonic::Request::new(
```
zancas, line 128: What's a _lightd_ anyway?

zingo_wallet/src/interface.rs
```diff
@@ -0,0 +1,191 @@
...
+                    .await?
+                    .into_inner();
+
+                let chain_name = &lightd_info.chain_name;
+                let chain_type: ChainType = chain_from_str(chain_name)?;
+
+                let birthday: zcash_primitives::consensus::BlockHeight =
+                    lightd_info.block_height.try_into()?;
+                (chain_type, birthday)
+            };
```
zancas, line 142: Why not make the RHS of this let statement be a Single Use Helper Fn or "SUHFn".

This would suck code out of the body of this biggie fn, force an explicit type signature, inspire a name for the helper itself, and generally be elegant IMHO.

zancas: Also helper fns (as named items) are relatively easy to reference, and e.g. test.

zingo_wallet/src/interface.rs
```diff
@@ -0,0 +1,191 @@
...
+
+            // this seems like a lot of set up. Do we really need all this right here??
```
zancas, line 144: It seems to me that the wallet-related functionality defined from here to the end of this fn might be useful outside the context of server addition.

So why not factor this out as another fn?

zingo_wallet/src/interface.rs
fluidvanadium: I will. But its another step.

zingo_wallet/src/interface.rs
```diff
@@ -0,0 +1,202 @@
...
+    #[error("Seed parse from string '{0}' failed with >{1}<.")]
+    ParseSeed(String, bip0039::Error),
+    #[error("Wallet creation failed with >{0}<.")]
+    CreateLightWallet(#[from] zingolib::wallet::error::WalletError),
+    #[error("Temporary data dir creation failed with >{0}<.")]
+    CreateDataDir(#[from] std::io::Error),
+    #[error("Wallet creation failed with >{0}<.")]
+    CreateLightClient(#[from] zingolib::lightclient::error::LightClientError),
+    #[error("Wallet creation failed with >{0}<.")]
+    StartSync(zingolib::lightclient::error::LightClientError),
```
dorianvp, line 35: How can adding a server result in a sync error? Are we implying that it automatically starts a sync process?

fluidvanadium: Good point. Renamed the method.

libtonode-tests/tests/concrete.rs
```diff
@@ -685,10 +685,10 @@ mod fast {
             && vt.recipient_address == Some(ZENNIES_FOR_ZINGO_REGTEST_ADDRESS.to_string())));
     }
 
-    /// This tests checks that `messages_containing` returns an empty vector when empty memos are included.
+    /// This tests checks thatRegtestEnvironmentg` returns an empty vector when empty memos are included.
```
zancas: I am not sure what this is supposed to say.

zingo_wallet/src/interface.rs
zancas: So is this resolved?

zingo_wallet/src/interface.rs
zancas: I suppose we can fix this another day.

zancas:
I have three questions:

(1)  This PR seems to me to be a non-invasive addition that adds a stack that's mostly parallel to the legacy LightClient-fronted stack.  Is that correct?

(2)  The consumer of the zwi appears to convert between custom and foreign BlockHeights VIA **u32** doesn't this eliminate some of the benefits of static type checking?

(3)  The zwi package defines a custom BlockHeight.   Why not define a BlockLength trait and constrain consumers to implement it?   Then we could ALSO offer a default implementer behind a feature gate such that it's not part of the pub API for lazy consumers.

zancas (dismissed):
I believe that this is mergeable based on its non-invasive, additional, approach.  That having been said, I think this is the initiation of a significant upgrade to our core wallet offering so I would like @Oscar-Pepper and @dorianvp to review it, before it merges.

libtonode-tests/tests/concrete.rs
```diff
@@ -917,10 +917,10 @@ mod fast {
     }
 
     /// Tests that value transfers are properly sorted by block height and index.
-    /// It also tests that retrieving the value transfers multiple times in a row returns the same results.
+    /// It also tests that retRegtestEnvironmentansfers multiple times in a row returns the same results.
```
dorianvp:
```suggestion
    /// It also tests that retrieving the value transfers multiple times in a row returns the same results.
```

libtonode-tests/tests/concrete.rs
dorianvp:
```suggestion
    /// This tests checks that `messages_containing` returns an empty vector when empty memos are included.
```

zingo_wallet/src/interface.rs
```diff
@@ -0,0 +1,224 @@
...
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum GetMaxScannedHeightError {
+    #[error("Todo")]
```
dorianvp, line 40: I see a bunch of TODOs here. Non-blocking IMO

dorianvp (commented):
Partial review

libtonode-tests/tests/concrete.rs
fluidvanadium: Oops, a misuse of select-and-replace.

fluidvanadium (commented):
Issue 1: Every time I create a big PR into zingolib, I typically `sync the testnet wallet` and `run testnet tests`. This time, it resulted in an issue, that I am not sure exactly how to handle. I don't want to hide this error: https://github.com/zingolabs/zingolib/issues/2170 , so this PR cannot merge exactly as is.

Issue 2: Error API? I realized the only thing that I do with the error that the trait poops out is unwrap it in test. I think we need to add a shim to allow useful error-handling in a generic fashion.

Issue one blocks merge, (but pulL reQUest AuThorS cAN'T rEQuEsT ChANGeS on tHeIR oWn pULL ReqUEsts)

Cargo.toml
```diff
@@ -8,17 +8,26 @@ members = [
...
+zingo-status = "0.1.0"
+pepper-sync = { path = "pepper-sync" }
+zingolib = { path = "zingolib" }
+
+### Zingo-common
+zingo-netutils = "1.1.0"
+zingo_common_components = { git = "https://github.com/zingolabs/zingo-common.git", version = "0.2" }
 
-# Zcash
+### nuttycom
```
Oscar-Pepper, line 30: why change this name to "nuttycom"?

Oscar-Pepper (changes requested):

Cargo.toml
dorianvp: LOL

zancas:
Given the clean separation of this code into it's own package, I think it's safe to merge.

zancas (approved):

zancas:
Since this has a failing test, I am setting it back to draft for now.

PR #1989, opened 2025-10-14
Update to track zebra

zancas:


Oscar-Pepper:
i dont understand whats happening here. is your zebra `2.5.0-zl.1` tag still the 2.5.0 tag with our own changes? it seems all the new github workflow stuff is zebras work which means its not v2.5.0 anymore?

zancas:
I believe that zebra 3.0.0 is scheduled to land.

Since it's going to be in production prior to NU6.1 we need to update our dependence on it, prior to NU6.1.

By tagging and updating incrementally we move towards that state.  In other words this tag is between 2.5.0 and the imminent 3.0.0.

zancas:
It would be fine to update closer to _main_ or to 3.0.0 as soon as it lands, but that will require more work, than this minimal change.

fluidvanadium:
superceded

PR #1995, opened 2025-10-24
Remove functions from public interface that end in panic todo!().

fluidvanadium:
Better that the consumer gets a compiler error than a production panic.


zancas:
I updated this with latest dev.

PR #2006, opened 2025-11-12
Infra update

fluidvanadium:
Cargo update including patched infrastructure and infrastructure api change handling.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -64,9 +64,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.4"
+version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
+checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
 dependencies = [
  "memchr",
 ]
@@ -220,7 +220,7 @@ checksum = "4734bde002bb3d52e27ab808faa971a143d48d11dbd836d5c02edd1756cdab06"
 dependencies = [
  "async-trait",
  "cfg-if",
- "derive-deftly",
+ "derive-deftly 1.0.1",
  "derive_builder_fork_arti",
  "derive_more",
  "educe",
@@ -287,7 +287,7 @@ checksum = "3109e49b1e4909e9db6515a30c633684d68cdeaa252f215214cb4fa1a5bfee2c"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.109",
+ "syn 2.0.108",
  "synstructure",
 ]
 
@@ -299,14 +299,14 @@ checksum = "7b18050c2cd6fe86c3a76584ef5e0baf286d038cda203eb6223df2cc413565f7"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.109",
+ "syn 2.0.108",
 ]
 
 [[package]]
 name = "async-compression"
-version = "0.4.33"
+version = "0.4.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "93c1f86859c1af3d514fa19e8323147ff10ea98684e6c7b307912509f50e67b2"
+checksum = "5a89bce6054c720275ac2432fbba080a66a2106a44a1b804553930ca6909f4e0"
 dependencies = [
  "compression-codecs",
  "compression-core",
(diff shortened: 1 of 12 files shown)


Cargo.lock
```diff
@@ -8392,7 +8388,7 @@ dependencies = [
 [[package]]
 name = "zcash_local_net"
 version = "0.1.0"
-source = "git+https://github.com/zingolabs/infrastructure.git?branch=dev#8053b165d65f3f1c4255def86fcfd1ec28056744"
+source = "git+https://github.com/zingolabs/infrastructure.git?branch=dev#87a555d404c931ae6ce2b60aeadf5cc9d9e665ee"
```
AloeareV, line 8391: I have mentioned this off-channel, but I don't like that this is the only source of truth for what version of infrastructure we changed from and to. It's much easier to look at the Cargo.toml than the Cargo.lock for this sort of information, which a tag or rev dependency (instead of a branch) would give us. 

Non-blocking.

Cargo.toml
```diff
@@ -116,6 +116,9 @@ zingo-status = { path = "zingo-status" }
 pepper-sync = { path = "pepper-sync" }
 zingolib = { path = "zingolib" }
 
+# Test Attributes
+test-log = "0.2"
```
AloeareV, line 120: This is a great find!

libtonode-tests/Cargo.toml
```diff
@@ -41,3 +41,4 @@ http = { workspace = true }
 
 [dev-dependencies]
 rustls.workspace = true
+test-log.workspace = true
```
AloeareV, line 44: I thought we were moving away from test-only dependencies being specified at the workspace level? I argued in favour of doing things like this, so I don't especially mind...I seem to remember you convinced me to not do this, though.

zingolib/src/testutils/chain_generics/conduct_chain.rs
```diff
@@ -25,11 +25,11 @@ pub trait ConductChain {
     async fn create_faucet(&mut self) -> LightClient;
 
     /// sets server parameters
-    fn zingo_config(&mut self) -> crate::config::ZingoConfig;
+    async fn zingo_config(&mut self) -> crate::config::ZingoConfig;
```
AloeareV, line 30: It seems wrong that this function needs to be async. I understand it's required by the 'configured activation heights' function becoming async in infrastructure...I'd be interested in visiting the justification for that change. 

Non-blocking, as this is an infrastructure question and not even a zingolib one.

zingolib/src/wallet/send.rs
```diff
@@ -213,16 +212,21 @@ impl LightWallet {
...
                 .write(&mut transaction_bytes)
                 .map_err(|_| TransmissionError::TransactionWrite)?;
-            let transaction = Transaction::read(
-                transaction_bytes.as_slice(),
-                consensus::BranchId::for_height(&network, height),
-            )
-            .map_err(|_| TransmissionError::TransactionRead)?;
+
+            // this block serves to save the transactions to the wallet. This consensus branch is not baked into the sent transaction right here. because only transaction_bytes will be sent.
+            let transaction = Transaction::read(transaction_bytes.as_slice(), consensus_branch_id)
```
AloeareV, line 227: Was `consensus::BranchId::for_height` returning a branch ID that differed from the branch ID that was calculated for the transaction we're sending? Or is this simply removing an additional source of truth? I'm in favor of using the ID we already calculated instead of regenerating it...but if we were regenerating it wrong, that's interesting to me in its own right.

AloeareV (dismissed):
This is great! Some questions. I've approved as I don't think any of these are blocking. Feel free to merge once you've resolved the conversations!

Cargo.lock
fluidvanadium: You aren't wrong. I know I have edited the Cargo.lock file even though encouraged not to.

This is the most convenient way for me.

libtonode-tests/Cargo.toml
fluidvanadium: Yes, I hoped someone would point this out.

The rational for that policy is to prevent 'test-only code' from being built into production binaries. Since test-log can't build production code, there is no risk here. Not sure how to concisely encode that into an updated policy.

zingolib/src/testutils/chain_generics/conduct_chain.rs
fluidvanadium: The doc-comment is wrong. I just angrily annotated myself. Will fix right now.

zingolib/src/wallet/send.rs
fluidvanadium: We eliminated an extra source of truth.

AloeareV (approved):

PR #2009, opened 2025-11-24
Doubled nextest timeout.

fluidvanadium:
Problem:
Sometimes the tests go over the timeout but barely.

Solution:
Increase timeout by a small factor.


fluidvanadium:
I notice that this is a bandaid for the issue ultimately caused by this: https://github.com/zingolabs/zingolib/issues/2021

PR #2031, opened 2025-12-03
add critical error handling until missing checkpoint bug is fixed

Oscar-Pepper:
part 2 of missing checkpoint bug.

there will be a part 3 which actually fixes the underlying bug but while we are still debugging and testing to root cause I have added this temporary error handling to avoid crashes and unrecoverable wallet states


fluidvanadium (approved):
This looks right

Untested

I would like to see a trace logging the situation here.

fluidvanadium:
Added traces
https://github.com/Oscar-Pepper/zingolib/pull/25

dorianvp (approved):

PR #2061, opened 2025-12-14
feat: offline wallet creation & various fixes

dorianvp:


zingo-cli/src/lib.rs
```diff
@@ -335,29 +366,36 @@ If you don't remember the block height, you can pass '--birthday 0' to scan from
...
+        let indexer_uri_str = matches
             .get_one::<http::Uri>("server")
             .map(ToString::to_string);
+
         log::info!("data_dir: {}", &data_dir.to_str().unwrap());
-        let server = zingolib::config::construct_lightwalletd_uri(server);
+
+        let indexer_uri: Option<http::Uri> = match indexer_uri_str {
+            Some(s) => Some(zingolib::config::parse_indexer_uri(s)?),
+            None => None,
```
Oscar-Pepper, line 377: minor:
a map would do nicely here

zingo-cli/src/lib.rs
```diff
@@ -90,6 +89,38 @@ pub fn build_clap_app() -> clap::ArgMatches {
...
+}
+
+/// Parses a chain name (`"mainnet"`, `"testnet"`, or `"regtest"`) into a [`ChainType`].
+///
+/// Returns an error for unknown names.
+pub fn chain_from_str(chain_name: &str) -> Result<ChainType, ChainFromStringError> {
+    match chain_name {
+        "testnet" => Ok(ChainType::Testnet),
+        "mainnet" => Ok(ChainType::Mainnet),
+        // TODO: FIXME
```
Oscar-Pepper, line 107: fixme? also i see these NU heights more than once, maybe we could create a const to DRY

zingo-cli/src/lib.rs
```diff
@@ -377,9 +415,15 @@ pub type CommandRequest = (String, Vec<String>);
...
 
-/// Used by the zingocli crate, and the zingo-mobile application:
-/// <https://github.com/zingolabs/zingolib/tree/dev/cli>
-/// <https://github.com/zingolabs/zingo-mobile>
+/// Builds a `ZingoConfig` and initializes a `LightClient` from (in order):
+/// an explicit seed phrase, a UFVK, an existing wallet file, or fresh entropy.
+///
+/// For fresh wallets, the birthday is chosen as follows: if an indexer is configured,
+/// we query the chain tip and start at `tip - 100`. Otherwise we
+/// fall back to a conservative activation height and create the wallet offline.
```
Oscar-Pepper, line 423:
```suggestion
/// For fresh wallets, the birthday is chosen as follows: if an indexer is configured,
/// we query the chain height and start at `chain height - 100`. Otherwise we
/// fall back to a conservative activation height and create the wallet offline.
```

to stay in line with pepper-sync terminology

zingolib/Cargo.toml
```diff
@@ -88,7 +93,7 @@ secrecy = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
 tempfile = { workspace = true, optional = true }
-thiserror = { workspace = true }
+thiserror.workspace = true
```
Oscar-Pepper, line 96: i think that we should either leavev this or change all of them. consistency is the key thing here

zingolib/src/lib.rs
```diff
@@ -26,3 +26,5 @@ extern crate rust_embed;
 #[derive(RustEmbed)]
 #[folder = "zcash-params/"]
 pub struct SaplingParams;
+
+pub use zebra_chain::parameters::testnet::ConfiguredActivationHeights;
```
Oscar-Pepper, line 30: nice

zingo-cli/Cargo.toml
```diff
@@ -3,12 +3,8 @@ name = "zingo-cli"
 version = "0.2.0"
 edition = "2024"
 
-[features]
-regtest = [ "zingolib/regtest", "zingo_common_components/for_test" ]
-
 [dependencies]
 pepper-sync = { workspace = true }
-zingo_common_components = { workspace = true, optional = true }
 zingolib = { workspace = true }
```
Oscar-Pepper, line 8: if we are removing regtest feature then zingo-cli must build zingolib with the zingolib regtest feature active all the time

zingo-cli/src/commands.rs
```diff
@@ -596,7 +604,28 @@ impl Command for InfoCommand {
     }
 
     fn exec(&self, _args: &[&str], lightclient: &mut LightClient) -> String {
-        RT.block_on(async move { lightclient.do_info().await })
+        RT.block_on(async move {
+            match lightclient.server_uri() {
+                Some(uri) => match grpc_connector::get_info(uri.clone()).await {
```
Oscar-Pepper, line 609: see comment below on removing `do_info`. I think we want to call a lightclient method that passes through this call to `get_info` and restrict visibility of the grpc_connector, im not sure this should be part of zingolibs public API

zingolib/src/lightclient.rs
```diff
@@ -179,28 +179,6 @@ impl LightClient {
...
-                    "chain_name" => i.chain_name,
-                    "sapling_activation_height" => i.sapling_activation_height,
-                    "consensus_branch_id" => i.consensus_branch_id,
-                    "latest_block_height" => i.block_height
-                };
-                o.pretty(2)
-            }
-            Err(e) => e,
-        }
-    }
```
Oscar-Pepper, line 202: instead of deleting this can we rename it to `server_info` and just pass through the result from grpc_connector::get_info? I don't think we want the grpc_connector as part of the public API it would be better if consumers can call a lightcleint method to get this information IMO

Oscar-Pepper: i think removing the json string output type is correct, this should happen in zingo-cli as you have done above

zingolib/src/lightclient/error.rs
```diff
@@ -17,37 +17,52 @@ pub enum LightClientError {
...
     CalculateShieldError(CalculateTransactionError<Infallible>),
+
     #[error("Transmission failed. {0}")]
     TransmissionError(#[from] TransmissionError),
+
     #[error("No proposal found.")]
     NoStoredProposal,
+
+    #[error("Indexer is unreachable.")]
+    UnreachableIndexer,
```
Oscar-Pepper, line 65: see comment below. i think this should be removed.

zingolib/src/lightclient/send.rs
```diff
@@ -35,44 +35,59 @@ pub mod send_with_proposal {
...
-                .calculate_transactions(proposal, sending_account)
-                .await
-                .map_err(SendError::CalculateSendError)?;
-            self.latest_proposal = None;
-
-            Ok(wallet
-                .transmit_transactions(self.server_uri(), calculated_txids)
-                .await?)
+            match self.server_uri() {
+                None => Err(SendError::UnreachableIndexer),
```
Oscar-Pepper, line 39: i dont think this is necessary and bloats the errors tree. we should be able to construct and store a tx offline. it is transmission that requires a URI and you have already added the error to transmission errors which is passed back through SendError via the `from` impl

zingolib/src/lightclient/send.rs
```diff
@@ -35,44 +35,59 @@ pub mod send_with_proposal {
...
-            let calculated_txids = wallet
-                .calculate_transactions(proposal, shielding_account)
-                .await
-                .map_err(SendError::CalculateShieldError)?;
-            self.latest_proposal = None;
-
-            Ok(wallet
-                .transmit_transactions(self.server_uri(), calculated_txids)
-                .await?)
+            match self.server_uri() {
```
Oscar-Pepper, line 60: same here

zingolib/src/wallet/error.rs
```diff
@@ -213,32 +213,43 @@ impl From<bip32::Error> for KeyError {
...
         "Server error: txid reported by the server does not match calculated txid.\ncalculated txid:\n{0}\ntxid from server: {1}"
     )]
     IncorrectTxidFromServer(TxId, TxId),
+
     /// Failed to scan transmitted transaction..
     #[error("Failed to scan transmitted transaction. {0}")]
     SyncError(#[from] pepper_sync::error::SyncError<WalletError>),
+
+    #[error("Indexer is unreachable.")]
+    UnreachableIndexer,
```
Oscar-Pepper, line 252: this error suggests that the indexer is set but the connection is lost as oppose to an indexer not being set at all

zingolib/src/config.rs
```diff
@@ -102,50 +100,25 @@ impl Parameters for ChainType {
...
-/// TODO: Add Doc Comment Here!
+
+/// Default donation amount.
 pub const ZENNIES_FOR_ZINGO_AMOUNT: u64 = 1_000_000;
+
 /// The lightserver that handles blockchain requests
 pub const DEFAULT_LIGHTWALLETD_SERVER: &str = "https://zec.rocks:443";
-/// Used for testnet
+
+/// Default indexer uri for testnet.
```
Oscar-Pepper, line 115: minor:
I prefer this comment to the one above, could you replace `DEFAULT_LIGHTWALLETD_SERVER` comment with this one but use `mainnet` instead of `testnet`?

zingolib/src/config.rs
```diff
@@ -163,87 +136,90 @@ pub fn load_clientconfig(
...
-                } else {
-                    "http://".to_string() + &s
-                };
-                let uri: http::Uri = s.parse().unwrap();
-                if uri.port().is_none() {
-                    s += ":9067";
-                }
-                s
-            }
+// TODO: Is this really necessary?
```
Oscar-Pepper, line 183: I think defaulting to port 9067 is useful as this is the standard port. im not sure about adding http prefix or checking for empty now you have fixed the offfline logic

Oscar-Pepper: we should also handle the error instead of unwrap. maybe we can clean this up

zingolib/src/config.rs
```diff
@@ -163,87 +136,90 @@ pub fn load_clientconfig(
...
-    /// TODO: Add Doc Comment Here!
-    pub chain: ChainType,
-    /// TODO: Add Doc Comment Here!
+    /// Optional indexer endpoint. `None` means offline.
+    pub indexer_uri: Option<http::Uri>,
+
+    /// The network type.
+    pub chain_type: ChainType,
+
+    /// Extra blocks to rewind when choosing a sync start height.
```
Oscar-Pepper: reorg_buffer_offset is no longer in ZingoConfig

zingolib/src/config.rs
```diff
@@ -318,24 +273,24 @@ impl ZingoConfigBuilder {
...
 
-    /// TODO: Add Doc Comment Here!
+    /// Sets how many accounts the wallet should use.
     pub fn set_no_of_accounts(&mut self, no_of_accounts: NonZeroU32) -> &mut Self {
         self.no_of_accounts = no_of_accounts;
         self
     }
 
-    /// TODO: Add Doc Comment Here!
+    /// Builds a [`ZingoConfig`]. If no indexer is set, it's interpreted as offline.
```
Oscar-Pepper, line 285: isnt it offline when URI is none? the unwrap_or_default followed by Some(lightwalletd_uri) means that it cant be interpreted as offline right?

zingolib/src/config.rs
```diff
@@ -541,16 +518,24 @@ impl ZingoConfig {
...
+                self.indexer_uri
+                    .as_ref()
+                    .read()
+                    .unwrap()
+                    .clone()
+                    .expect("Couldn't read configured server URI!")
+                    .clone(),
+            );
+        }
+        None
```
Oscar-Pepper, line 531:
```suggestion
        self.indexer_uri.read().expect("Couldn't read configured server URI!").map(|uri| {
           uri.clone()
        })
```

Oscar-Pepper: untested but you get the idea

zingolib/src/config.rs
```diff
@@ -644,57 +624,25 @@ mod tests {
...
         rustls::crypto::ring::default_provider()
             .install_default()
             .expect("Ring to work as a default");
         tracing_subscriber::fmt().init();
 
-        let valid_uri = crate::config::construct_lightwalletd_uri(Some(
+        let valid_uri = crate::config::parse_indexer_uri(
             crate::config::DEFAULT_LIGHTWALLETD_SERVER.to_string(),
-        ));
+        )
```
Oscar-Pepper, line 632: it seems you have removed a load config test that should contain this URI (and incorrectly tested serverless previously) but now have added a serverless test which creates a config *wtih* a server?

can we call this test `load_clientconfig` and duplcate it and call it `load_clientconfig_serverless` with a `None` in the URI field and change the assertions accordingly?

Oscar-Pepper (changes requested):
great work

i havent looked at how a zingo-cli user should set the server to offline mode. i also havent looked thouroughly at how zingolib consumers such as zingo-mobile would set to serverless but my assumption is they can directly create a config with URI set to None. i will look into these things on my final review

Oscar-Pepper:
oh also the README will need updating if you ahve removed the regest feature flag from zingo-cli

zingo-cli/src/lib.rs
dorianvp: Yeah. I was thinking that these activation heights shouldn't be hard-coded. Maybe the desired regtest has a different set of activation heights.

One way is to get them from the node itself, though that will require some work on zingo-cli. Either that or allowing the user to define a config, but I'd like to avoid creating a file just for that parameter.

zingo-cli/Cargo.toml
dorianvp: IMO this is fine. I'm specifically thinking about distributing binaries.

zingolib/src/lightclient.rs
dorianvp: Okay, I'll do that.

zingolib/src/config.rs
dorianvp: I'll delete the field then.

zingolib/src/config.rs
dorianvp: You're right, this was a mistake from before! I'll update this.

zancas:
Superceded?

zancas:
Superseded: offline wallet creation and the surrounding behavior shipped through the Offline-mode arc (`--offline`, the Indexerless state, the Library Birthday for offline NewSeed creation — ADRs 0001, 0007, 0008).

PR #2072, opened 2026-01-01
fix: respect coinbase maturity in balance calculations (#2069)

### LINKED ISSUE #2069: Balance does not take into account coinbase maturity
This is particularly visible when attempting to perform a quickshield with the miner's spending key.

It currently shows it as confirmed transparent balance.

Timi16:
## Overview

The wallet previously displayed **transparent coinbase outputs** as spendable **before** they reached the required **100-block maturity period**.  
This caused confusion, especially when performing a **quickshield with a miner’s spending key**, where the wallet showed confirmed funds that could not actually be spent.

This change fixes the balance calculation logic to correctly enforce **coinbase maturity rules**, ensuring the displayed balance matches what is truly spendable.

---

## Problem

Transparent coinbase outputs must mature for **100 blocks** before they can be spent.  
However, the wallet balance UI included these outputs as soon as they were confirmed.

### Symptoms
- Confirmed transparent balance appears higher than spendable balance
- Quickshield fails despite sufficient “confirmed” funds
- Mismatch between balance display and spending behavior

---

## Root Cause

The codebase used **two different logic paths** for transparent outputs:

### Spending Path (Correct )
- `spendable_transparent_coins()` in `output.rs`
- Properly enforces the 100-block coinbase maturity rule

### Balance Calculation Path (Incorrect )
- `confirmed_balance()` and related functions in `balance.rs`
- Only checks `is_confirmed()`
- **Does not check coinbase maturity**

Because of this, immature coinbase outputs were incorrectly included in the confirmed balance.

---

## Solution

Coinbase maturity checks have been added to the balance calculation logic.

### What Changed
- Added a `COINBASE_MATURITY` constant set to **100 blocks**
- Updated:
  - `confirmed_balance()`
  - `confirmed_balance_excluding_dust()`
- Transparent outputs originating from coinbase transactions are now:
  - Included **only if** they have at least **100 confirmations**
- Shielded coinbase outputs (Sapling / Orchard) remain unaffected, as they do not have a 100-block maturity requirement

---

## Files Modified

```text
zingolib/src/wallet/balance.rs
```

---

## Result

- Wallet balances now accurately reflect **spendable transparent funds**
- Prevents premature inclusion of immature coinbase outputs
- Aligns balance display logic with consensus rules and the spending path
- Improves reliability of miner and quickshield workflows

---

## Notes

This change affects **balance display only**.  
Spending logic already correctly enforced coinbase maturity.

### DIFF
--- a/libtonode-tests/tests/concrete.rs
+++ b/libtonode-tests/tests/concrete.rs
@@ -4330,7 +4330,6 @@ async fn mine_to_transparent_coinbase_maturity() {
     check_client_balances!(faucet, o: 0 s: 0 t: 1_875_000_000);
     
     // Balance should be 0 because coinbase needs 100 confirmations
-    // TODO: After fix, this should be 0
     assert_eq!(
         faucet
             .wallet
--- a/libtonode-tests/tests/concrete.rs
+++ b/libtonode-tests/tests/concrete.rs
@@ -6,6 +6,8 @@ use json::JsonValue;
 use zcash_address::unified::Fvk;
 use zcash_primitives::transaction::fees::zip317::MINIMUM_FEE;
 
+use pepper_sync::wallet::TransparentCoin;
+use zcash_protocol::PoolType;
 use zcash_protocol::value::Zatoshis;
 use zebra_chain::parameters::testnet;
 use zingo_test_vectors::{BASE_HEIGHT, block_rewards, seeds::HOSPITAL_MUSEUM_SEED};
@@ -16,8 +18,6 @@ use zingolib::wallet::keys::unified::UnifiedKeyStore;
 use zingolib::wallet::summary::data::{CoinSummary, NoteSummary};
 use zingolib::{check_client_balances, get_base_address_macro};
 use zingolib_testutils::scenarios::{self, increase_height_and_wait_for_client};
-use pepper_sync::wallet::TransparentCoin;
-use zcash_protocol::PoolType;
 
 fn check_expected_balance_with_fvks(
     fvks: &Vec<&Fvk>,
@@ -4319,16 +4319,13 @@ mod basic_transactions {
 
 #[tokio::test]
 async fn mine_to_transparent_coinbase_maturity() {
-    let (local_net, mut faucet, _recipient) = scenarios::faucet_recipient(
-        PoolType::Transparent,
-        for_test::all_height_one_nus(),
-        None,
-    )
-    .await;
-    
+    let (local_net, mut faucet, _recipient) =
+        scenarios::faucet_recipient(PoolType::Transparent, for_test::all_height_one_nus(), None)
+            .await;
+
     // After 3 blocks (BASE_HEIGHT), faucet has mined transparent coinbase
     check_client_balances!(faucet, o: 0 s: 0 t: 1_875_000_000);
-    
+
     // Balance should be 0 because coinbase needs 100 confirmations
     assert_eq!(
         faucet
@@ -4340,12 +4337,12 @@ async fn mine_to_transparent_coinbase_maturity() {
             .into_u64(),
         0
     );
-    
+
     // Mine 100 more blocks to mature the coinbase
     increase_height_and_wait_for_client(&local_net, &mut faucet, 100)
         .await
         .unwrap();
-    
+
     // Now balance should include the mature coinbase
     // (Plus new coinbase rewards from the 100 blocks)
     let mature_balance = faucet
@@ -4355,11 +4352,10 @@ async fn mine_to_transparent_coinbase_maturity() {
         .confirmed_balance_excluding_dust::<TransparentCoin>(zip32::AccountId::ZERO)
         .unwrap()
         .into_u64();
-    
+
     // Should have original coinbase + 100 new blocks worth
     assert_eq!(mature_balance, 1_875_000_000);
 }
-
 
 // FIXME: does not assert dust was included in the proposal
 #[tokio::test]
--- a/libtonode-tests/tests/concrete.rs
+++ b/libtonode-tests/tests/concrete.rs
@@ -4339,7 +4339,7 @@ async fn mine_to_transparent_coinbase_maturity() {
             .confirmed_balance_excluding_dust::<TransparentCoin>(zip32::AccountId::ZERO)
             .unwrap()
             .into_u64(),
-        1_875_000_000 // FIXME: Should be 0 before fix!
+        0
     );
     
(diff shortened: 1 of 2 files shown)


Timi16:
https://github.com/zingolabs/zingolib/issues/2069 Here's the issues just incase

pacu (dismissed):
Thanks for your contribution!

I think that should:
1. refactor the repeated code
2. add tests for the logic you are exercising 
3. avoid the "return true" and use structured code that's error prone
4. the COINBASE_MATURITY constant should probably be obtained from the librustzcash crates (I'm not sure if it's defined, but if it is, then that should be used instead of redefining it)

Timi16:
already on the changes

Timi16:
I've added the fix. For the tests, would you prefer:

Unit tests
Integration tests with regtest
Skeleton tests to complete in follow-up PR

What's the team's preference?

dorianvp (changes requested):
I am getting the following compilation errors:

```bash
error[E0308]: mismatched types
   --> zingolib/src/wallet/balance.rs:166:63
    |
166 |             let confirmations = current_height.saturating_sub(tx_height);
    |                                                -------------- ^^^^^^^^^ expected `u32`, found `BlockHeight`
    |                                                |
    |                                                arguments to this method are incorrect
    |
note: method defined here
   --> /home/dorian/.cargo/git/checkouts/librustzcash-c2c6180a1a11d4fa/3ba772c/components/zcash_protocol/src/consensus.rs:34:12
    |
 34 |     pub fn saturating_sub(self, v: u32) -> BlockHeight {
    |            ^^^^^^^^^^^^^^
help: call `Into::into` on this expression to convert `BlockHeight` into `u32`
    |
166 |             let confirmations = current_height.saturating_sub(tx_height.into());
    |                                                                        +++++++

error[E0308]: mismatched types
   --> zingolib/src/wallet/balance.rs:168:30
    |
168 |             confirmations >= COINBASE_MATURITY
    |             -------------    ^^^^^^^^^^^^^^^^^ expected `BlockHeight`, found `u32`
    |             |
    |             expected because this is `BlockHeight`
    |
help: call `Into::into` on this expression to convert `u32` into `BlockHeight`
    |
168 |             confirmations >= COINBASE_MATURITY.into()
    |                                               +++++++
```

Regarding tests, maybe a simple one that compares balances before & after reaching blockheight would be enough? Take a look at the currently existing ones under `libtonode-tests/tests/concrete.rs`, in particular, `mine_to_transparent`...

Timi16:
Okay thanks would check it out

Timi16:
OKay added tests and fix took a while but hope this should do the trick

Timi16:
Hey @dorianvp can you check it out

libtonode-tests/tests/concrete.rs
```diff
@@ -4313,6 +4315,53 @@ mod basic_transactions {
...
+        for_test::all_height_one_nus(),
+        None,
+    )
+    .await;
+    
+    // After 3 blocks (BASE_HEIGHT), faucet has mined transparent coinbase
+    check_client_balances!(faucet, o: 0 s: 0 t: 1_875_000_000);
+    
+    // Balance should be 0 because coinbase needs 100 confirmations
+    // (Currently showing as confirmed due to bug)
```
dorianvp:
```suggestion
```

libtonode-tests/tests/concrete.rs
```diff
@@ -4313,6 +4315,53 @@ mod basic_transactions {
...
+    // TODO: After fix, this should be 0
```
dorianvp:
```suggestion
```

libtonode-tests/tests/concrete.rs
```diff
@@ -4313,6 +4315,53 @@ mod basic_transactions {
...
+    assert_eq!(
+        faucet
+            .wallet
+            .read()
+            .await
+            .confirmed_balance_excluding_dust::<TransparentCoin>(zip32::AccountId::ZERO)
+            .unwrap()
+            .into_u64(),
+        1_875_000_000 // FIXME: Should be 0 before fix!
```
dorianvp:
```suggestion
        0
```

libtonode-tests/tests/concrete.rs
```diff
@@ -4313,6 +4315,53 @@ mod basic_transactions {
...
+    let mature_balance = faucet
+        .wallet
+        .read()
+        .await
+        .confirmed_balance_excluding_dust::<TransparentCoin>(zip32::AccountId::ZERO)
+        .unwrap()
+        .into_u64();
+    
+    // Should have original coinbase + 100 new blocks worth
+    assert!(mature_balance > 1_875_000_000);
```
dorianvp:
```suggestion
    assert_eq!(mature_balance, 1_875_000_000);
```

zingolib/src/wallet/balance.rs
```diff
@@ -130,6 +135,49 @@ fn format_zatoshis(zatoshis: Zatoshis) -> String {
...
+            if current_height_u32 < tx_height_u32 {
+                return false;
+            }
+            
+            let confirmations = current_height_u32 - tx_height_u32;
+            confirmations >= COINBASE_MATURITY
+        } else {
+            true
+        }
+    } // <-- This closing brace was missing
```
dorianvp:
```suggestion
    }
```

zingolib/src/wallet/balance.rs
```diff
@@ -496,6 +549,25 @@ impl LightWallet {
 
 #[cfg(any(test, feature = "testutils"))]
 mod test {
+    use super::*;
+    
+    // TODO: Add unit tests for coinbase maturity
```
dorianvp, line 550: Remove these comments.

dorianvp (dismissed):
ACK with suggestions. Also, I would run `clippy` and `cargo-fmt` so that CI does not complain. Good job @Timi16!!

Timi16:
Alright thanks @dorianvp  would change this comments

zingolib/src/wallet/balance.rs
dorianvp: I'll do this in a follow-up PR

Timi16:
Alright @dorianvp thank you made this changes

dorianvp (dismissed):

zingolib/src/wallet/balance.rs
Timi16: Alright Thank you , Is this okay to merge?

dorianvp (dismissed):

dorianvp (approved):

PR #2079, opened 2026-01-09
fix: add timeouts for network requests

dorianvp:


juanky201271 (approved):
I tested in zingo-cli & zingo-mobile ans it is working as expected, using sa.zec.rocks:443 while syncing you can see the time-out error preventing the App is blocked.

PR #2080, opened 2026-01-10
Fix memo parsing for strings starting with 0x

### LINKED ISSUE #2075: adding a `0x..` string to memo  in `send` results in tx failure
When adding a string starting with `0x..` as the memo to a `send` command and then entering `confirm`, the CLI fails with the error message : 

```json
{
  "error": "The sending transaction could not be calculated. Failed to calculate transaction. The underlying datasource produced the following error: Failed to scan calculated transaction."
}
```

### Steps to Reproduce: 

1. Launch zingo-cli
2. Use the send command: `send <recipient_address> <amount> <memo>` where `<memo>` is set as a string that starts with 0x (such as an ETH address)
3. Enter `confirm` to proceed with the transaction
4. Observe the CLI fails with the above error message

### Version details

```bash
$ zingo-cli version
2.0.9-91-g29d621f
```

Timi16:
When a memo string starts with `0x` (like an Ethereum address), the current code attempts to decode it as hexadecimal. For valid hex strings like Ethereum addresses, the decode succeeds but produces data that causes transaction failures during scanning.

## Changes
- Modified `interpret_memo_string` in `zingolib/src/wallet/utils.rs` to only use hex-decoded data if it's a valid memo size (≤512 bytes)
- Added comprehensive unit tests covering:
  - Plain text memos
  - Valid hex strings
  - Ethereum addresses (valid hex but should be treated as text)
  - Invalid hex strings
  - Oversized memos

## Testing
Added unit tests in `zingolib/src/wallet/utils.rs`:
- `test_interpret_memo_string_plain_text`
- `test_interpret_memo_string_valid_hex`
- `test_interpret_memo_string_ethereum_address`
- `test_interpret_memo_string_invalid_hex`
- `test_interpret_memo_string_too_long`


pepper-sync/src/scan/transactions.rs
```diff
@@ -457,13 +457,36 @@ fn parse_encoded_memos<N, Nf: Copy>(wallet_notes: &[WalletNote<N, Nf>]) -> Vec<P
...
+                    match zingo_memo::parse_zingo_memo(bytes_512) {
+                        Ok(encoded_memo) => return Some(encoded_memo),
+                        Err(e) => {
+                            tracing::error!("Failed to parse new format memo: {e}");
+                            return None;
+                        }
                     }
                 }
+
+                // Fall back to legacy format (511 bytes without 0xFF)
```
Oscar-Pepper: I think there was some confusion here, the memo has already been parsed to zip302 compliancy here https://github.com/zingolabs/zingolib/blob/dev/pepper-sync/src/scan/transactions.rs#L390. when the Memo is of `Arbitrary` variant, the first of the 512 bytes was 0xFF. Therefore, the number of bytes will always be 511 here and will not be a memo intended for the recipient to read as a human-readable string. I will put some information for what my assumptions are for the problem and a potential solution in the review comment

Oscar-Pepper (changes requested):
Hi @Timi16 thanks for looking into this. I will post in the issue https://github.com/zingolabs/zingolib/issues/2075 to get more info as i'm struggling to understand under which conditions this bug occurs, i cant reproduce it myself. Please see this issue for more context. The changes you have made are to the parsing of an `Arbitrary` memo - starting with 0xFF byte as stated in zip302 - which are used internally by zingo and are not displayed to the user or intended to be human-readable. Therefore, this code needs to remain unchanged. I assume the issue we are trying to solve is when the ethereum address is inputted by the user as text. In this case, the first byte will be <= 0xF4 resulting is a memo of `Text` variant where the remaining 511 bytes will be parsed to a string. This string *should* be able to contain any valid UTF-8 and its not clear to me why any text memo starting with "0x" will be treated differently. my tests support this, here is an example:
```
    txid: cfd5f064237370c7d60131bdbe86a4950d206d995c06991f415a2f6bfbdd485f
    datetime: 2026-01-14 06:31:03 UTC
    status: confirmed
    blockheight: 3204220
    kind: send-to-self
    value: 0
    fee: 10000
    zec price: not available
    orchard notes:
        {
            value: 100000
            spend status: unspent
            output index: 0
            memo: 0xtest
        }
```

pepper-sync/src/scan/transactions.rs
Timi16: Hi @Oscar-Pepper,
Thank you for the clarification! I completely misunderstood the issue.
You're right, I was fixing the scanner (pepper-sync) which handles already-stored memos from the blockchain, when the actual bug is in how user input is processed before the transaction is even created.
The issue is in interpret_memo_string in zingolib/src/wallet/utils.rs where user text input is being hex-decoded when it starts with "0x".
I'll Revert all the changes to zingo-memo and pepper-sync
Focus only on fixing interpret_memo_string to treat all input as plain UTF-8 text (remove the hex-decoding logic)

This makes much more sense - users should be able to type "0x742d35Cc..." as a text memo without it being treated as hexadecimal data.
Let me update the PR with the correct fix. Thanks for your patience! 🙏

Timi16: Should I close this PR and start fresh, or just revert the commits and push the correct fix @Oscar-Pepper?

Oscar-Pepper: Ah i see, I can reproduce the error now as i wasnt entering valid hex after the "0x". Ok i will discus with the team to completely remove this hex decoding. It may be cleaner to close this PR and create a new one but its up to you. I think we will want to just remove the `intepret_memo_string` fn and replace all calls to it with `MemoBytes::from_bytes(&Vec::from(memo_str.as_bytes()))
        .map_err(|_| format!("Error creating output. Memo '{memo_str:?}' is too long"))`

Timi16: If you don't mind could i make the changes recquested or close the PR and leave for the team to handle?

Oscar-Pepper: oh sorry i see you have already made the changes i will review now

Oscar-Pepper (approved):
great work. thanks for your contribution!

Oscar-Pepper:
hi @Timi16 I have pushed a commit instead of requesting changes. I hope this is ok, otherwise I can PR into your branch if you prefer. please review my commit and if all ok we can merge.

Timi16:
> hi @Timi16 I have pushed a commit instead of requesting changes. I hope this is ok, otherwise I can PR into your branch if you prefer. please review my commit and if all ok we can merge.

Yes you can all is Okay thank you

Timi16:
> great work. thanks for your contribution!

I Thank the Zingolib team for there support

Oscar-Pepper (approved):

Issue #2088, opened 2026-01-15
Build.rs fails with "Permission denied" on macOS during param download`


Timi16:
**Description:**

The zingolib build script (`build.rs`) fails on macOS with a "Permission denied (os error 13)" error when trying to download Zcash parameters, even when the parameters already exist in `~/.zcash-params/`.

**Environment:**
- OS: macOS (MacBook Air)
- Rust version: 1.85
- zingolib branch: `dev` (commit `6a08f94f`)

**Steps to reproduce:**
1. Add zingolib as dependency: `zingolib = { git = "https://github.com/zingolabs/zingolib", branch = "dev", features = ["regtest"] }`
2. Run `cargo build --release`

**Error output:**
```
error: failed to run custom build command for `zingolib v2.1.2`
--- stdout
Checking if params are available...
Error downloading params: Permission denied (os error 13)
--- stderr
thread 'main' panicked at zingolib/build.rs:51:13:
explicit panic

Timi16:
Fix to this issues https://github.com/zingolabs/zingolib/pull/2089

dorianvp:
Hey, I tried reproducing the issue but was not able to. Here's the setup I used: https://github.com/dorianvp/zl-repro.

Did this issue happen under a container environment? And also, do you know which user/program created or modified the `~/.zcash-params` folder?

Edit: I see you mentioned `~/.zcash-params`, but on macOS that is not where params are stored.

Timi16:
Hmmm try this and cd to zeckit-faucet 
https://github.com/Timi16/ZecKit/tree/81d8cfdb05fb53f08fce89b770add0c6b98fb56a

dorianvp:
I get this output:

```bash
~/Desktop/ZecKit/zeckit-faucet
81d8cfd $ cargo build --release
warning: unused manifest key: profile.release.http
warning: unused manifest key: profile.release.zcash_primitives
   Compiling proc-macro2 v1.0.105
   Compiling quote v1.0.43
   Compiling unicode-ident v1.0.22
   Compiling libc v0.2.180
   Compiling cfg-if v1.0.4
   Compiling typenum v1.19.0
   Compiling version_check v0.9.5
   Compiling serde_core v1.0.228
   Compiling find-msvc-tools v0.1.8
   Compiling shlex v1.3.0
   Compiling subtle v2.6.1
   Compiling serde v1.0.228
   Compiling memchr v2.7.6
   Compiling autocfg v1.5.0
[...]

PR #2089, opened 2026-01-15
Fix build.rs permission errors on macOS

Timi16:
- Check if params exist locally before downloading
- Gracefully handle download failures
- Fall back to ~/.zcash-params if download fails
- Add better error messages


fluidvanadium (dismissed):
Read the file operations. They are correct/safe.

Timi16:
> Read the file operations. They are correct/safe.

Alright thank you @fluidvanadium would await a merge

dorianvp:
https://github.com/zingolabs/zingolib/issues/2088#issuecomment-3763931616.

dorianvp (changes requested):
I see that in this PR you're re-implementing a bunch of stuff that `zcash_proofs` already does internally, such as looking for the default dir and checking permissions.

What I see that is different is that, by skipping the call to `download_sapling_parameters`, you're also bypassing the integrity check. In other words, this PR introduces a regression in integrity checking.

Timi16:
You're absolutely right about the integrity check regression. That was my mistake.
The root issue is that zcash_proofs::download_sapling_parameters() is failing with "Permission denied (os error 13)" even when:

The params exist at ~/.zcash-params/
The params are readable (ls -la shows correct ownership)
The directory is accessible

Timi16:
The question is: What file operation in zcash_proofs::download_sapling_parameters() requires write permissions that's causing this to fail on macOS?
If we can identify that, we could:

Submit a fix to zcash_proofs to make that operation more permissive
Or handle it gracefully in zingolib's build script

Could you help identify what specific operation in zcash_proofs needs write access during verification?

Timi16:
Looks like i am facing an issues here let me solve it

pacu (changes requested):
This PR is actually doing more than it states.

it should only address the permission errors (if applicable). Although I'm not sure that the build.rs should account for writing permissions on a platform only. If it does it should do it in all of them or none of them given that's a a problem that might have many sources. (low disk space for example). What's your opinion on this @dorianvp ?

dorianvp:
I think this PR needs to be split into one that deals with the regtest stuff and another separate one related to the permissions error. I've yet to see a repro for the permissions issue, I'm wondering whether it's an issue originating from `zcash_proofs`, because the error message seems to be pointing somewhere else...

Timi16:
> I think this PR needs to be split into one that deals with the regtest stuff and another separate one related to the permissions error. I've yet to see a repro for the permissions issue, I'm wondering whether it's an issue originating from `zcash_proofs`, because the error message seems to be pointing somewhere else...

Would create a repo and screenshot if that works?

dorianvp:
> > I think this PR needs to be split into one that deals with the regtest stuff and another separate one related to the permissions error. I've yet to see a repro for the permissions issue, I'm wondering whether it's an issue originating from `zcash_proofs`, because the error message seems to be pointing somewhere else...
> 
> Would create a repo and screenshot if that works?

Sure. The idea is to reproduce this with the same code path, but manually with `zcash_proofs`.

Timi16:
> > > I think this PR needs to be split into one that deals with the regtest stuff and another separate one related to the permissions error. I've yet to see a repro for the permissions issue, I'm wondering whether it's an issue originating from `zcash_proofs`, because the error message seems to be pointing somewhere else...
> > 
> > 
> > Would create a repo and screenshot if that works?
> 
> Sure. The idea is to reproduce this with the same code path, but manually with `zcash_proofs`.

Alright lets try this

Timi16:
@dorianvp other changes removed the Regtest stuff https://github.com/zingolabs/zingolib/pull/2147

Timi16:
> @dorianvp other changes removed the Regtest stuff #2147

Alright any update on this @dorianvp

dorianvp:
This PR needs a repro, as requested above.

Timi16:
> This PR needs a repro, as requested above.

Would create a repo soon

zancas:
I am setting this to draft until we can reproduce.

zancas:
Closing until repro.

Timi16:
Hey @zancas

Timi16:
Would provide a repo for this shortly

Timi16:
It kinda of works though don't know if it was the permission change that made it to https://github.com/Timi16/Zl-repo

PR #2090, opened 2026-01-16
fix large wallet file size due to excessive nullifier and outpoint maps in some cases

Oscar-Pepper:
- sync no longer maps outpoints when nullifier map limit is reached
- sync no longer maps nullifiers and outpoints for chain tip on low memory settings
- each batch is now checked to ensure max nullifier map size limit is never exceeded


Oscar-Pepper:
closed for #2091

Issue #2087, opened 2026-01-15
large file size due to large blockrange difference between latest sapling and orchard incomplete shards


Oscar-Pepper:
we store the nullifiers of the whole chain tip, where chain tip means the sapling or orchard latest incomplete shard (whichever is bigger). if there are not many sapling outputs on chain then it takes a long time to complete the the shard which means it could be storing many many orchard shards worth of nullifiers in some cases.

current latest sapling and orchard complete shard block ranges (current chain height: 3205458):

sapling:
    BlockHeight(
        2769539,
    )..BlockHeight(
        3038259,
    ),

orchard:
    BlockHeight(
        3015712,
    )..BlockHeight(
        3067272,
    ),
    BlockHeight(
        3067271,
    )..BlockHeight(
        3096347,
    ),
    BlockHeight(
        3096346,
    )..BlockHeight(
        3115356,
    ),
    BlockHeight(
        3115355,
    )..BlockHeight(
        3128376,
    ),
    BlockHeight(
        3128375,
    )..BlockHeight(
        3142650,
    ),
    BlockHeight(
        3142649,
    )..BlockHeight(
        3165434,
    ),
    BlockHeight(
        3165433,
    )..BlockHeight(
        3192320,
    ),

as can be seen from these ranges, there are 6 completed orchard shards within the latest incomplete sapling shard range. this means even on the lowest performance settings, up to 7 shards of nullifiers will be stored in memory in the case of the current chain.

PR #2091, opened 2026-01-16
fix large wallet file size due to excessive nullifier and outpoint maps in some cases

Oscar-Pepper:
- sync only maps outpoints of targetted ranges (found note or higher priority) due to all transparent spend locations being known before scanning begins
- sync no longer maps nullifiers for chain tip on low memory settings
- each batch is now checked to ensure max nullifier map size limit is never exceeded

currently, the nullifiers collected when scanning a scan range are not added to the wallets nullifier map if the map exceeded a specified limit **before** the range started scanning. furthermore, the `map_nullifiers` bool applies to the entire scan range (usually a "shard range") as this is decided before the batcher splits these larger ranges into batches of fixed output size. as described in issue #2087, a single orchard shard range (or incomplete shard range) could contain several sapling shards which means a potentially huge number of nullifiers which need to be stored in memory and in the wallet file. combining all this with the sync engine always mapping nullifiers for the chain tip range (for improved spendable balance) and in some cases the memory usage and wallet file can become excessively large due to mapping a huge number of nullifiers. this is further exasperated by the fact that we currently *always* map outpoints even when `map_nullifiers` is false.

This PR fixes these issues firstly by introducing an `nullifier_map_limit_exceeded` bool which is assumed `false` when sync begins. This replaces the `map_nullifiers` bool for scan range selection during pre-scanning. This `nullifier_map_limit_exceeded` bool is set `true` in post-scan processing if the nullifiers in the scanned batch will cause the wallets nullifier map to exceed the maximum specified size, if added. the nullifiers in this batch and any future batches will no longer be added to the wallets nullifier map, only checked for spends and discarded. just as before, these ranges are marked `ScannedWithoutMapping` priority and the nullifiers will be re-fetched later in the sync process to ensure spends are not missed for newly added outputs.

Secondly, outpoints are only mapped for targetted ranges (found note or higher) as all transparent spend locations are known before scanning begins. `update_transparent_spends` has been improved to allow additional maps to be checked for spends in the case that the outpoints are not being added to the wallet main outpoint map, just like `update_shielded_spends`.

Thirdly, the chain tip is no longer mapped. This prevents the issue described above at the cost of improving the likelyhood of spendable notes early on in the sync process. Early note spendability will be improved in a following PR to account for this.

Finally, this PR also fixes a bug where, in rare edges cases, `ScannedWithoutMapping` scans which re-fetch nullifiers could complete before lower ranges have completed scanning, potentially missing spends. This PR improves scan range seelction to decrease the chance of this occuring and also adds logic to discard these nullifiers and reset the scan range to `ScannedWithoutMapping` again to be re-fetched safely. This resulted in an improvement to `punch_scan_priority` to allow a larger range of `Scanning` priority to be "punched in" with the potentially smaller `ScannedWithoutMapping` range in the case of batching when the initial scan range exceeded the maximum nullifier batch size.

Due to the improved logic for checking the nullifier map size limit is not exceeded, the `Medium` performance level max nullifier map size limit has been increased from 0 to 125_000 to allow an approx. 4MB nullifier map (125_000 x 32 bytes). this is much smaller than the 40MB we have observed in the case of the current chain (for medium and even low performance settings) but enough to allow some early spendable note benefits

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9222,7 +9222,7 @@ dependencies = [
 
 [[package]]
 name = "zingolib"
-version = "2.1.0"
+version = "2.1.2"
 dependencies = [
  "append-only-vec",
  "bech32",
--- a/pepper-sync/src/sync.rs
+++ b/pepper-sync/src/sync.rs
@@ -23,7 +23,7 @@ use zcash_protocol::consensus::{self, BlockHeight};
 use zingo_status::confirmation_status::ConfirmationStatus;
 
 use crate::client::{self, FetchRequest};
-use crate::config::SyncConfig;
+use crate::config::{PerformanceLevel, SyncConfig};
 use crate::error::{
     ContinuityError, MempoolError, ScanError, ServerError, SyncError, SyncModeError,
     SyncStatusError,
@@ -456,6 +456,7 @@ where
                     scan_range,
                     scan_results,
                     initial_verification_height,
+                    config.performance_level,
                 )
                 .await?;
                 wallet_guard.set_save_flag().map_err(SyncError::WalletError)?;
@@ -848,6 +849,7 @@ where
 }
 
 /// Scan post-processing
+#[allow(clippy::too_many_arguments)]
 async fn process_scan_results<W>(
     consensus_parameters: &impl consensus::Parameters,
     wallet: &mut W,
@@ -856,6 +858,7 @@ async fn process_scan_results<W>(
     scan_range: ScanRange,
     scan_results: Result<ScanResults, ScanError>,
     initial_verification_height: BlockHeight,
+    performance_level: PerformanceLevel,
 ) -> Result<(), SyncError<W::Error>>
 where
     W: SyncWallet
@@ -870,15 +873,17 @@ where
         Ok(results) => {
             let ScanResults {
                 mut nullifiers,
-                outpoints,
+                mut outpoints,
                 scanned_blocks,
                 wallet_transactions,
                 sapling_located_trees,
                 orchard_located_trees,
-                map_nullifiers,
+                mut map_nullifiers,
             } = results;
 
             if scan_range.priority() == ScanPriority::ScannedWithoutMapping {
+                spend::update_transparent_spends(wallet, Some(&mut outpoints))
+                    .map_err(SyncError::WalletError)?;
                 spend::update_shielded_spends(
                     consensus_parameters,
                     wallet,
@@ -925,6 +930,20 @@ where
                     true,
                 );
             } else {
+                // nullifiers and outpoints are not mapped if nullifier map size limit will be exceeded
+                let nullifier_map = wallet.get_nullifiers().map_err(SyncError::WalletError)?;
+                if map_nullifiers
+                    && max_nullifier_map_size(performance_level).is_some_and(|max| {
+                        nullifier_map.orchard.len()
+                            + nullifier_map.sapling.len()
+                            + nullifiers.orchard.len()
+                            + nullifiers.sapling.len()
+                            > max
+                    })
+                {
+                    map_nullifiers = false;
+                }
+
                 update_wallet_data(
                     consensus_parameters,
                     wallet,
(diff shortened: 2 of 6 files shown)


Oscar-Pepper:
fixes #2087

Cargo.lock
```diff
@@ -9222,7 +9222,7 @@ dependencies = [
 
 [[package]]
 name = "zingolib"
-version = "2.1.0"
+version = "2.1.2"
```
zancas, line 9225: Is there a CHANGELOG that documents all pub (if any) interface changes due to the migration from 2.1.0 to 2.1.2.

From versioning I don't expect any BREAKING changes, but is it possible that there are new pub APIs for a x.y.0 --> x.y.2 change?

Oscar-Pepper: The version has already been updated and has a release tag. Unfortunately the cargo lock was not checked in so this is just an automatic change to the lock when you build the 2.1.2 tag. But no there was no change to api for this version change

pepper-sync/src/sync/state.rs
```diff
@@ -368,18 +367,20 @@ pub(super) fn set_scan_priority(
 /// Any scan ranges that fully contain the `block_range` will be split out with the given `scan_priority`.
 /// Any scan ranges with `Scanning` or `Scanned` priority or with higher (or equal) priority than
 /// `scan_priority` will be ignored.
-fn punch_scan_priority(
+/// `split_scanning_ranges` is used for rare cases where a range already scanning needs to be reset back to its original priority and should be used with caution.
+pub(super) fn punch_scan_priority(
     sync_state: &mut SyncState,
     block_range: Range<BlockHeight>,
     scan_priority: ScanPriority,
+    split_scanning_ranges: bool,
```
zancas, line 375: Is this only used in the next PR?

zancas:
I haven't validated that the scan ordering works as intended.   I would be a lot more comfortable signing off on this PR if we had continuous benchmarking in place.

e.g.   https://bencher.dev/perf/zcash-speed/plots

pepper-sync/src/sync/state.rs
Oscar-Pepper: no, this is used to provide a new case of "punching in" where scanning ranges are allowed to be split with a section of it being set to a specified priority. this is now needed in the work in this PR as it adds logic to check a rare edge case where nullifiers were re-fetched pre-maturely. In this case, it discards these nullifiers and sets that chunk of "Scanning"  range back to "ScannedWithoutMapping". it is described in the PR description:

"Finally, this PR also fixes a bug where, in rare edges cases, ScannedWithoutMapping scans which re-fetch nullifiers could complete before lower ranges have completed scanning, potentially missing spends. This PR improves scan range seelction to decrease the chance of this occuring and also adds logic to discard these nullifiers and reset the scan range to ScannedWithoutMapping again to be re-fetched safely. This resulted in an improvement to punch_scan_priority to allow a larger range of Scanning priority to be "punched in" with the potentially smaller ScannedWithoutMapping range in the case of batching when the initial scan range exceeded the maximum nullifier batch size."

Oscar-Pepper:
> I haven't validated that the scan ordering works as intended. I would be a lot more comfortable signing off on this PR if we had continuous benchmarking in place.
> 
> e.g. https://bencher.dev/perf/zcash-speed/plots

I agree this benchmarking framework is top priority but unfortunately this critical bug fix cant wait for this framework. I will start looking into benchmarking now

pepper-sync/src/sync.rs
```diff
@@ -917,14 +912,106 @@ where
...
+                    .find(|scan_range| scan_range.priority() != ScanPriority::Scanned)
+                    .expect("the scan range being processed is not yet set to scanned so at least one unscanned range must exist");
+                if !first_unscanned_range
+                    .block_range()
+                    .contains(&scan_range.block_range().start)
+                    || !first_unscanned_range
+                        .block_range()
+                        .contains(&(scan_range.block_range().end - 1))
+                {
+                    // in this rare edge case, a scanned `ScannedWithoutMapping` range was the highest priority yet it was not the first unscanned range so it must be discarded to avoid missing spends
```
zancas, line 929: Given that ScannedWithoutMapping is the lowest priority apart from Scanned and Scanning, how can this edge case be hit?

Is this only possible as the result of a race between workers?

zancas (approved):
No further questions your honor.

PR #2095, opened 2026-01-26
notes on fix_large_wallet

zancas:
I pushed a copy of @Oscar-Pepper 's branch to zingolabs to make it easier to compare with the changes I am proposing on top.

### DIFF
--- a/pepper-sync/src/lib.rs
+++ b/pepper-sync/src/lib.rs
@@ -136,3 +136,9 @@ pub(crate) struct Orchard;
 impl SyncDomain for Orchard {
     const SHIELDED_PROTOCOL: ShieldedProtocol = ShieldedProtocol::Orchard;
 }
+
+use zcash_protocol::consensus;
+
+/// Trait alias for consensus parameters that can be shared across threads.
+trait SyncParameters: consensus::Parameters + Sync + Send + 'static {}
+impl<T: consensus::Parameters + Sync + Send + 'static> SyncParameters for T {}
--- a/pepper-sync/src/scan.rs
+++ b/pepper-sync/src/scan.rs
@@ -11,9 +11,10 @@ use incrementalmerkletree::Position;
 use zcash_client_backend::proto::compact_formats::{CompactBlock, CompactTx};
 use zcash_keys::keys::UnifiedFullViewingKey;
 use zcash_primitives::{transaction::TxId, zip32::AccountId};
-use zcash_protocol::consensus::{self, BlockHeight};
+use zcash_protocol::consensus::BlockHeight;
 
 use crate::{
+    SyncParameters,
     client::FetchRequest,
     error::{ScanError, ServerError},
     sync::ScanPriority,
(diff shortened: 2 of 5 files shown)


pepper-sync/Cargo.toml
```diff
@@ -52,7 +52,7 @@ subtle.workspace = true
 
 # Async
 futures.workspace = true
-tokio.workspace = true
+tokio = { workspace = true, features = ["sync", "time", "rt", "macros", "rt-multi-thread"] }
```
Oscar-Pepper: IMO its best practice to not add features unless they are needed otherwise it just inflates the binary size and dep tree for no reason. may i ask what is the purpose/intention of adding these features?

Cargo.toml
```diff
@@ -105,7 +105,7 @@ shellwords = "1"
 subtle = "2"
 tempfile = "3"
 thiserror = "2"
-tokio = "1"
+tokio = "1"  # tokio is now on 1.49, we're on 1.48.
```
Oscar-Pepper: I dont think we want to add the extra burden of keeping comments like this up-to-date. its likely that a cargo update will happen and this comment will stay unchanged. Im not sure what the benefit of this comment is? the cargo lock will provide the necessary info for the *actual* current version and crates.io etc. cna be used to find the *actual* latest version.

Oscar-Pepper (changes requested):
Im not sure why these changes are relevant to the fix_large_wallet_file_size branch. i have left comments regarding the proposed changes

Cargo.toml
zancas: Yeah...  probably could have just made it a GH comment instead of committing...

pepper-sync/Cargo.toml
zancas: Those features _are_ necessary for **pepper-sync**!

The reason **pepper-sync** worked without explicit reference in the manifest (prior to this) is solely due to feature unification.

That having been said, I'm not clear that marking them explicitly in this way is the best idea.  In particular, I haven't figured out how to ensure that the list of features included in this way is actually the full necessary set.

So, in short, there is no binary bloat because these aren't actually causing changes in the binary because they're all already **implicitly** included (due to feature unification).   I don't know exactly how to express this (possibly it's fine to just continue depending on feature unification for implicit incorporation) cleanly, but I do think it's something worth thinking about.

Possibly a near term step in the right direction is an issue, and/or "discussion".

zancas:
> Im not sure why these changes are relevant to the fix_large_wallet_file_size branch. i have left comments regarding the proposed changes

Note that the PR is draft.  I want to efficiently communicate my understanding, not necessarily to change the implementation.

This means that "change relevance" to the feature *might* not be more than that I noticed them while studying the feature.  For changes which turn out to be worth actually merging, AND aren't particularly relevant to this feature, those will be fine to merge later (into dev).  I prefer to note them here, since this is the version of the codebase I am focused on.   Then we can separate work that's actually relevant to the feature, from other notes where appropriate.

In cases where I want to change the implementation I'll denote that somehow...    *at least* that PR (which could be a modification of this one) will be non-draft.

SOO..   in short.    I don't necessarily expect this PR to be a formal proposal of changes to the feature..  this is still early investigation.

On that note I think I am reserving the right to rebase the `fix_large_wallet_file_size_za_01` branch until I "freeze" it by marking any PR that references it "Review Ready".

PR #2097, opened 2026-01-27
Improve note spendability

### LINKED ISSUE #2102: Spendable notes algorithm is limited
The "spend horizon" is the lowest block-height of a contiguous range of scanned blocks at the tip of the chain. 

One of the criteria for a note being spendable is: 

_The height of the note is equal to or above the "spend horizon"._

This means all nullifiers in this range have been mapped and, therefore, any note scanned in this range is known to not be spent if the nullfiier derived from the note does not match a nullifier in the wallet's nullifier map.

This has been complicated by the introduction of `ScannedWithoutMapping` scan range priority where the nullifiers are not mapped due to memory constraints and are instead checked for spends, when scanned, and then discarded to be re-fetched later on in the sync process.

Currently, the "spend horizon"-based criteria for spendability only applies to `Scanned` ranges as `ScannedWithoutMapping`-priority scans may or may not have checked the temporarily-collected nullifiers against the nullifiers derived from scan-detected notes.

This depends on whether the `ScannedWithoutMapping`-priority scan occurred before or after the range containing the note, was scanned. If before, the nullifiers would have been discarded and there would be no way to guarantee a note found later - but lower in the chain - is not spent unless this range of nullifiers were refetched. This issue becomes worsened by https://github.com/zingolabs/zingolib/pull/2091 due to the fact that the chain tip is no longer guaranteed to map nullifiers, making the spend horizon smaller.

Oscar-Pepper:
ontop of #2125

fixes #2102 

The problem stated in #2101 is solved in this PR by adding a 'refetched_nullifier_ranges' field to wallet notes which are all the scan ranges with `ScannedWithoutMapping` priority above the newly scanned note. This provides the data necessary to check whether the ranges where nullifier were not mapped had been refetched which allows the "spend horizon" to also include `ScannedWithoutMapping` ranges, massively improving note spendability

### DIFF
--- a/pepper-sync/src/scan/transactions.rs
+++ b/pepper-sync/src/scan/transactions.rs
@@ -389,6 +389,7 @@ where
                 position,
                 memo: Memo::from_bytes(memo_bytes.as_ref())?,
                 spending_transaction: None,
+                refetch_nullifier_ranges: Vec::new(),
             });
         }
     }
--- a/pepper-sync/src/sync.rs
+++ b/pepper-sync/src/sync.rs
@@ -552,6 +552,19 @@ where
             panic!("sync data must exist!");
         }
     };
+    // once sync is complete, all nullifiers will have been re-fetched so this note metadata can be discarded.
+    for transaction in wallet_guard
+        .get_wallet_transactions_mut()
+        .map_err(SyncError::WalletError)?
+        .values_mut()
+    {
+        for note in transaction.sapling_notes.as_mut_slice() {
+            note.refetch_nullifier_ranges = Vec::new();
+        }
+        for note in transaction.orchard_notes.as_mut_slice() {
+            note.refetch_nullifier_ranges = Vec::new();
+        }
+    }
     wallet_guard
         .set_save_flag()
         .map_err(SyncError::WalletError)?;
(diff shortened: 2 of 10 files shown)


zancas:
It seems like we don't have a way to assess the way changes affect performance.

Oscar-Pepper:
> It seems like we don't have a way to assess the way changes affect performance.

juanky and edickson have been testing this PR and the following PR (https://github.com/zingolabs/zingolib/pull/2115) manually. we should prioritise test framework but these PRs need to land as stable has been left in a regressed state since the last PR merged. this one fixes the note spendability issues introduced in the previous PR and actually makes many improvements so its top priority

zancas:
This intricate logic:

https://github.com/zingolabs/zingolib/pull/2097/changes#diff-94df7e7b584e94c8d171cefff125b80bd550c6a362b764304d89944f5f6cdfedR508-R523

Is an exact duplicate of this intricate logic:

https://github.com/zingolabs/zingolib/pull/2097/changes#diff-df8fe057121de50cf05cb973b7de09fe1749bdf480aed314eaec78d480d98cb0R300-R318

Unless I miscounted among the offset ) } ) at the end by a line.    That would be an easy mistake to make in code that's structured like this.

Please DRY this.

Apologies in advance if it's not EXACT..  it's at LEAST extremely similar, if not verbatim.

zancas (dismissed):
I think we should be extremely diligent about writing DRY code.

zancas:
This PR DRY's the code:  https://github.com/Oscar-Pepper/zingolib/pull/29

zancas (approved):
We need to extend our test framework to cover more chain-vs-wallet scenarios:

https://github.com/zingolabs/zingolib/issues/2153

PR #2112, opened 2026-02-01
Rework "resend" and "remove" tx

### LINKED ISSUE #2144: "resend" and "remove transaction" are not user friendly
The resend API requires a transaction that failed to transmit to stay in the `Calculated` state so a user can choose to either resend or remove it. This has proved a major issue as the notes spent in this transaction must not be reset back to unspent to prevent them being selected for other sends, resulting in the balance not including these unspent notes until the user decides how to handle this failed send. This means that when a send fails the user may experience a loss of balance if they are not aware they must remove the transaction to reset the selected notes spend status.

 The remove API has also been problematic as users may believe that removing a mempool transaction, for example, may reverse the transaction only to see the transactions later re-appear from the mempool or in a confirmed block.

Oscar-Pepper:
fixes #2144

This PR fixes the issue described in #2144 by adding a `Failed` variant to ConfirmationStatus. The new logic changes the status flow of transactions so that when a transaction fails to transmit it is updated from `Calculated` to `Failed`. An automatic retry has been added to attempt to retry the transmission 3 times before failing. On setting the transaction to `Failed`, the notes selected for spending are reset to "unspent", restoring the balance and allowing them to be immediately re-selected for other sends. This `Failed` transaction is stored in the wallet so the user may retrieve info such as the memo before removing the transaction. The "remove transaction" API now only allows the removal of failed transactions. If a transaction has been in the `Transmitted` or `Mempool` state for 3 confirmed new blocks without being confirmed, all spends will be reset to restore the balance and allow the user to spend these notes in another transactions, preventing funds from getting stuck in invalid transactions. If a `Transmitted` or `Mempool` transaction has expired if will set to `Failed` allowing the user to remove the transaction from the wallet. Confimation status now also includes methods for determining unconfirmed (pending ) and failed states.

This PR also cleans up the send code moving methods and fields to their correct locatons, improving the error structure and removes outdated unused "send progress".


Oscar-Pepper:
closed for #2171

zancas:
Pieces of #2171 have merged..   I'll merge dev into this when dev is ready.

juanky201271:
We can safely close this PR... because this one was merged... #2181

PR #2115, opened 2026-02-01
Fix range resets for refetching nullifiers

### LINKED ISSUE #2116: pepper-sync needs a 'RefetchingNullifers' scan range priority
an additional scan range priority is needed to track ranges where nullifiers are being refetched due to an issue where the already scanned range is reset to 'FoundNote' priority when sync is stopped and restarted. Currently, when nullifiers are re-fetched the range is set from `ScannedWithoutMapping` to `Scanning`. When sync is restarted all `Scanning` ranges are reset back to `FoundNote` and there is no information to distinguish ranges that have already been scanned and are just re-fetching nullifiers or ranges that were being scanned. An additional `RefetchingNullifiers` scan range is needed to store this information in the wallet between sync runs. This will also allow scan progress to not need to include ranges that have already been scanned, fixing a related issue where the sync completion % drops during nullifier re-fetching.

Oscar-Pepper:
on top of #2097 

fixes #2116 

As described in the issue above, this PR creates a new `RefetchingNullifiers` scan range priority which is used to allow the wallet to store information about which ranges were already scanned but were being partially re-scanned only to re-fetch the nullifiers. This also allows the sync completion % to never drop when nullifiers are being re-fetched which is confusing to the user.

### DIFF
--- a/pepper-sync/README.md
+++ b/pepper-sync/README.md
@@ -34,7 +34,7 @@ Pepper-sync is a rust-based sync engine library for wallets operating on the zca
 8. Set the first 10 blocks after the highest previously scanned blocks to "verify" priority to check for re-org.
 
 ## Scanning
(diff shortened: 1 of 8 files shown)


zancas (approved):
I'd prefer that we add more tests, but the motivating bug is fixed by this logic, if I understand it correctly.

PR #2125, opened 2026-02-03
rework pepper-sync wallet height logic to avoid panic when birthday i…

Oscar-Pepper:
…s between 1-100 blocks above the chain


zancas:
After helperizing and renaming a bunch on my own branch, I am comfortable that I understand this code.

I would prefer to test new code, but since I am doing that work on another branch and this change is relatively conservative I will note that WE ARE FOLLOWING UP WITH TESTS.

zancas (approved):

zancas:
Locally tested...   CI was slowww...

PR #2140, opened 2026-02-08
change MAX_VERIFICATION_WINDOW -> MAX_REORG_ALLOWANCE

zancas:
Use "REORG" and "ALLOWANCE" to give hints to naive readers. 

The MAX_REORG_ALLOWANCE is the number of blocks above the proxy-reported best-chain-height, that the wallet is allowed to have recorded as its fully scanned height.

If the wallet has scanned more that MAX_REORG_ALLOWANCE blocks ABOVE the proxy's best height this is evidence that there is a deviation from normal consensus parameter values.

Most likely there has been a (massive) reorganization.

I chose this name to reflect this:

MAX:                   this is an upper bound (tolerance)
REORG:              this is why the threshold might be exceeded
ALLOWANCE:    values less than this are allowed to recover

Other names that seem viable to me might involve the word "TOLERANCE" instead of "ALLOWANCE".

It's also the case that this value is only exceeded in catastrophic cases, so perhaps that information could somehow be incorporated?

I don't think this name is perfect, but I think it's a bit more explicit than the current name.

### DIFF
--- a/pepper-sync/README.md
+++ b/pepper-sync/README.md
@@ -16,7 +16,7 @@ Pepper-sync is a rust-based sync engine library for wallets operating on the zca
 ## Terminology
 - Chain height - highest block height of best chain from the server.
 - Chain tip - the range of blocks at the top of the blockchain; Starting from the lowest block which contains the last note commitment to the latest shard of each shielded protocol; Ending at the chain height.
-- Wallet height - highest block height of blockchain known to the wallet.
+- Last Known Chain Height - highest block height of blockchain known to the wallet.
 - Fully scanned height - block height in which the wallet has completed scanning all blocks equal to and below this height.
 - Shard range - the range of blocks that contain all note commitments to a fully completed shard for a given shielded protocol.
 - Nullifier map - a map of all the nullifiers collected from each transaction's shielded inputs/spends during scanning.
--- a/pepper-sync/src/sync.rs
+++ b/pepper-sync/src/sync.rs
@@ -49,7 +49,7 @@ pub(crate) mod state;
 pub(crate) mod transparent;
 
 const MEMPOOL_SPEND_INVALIDATION_THRESHOLD: u32 = 3;
-pub(crate) const MAX_VERIFICATION_WINDOW: u32 = 100;
+pub(crate) const MAX_REORG_ALLOWANCE: u32 = 100;
 const VERIFY_BLOCK_RANGE_SIZE: u32 = 10;
 
 /// A snapshot of the current state of sync. Useful for displaying the status of sync to a user / consumer.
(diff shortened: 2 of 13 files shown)


Oscar-Pepper (dismissed):
I dont mind changing this name but the name selected and the PR description seems to be missing the primary use case for this value. The primary use case is the maximum number of blocks that the wallet will be truncated due to re-org before it returns an error. this is because re-org will never statistically go back more than 100 blocks so it indicates something else has gone wrong, i.e. the wallet is connected to a bad server or incorrect chain. Does this new name describe this effectively from your point of view?

zancas:
> The primary use case is the maximum number of blocks that the wallet will be truncated due to re-org before it returns an error. this is because re-org will never statistically go back more than 100 blocks so it indicates something else has gone wrong

Yes!  This is what I was trying to convey..  that if the difference between last_known_chain_height, and chain_height is greater than or equal to, the "ALLOWANCE" then something has gone wrong.

I was focused on this line of code:

https://github.com/zingolabs/zingolib/blob/457797426d25c873c20968040b14b0bd02a7a9a8/pepper-sync/src/sync.rs#L370

If it's used differently elsewhere I might have missed important use cases when trying to think of name.

I was trying to express that differences of larger magnitude are not, in some sense, allowed.

Oscar-Pepper:
> > The primary use case is the maximum number of blocks that the wallet will be truncated due to re-org before it returns an error. this is because re-org will never statistically go back more than 100 blocks so it indicates something else has gone wrong
> 
> Yes! This is what I was trying to convey.. that if the difference between last_known_chain_height, and chain_height is greater than or equal to, the "ALLOWANCE" then something has gone wrong.
> 
> I was focused on this line of code:
> 
> https://github.com/zingolabs/zingolib/blob/457797426d25c873c20968040b14b0bd02a7a9a8/pepper-sync/src/sync.rs#L370
> 
> If it's used differently elsewhere I might have missed important use cases when trying to think of name.
> 
> I was trying to express that differences of larger magnitude are not, in some sense, allowed.

ah yes ok, so its primary use case is for the maximum verification window (in other words the largest number of blocks re-org can truncate the wallet before erroring) here:
https://github.com/zingolabs/zingolib/blob/dev/pepper-sync/src/sync.rs#L1128-L1132

zancas:
OK, it seems like the MAX_REORG_ALLOWANCE works there too.

Oscar-Pepper (approved):

Oscar-Pepper:
it looks lijke my approval hasnt satisfied the "review required" check. maybe because i merged the PR into this PR

zancas:
Yikes I can neither approve, nor merge (without bypassing)...

PR #2142, opened 2026-02-08
typo fix

zancas:
Builds on #2141 .

The motivation is to cleanly encapsulate proposed changes.  Among other things this will ease review!

### DIFF
--- a/pepper-sync/src/sync.rs
+++ b/pepper-sync/src/sync.rs
@@ -354,7 +354,7 @@ where
         .await
     });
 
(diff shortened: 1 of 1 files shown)


Oscar-Pepper (commented):
i would argue that sync has already started but scanning is a sub-task within sync that has yet to start. i am open to this change if we want to just use sync as another name for scanning but i see benefits from maintaining this distinguished meaning. If sync has not started then transparent address discovery is part of what? I see this, for example, as part of sync but not part of scanning.

PR #2143, opened 2026-02-08
Upgrade init height constraint code

zancas:
Builds on #2142 

This conservative change:

* helperizes recently introduced bug-fix code
* starts a mock/unit/patch framework to test the helper
* starts adding tests of the helper

Because this change both adds tests (and more importantly test tooling) and makes no logical change to production code, I will mark it "test-only".

### DIFF
--- a/pepper-sync/src/error.rs
+++ b/pepper-sync/src/error.rs
@@ -28,7 +28,7 @@ where
     SyncModeError(#[from] SyncModeError),
     /// Chain error.
     #[error("wallet height is more than {0} blocks ahead of best chain height")]
-    ChainError(u32),
+    LastWalletKnownHeightAboveAllowance(u32),
     /// Shard tree error.
     #[error("shard tree error. {0}")]
     ShardTreeError(#[from] ShardTreeError<Infallible>),
--- a/pepper-sync/src/error.rs
+++ b/pepper-sync/src/error.rs
@@ -27,8 +27,8 @@ where
     #[error("sync mode error. {0}")]
     SyncModeError(#[from] SyncModeError),
     /// Chain error.
-    #[error("wallet height is more than {0} blocks ahead of best chain height")]
-    LastLocalKnownHeightAboveAllowance(u32),
+    #[error("wallet height {0} is more than {1} blocks ahead of best chain height {2}")]
+    LastLocalKnownHeightAboveAllowance(BlockHeight, u32, BlockHeight),
     /// Shard tree error.
     #[error("shard tree error. {0}")]
     ShardTreeError(#[from] ShardTreeError<Infallible>),
--- a/pepper-sync/src/error.rs
+++ b/pepper-sync/src/error.rs
@@ -28,7 +28,7 @@ where
     SyncModeError(#[from] SyncModeError),
     /// Chain error.
     #[error("wallet height {0} is more than {1} blocks ahead of best chain height {2}")]
-    LastLocalKnownHeightAboveAllowance(BlockHeight, u32, BlockHeight),
+    ChainError(BlockHeight, u32, BlockHeight),
     /// Shard tree error.
     #[error("shard tree error. {0}")]
     ShardTreeError(#[from] ShardTreeError<Infallible>),
--- a/pepper-sync/src/error.rs
+++ b/pepper-sync/src/error.rs
@@ -28,7 +28,7 @@ where
     SyncModeError(#[from] SyncModeError),
     /// Chain error.
     #[error("wallet height is more than {0} blocks ahead of best chain height")]
-    LastWalletKnownHeightAboveAllowance(u32),
+    LastLocalKnownHeightAboveAllowance(u32),
     /// Shard tree error.
     #[error("shard tree error. {0}")]
     ShardTreeError(#[from] ShardTreeError<Infallible>),
--- a/pepper-sync/src/mocks.rs
+++ b/pepper-sync/src/mocks.rs
@@ -207,7 +207,7 @@ impl SyncBlocks for MockWallet {
         &mut self,
     ) -> Result<&mut std::collections::BTreeMap<BlockHeight, crate::wallet::WalletBlock>, Self::Error>
     {
-        todo!()
+        Ok(&mut self.wallet_blocks)
     }
 }
 impl SyncTransactions for MockWallet {
--- a/pepper-sync/src/mocks.rs
+++ b/pepper-sync/src/mocks.rs
@@ -18,10 +18,12 @@ pub(super) enum MockWalletError {
 
 impl MockWalletError {}
 type SyncStatePatch = Box<dyn Fn(&SyncState) -> Result<&SyncState, MockWalletError>>;
+type GetBirthdayPatch = Box<dyn Fn(&BlockHeight) -> Result<BlockHeight, MockWalletError>>;
 pub(super) struct MockWallet {
     birthday: BlockHeight,
     sync_state: SyncState,
     get_sync_state_patch: Option<SyncStatePatch>,
+    get_birthday_patch: Option<GetBirthdayPatch>,
     wallet_blocks: BTreeMap<BlockHeight, WalletBlock>,
     wallet_transactions: HashMap<TxId, WalletTransaction>,
     nullifier_map: NullifierMap,
@@ -35,6 +37,7 @@ impl Default for MockWalletBuilder {
             birthday: BlockHeight::from_u32(0),
             sync_state: SyncState::new(),
             get_sync_state_patch: None,
+            get_birthday_patch: None,
             wallet_blocks: BTreeMap::new(),
             wallet_transactions: HashMap::new(),
             nullifier_map: NullifierMap::new(),
@@ -46,7 +49,8 @@ impl Default for MockWalletBuilder {
 
 pub(super) struct MockWalletBuilder {
     birthday: BlockHeight,
-    get_sync_state_patch: Option<Box<dyn Fn(&SyncState) -> Result<&SyncState, MockWalletError>>>,
+    get_sync_state_patch: Option<SyncStatePatch>,
+    get_birthday_patch: Option<GetBirthdayPatch>,
     sync_state: SyncState,
     wallet_blocks: BTreeMap<BlockHeight, WalletBlock>,
     wallet_transactions: HashMap<TxId, WalletTransaction>,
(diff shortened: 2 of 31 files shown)


pepper-sync/src/error.rs
```diff
@@ -28,7 +28,7 @@ where
     SyncModeError(#[from] SyncModeError),
     /// Chain error.
     #[error("wallet height is more than {0} blocks ahead of best chain height")]
-    ChainError(u32),
+    LastWalletKnownHeightAboveAllowance(u32),
```
Oscar-Pepper: I personally dont think this rename describes the errors better than the current name, it only describes one case of why it got to the error. The reason it is called "ChainError" is because sync has reason to believe that the wallet is actually connected to the wrong chain compared with the chain the wallet was previously connected to. this may actually happen for other reasons (for example when the birthday is higher than the chain height) than the reason described by this new error name change. I think a general error name that describes a range of cases that result in sync believing the chain is incorrect should be preffred over describing a single case of how this error may be hit.

pepper-sync/src/mocks.rs
```diff
@@ -0,0 +1,268 @@
...
+    get_sync_state_patch: Option<SyncStatePatch>,
+    get_birthday_patch: Option<GetBirthdayPatch>,
+    wallet_blocks: BTreeMap<BlockHeight, WalletBlock>,
+    wallet_transactions: HashMap<TxId, WalletTransaction>,
+    nullifier_map: NullifierMap,
+    outpoint_map: BTreeMap<OutputId, ScanTarget>,
+    shard_trees: ShardTrees,
+}
+
+impl Default for MockWalletBuilder {
```
Oscar-Pepper: can we have impls of structs below the structs?

pepper-sync/src/sync.rs
```diff
@@ -433,7 +412,7 @@ where
     )
     .await?;
 
-    let initial_verification_height = wallet_guard
+    let lowest_unscanned_height = wallet_guard
```
Oscar-Pepper: this is not true! there may be unscanned ranges below the highest scanned height! this is exactly as the name suggests... it is the "first" or "intial" height that verifaction starts from, this verification height may get lower due to re-org detection but the initial height must be known to track if re-org has gone above the max verification window

pepper-sync/src/sync.rs
```diff
@@ -597,6 +576,202 @@ where
     })
 }
 
+/// This ensures that the height used as the lower bound of the sync is valid.
```
Oscar-Pepper, line 594: this is not the "lower bound" of sync, that would be the lowest unscanned range in the wallet sync state or the wallet birthday depending on which way you want to look at it. sync needs a "wallet height" aka the highest block that the wallet contains data about. if the wallet contains no data then the birthday-1 is used, if re-org has invalidated some of the data in the wallet, this needs to be stabilised with a new wallet height returned. we can call this "last known chain height" thats fine, although there are exceptions like when the wallet is syncing for the first time of course or when the "last known chain hieght" is invalid due to being higher than the current chain height

pepper-sync/src/sync.rs
```diff
@@ -597,6 +576,202 @@ where
...
+/// The comparison takes two input heights and uses several constant bounds to select the best height
+/// to bound the sync.
```
Oscar-Pepper, line 596: again this is not a bound of sync its the lower bound of the new scan range that *may* be created (if the chain height is higher than this lower bound)

pepper-sync/src/sync.rs
```diff
@@ -597,6 +576,202 @@ where
...
+/// The input parameter heights are:
+///
+///   (1) chain_height:
+///       * the best block-height reported by the proxy (zainod or lwd)
+///   (2) last_known_chain_height
+///       * the last max height the wallet recorded from earlier scans
```
Oscar-Pepper, line 602: i think this is better described as the chain height the wallet recorded from previous scans, just as the name suggests

pepper-sync/src/sync.rs
```diff
@@ -597,6 +576,202 @@ where
...
+///
+/// The constants are:
+///   (1) MAX_REORG_ALLOWANCE:
+///       * the number of blocks AHEAD of proxy_reported_chain_height allowed
```
Oscar-Pepper, line 606: this is better described as the maximum number of blocks that can re-orged

pepper-sync/src/sync.rs
```diff
@@ -597,6 +576,202 @@ where
...
+            if last_known_chain_height - chain_height >= MAX_REORG_ALLOWANCE {
+                // There's a human attention requiring problem, the wallet supplied
+                // last_known_chain_height is more than MAX_REORG_ALLOWANCE **above**
+                // the proxy's reported height.
+                return Err(SyncError::LastWalletKnownHeightAboveAllowance(
+                    MAX_REORG_ALLOWANCE,
+                ));
+            }
+            truncate_wallet_data(wallet, chain_height)?;
+            // The wallet reported height is above the current proxy height
```
Oscar-Pepper, line 634:
```suggestion
            // The last known chain height to the wallet is above the current chain height
```

this would be my suggestion, however these comments are unecessary, that is why we chose names and write code that describes this already. this comment gives no extra information than the logic above. I prefer letting code speak for itself here.

pepper-sync/src/sync.rs
```diff
@@ -597,6 +576,202 @@ where
...
+    W: SyncBlocks + SyncTransactions + SyncNullifiers + SyncOutPoints + SyncShardTrees,
+    P: zcash_protocol::consensus::Parameters,
+{
+    let sync_state = wallet.get_sync_state().map_err(SyncError::WalletError)?;
+    if let Some(last_known_chain_height) = sync_state.last_known_chain_height() {
+        if last_known_chain_height > chain_height {
+            if last_known_chain_height - chain_height >= MAX_REORG_ALLOWANCE {
+                // There's a human attention requiring problem, the wallet supplied
+                // last_known_chain_height is more than MAX_REORG_ALLOWANCE **above**
+                // the proxy's reported height.
```
Oscar-Pepper, line 626: the only reason we have to return an error here is because we dont have the ability to truncate past the max re-org window so we cannot restore back to the current chain height

pepper-sync/src/sync.rs
```diff
@@ -597,6 +576,202 @@ where
...
+                ));
+            }
+            truncate_wallet_data(wallet, chain_height)?;
+            // The wallet reported height is above the current proxy height
+            // reset to the proxy height.
+            return Ok(chain_height);
+        }
+        // The last wallet reported height is equal or below the proxy height
+        // since it was derived from a bday in an earlier scan, we don't check its
+        // lower bound.
```
Oscar-Pepper, line 640: i dont understnad this comment, in this case it is just returning the chain height from the previous sync. it has nothing to do with birthdays

pepper-sync/src/sync.rs
```diff
@@ -597,6 +576,202 @@ where
...
+        // TODO:  Think through the possibilty of syncing to a short side chain,
+        // this logic doesn't explicitly handle the scan, but the overlap of range
+        // bounds may handle it elsewhere.
+        Ok(last_known_chain_height)
+    } else {
+        dbg!("There was a syncstate, but no last_known_chain_height!!");
```
Oscar-Pepper, line 646: please remove dbg

pepper-sync/src/sync.rs
```diff
@@ -597,6 +576,202 @@ where
...
+        let raw_bday = wallet.get_birthday().map_err(SyncError::WalletError)?;
```
Oscar-Pepper, line 647: what is raw? its just a birthday

pepper-sync/src/sync.rs
```diff
@@ -597,6 +576,202 @@ where
...
+        if raw_bday > chain_height {
+            // Human attention requiring error, a bday *above* the proxy reported
+            // chain tipe has been provided
+            return Err(SyncError::LastWalletKnownHeightAboveAllowance(
+                raw_bday - chain_height,
+            ));
+        }
+        // The bday is set with a floor of the Sapling Epoch -1
+        // TODO:  Confirm that the Sapling Epoch -1 is the correct floor
+        Ok(sapling_floored_bday(consensus_parameters, raw_bday) - 1)
```
Oscar-Pepper, line 659: i dont see the necessity to abbreviate birthday to bday here, we all have LSPs now and i think code is more readable this way. this is just my style choice but this would be consistent with the rest of peppersync

pepper-sync/src/sync.rs
```diff
@@ -597,6 +576,202 @@ where
...
+        let raw_bday = wallet.get_birthday().map_err(SyncError::WalletError)?;
+        if raw_bday > chain_height {
+            // Human attention requiring error, a bday *above* the proxy reported
+            // chain tipe has been provided
+            return Err(SyncError::LastWalletKnownHeightAboveAllowance(
+                raw_bday - chain_height,
+            ));
+        }
+        // The bday is set with a floor of the Sapling Epoch -1
+        // TODO:  Confirm that the Sapling Epoch -1 is the correct floor
```
Oscar-Pepper, line 658: sapling epoch -1 is correct because this means that the newly create range will start from sapling epoch (wallet height + 1)

pepper-sync/src/sync.rs
```diff
@@ -597,6 +576,202 @@ where
...
+            blossom: Some(BlockHeight::from_u32(3)),
+            heartwood: Some(BlockHeight::from_u32(3)),
+            canopy: Some(BlockHeight::from_u32(3)),
+            nu5: Some(BlockHeight::from_u32(3)),
+            nu6: Some(BlockHeight::from_u32(3)),
+            nu6_1: Some(BlockHeight::from_u32(3)),
+        };
+        use crate::{error::SyncError, mocks::MockWalletError, sync::checked_wallet_height};
+        // It's possible an error from an implementor's get_sync_state could bubble up to checked_wallet_height
+        // this test shows that such an error is raies wrapped in a WalletError and return as the Err variant
```
Oscar-Pepper, line 680: raies?

pepper-sync/src/sync.rs
```diff
@@ -597,6 +576,202 @@ where
...
+                res,
+                Err(SyncError::WalletError(
+                    crate::mocks::MockWalletError::AnErrorVariant(ref s)
+                )) if s == test_error
+            ));
+        }
+
+        mod no_last_known_chain_height {
+            use super::*;
+            // If there are know scan_ranges in the SyncState t
```
Oscar-Pepper:
```suggestion
            // If there are no scan_ranges in the SyncState
```

Oscar-Pepper (changes requested):
really like the addition of unit tests an unit test framework! i have left some comments that hopefully give more insight into the code touched here

pepper-sync/src/error.rs
zancas: OK, I will change it back.   I am also adding more fields so that the error is more informatiive.

pepper-sync/src/sync.rs
zancas: Gotcha'.   Changing from "lowest"..    I am not very clear on what "verification" means.

zancas: Changed to "initial_unscanned_height".

zancas:
#2154 supercedes this.

PR #2146, opened 2026-02-11
rework waitsync flag to use the lightclient await_sync method

Oscar-Pepper:


Tomas-M:
I tested this patch on top of current sources.
Compilation produces the warning mentioned below. Furthermore, it looks like the intended behavior (waiting for sync and then executing the command) does not work. I tried zingo-cli --waitsync messages, and it produced empty output (the wallet has indeed some messages). The ouput is only 
```
Launching sync task...
Launching save task...
Save task shutdown successfully.
Zingo CLI quit successfully.
```
(without this patch, the output also includes messages).

Compilation warning is this:
```
warning: field `params` is never read
   --> zingo-cli/src/lib.rs:275:5
    |
274 | pub struct ConfigTemplate {
    |            -------------- field in this struct
275 |     params: Vec<String>,
    |     ^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default
```

Tomas-M:
It seems to me that lightclient.await_sync().await probably doesnt work for waitsync, it does not wait if the sync was not started yet (maybe in another thread, start not completed yet? just a guess)

Oscar-Pepper:
> I tested this patch on top of current sources. Compilation produces the warning mentioned below. Furthermore, it looks like the intended behavior (waiting for sync and then executing the command) does not work. I tried zingo-cli --waitsync messages, and it produced empty output (the wallet has indeed some messages). The ouput is only
> 
> ```
> Launching sync task...
> Launching save task...
> Save task shutdown successfully.
> Zingo CLI quit successfully.
> ```
> 
> (without this patch, the output also includes messages).
> 
> Compilation warning is this:
> 
> ```
> warning: field `params` is never read
>    --> zingo-cli/src/lib.rs:275:5
>     |
> 274 | pub struct ConfigTemplate {
>     |            -------------- field in this struct
> 275 |     params: Vec<String>,
>     |     ^^^^^^
>     |
>     = note: `#[warn(dead_code)]` on by default
> ```

Ok thanks for the info. My testing has not reproduced these issues but I will try to create a reproduction so we can fix this asap

Oscar-Pepper:
> It seems to me that lightclient.await_sync().await probably doesnt work for waitsync, it does not wait if the sync was not started yet (maybe in another thread, start not completed yet? just a guess)

For context, the sync handle is given to the LightClient before the `.sync()` method exits so when `.await_sync()` is called the handle must already exist. I will look into what might be happening here, there has been some indication there may be a `poll_sync()` bug that may cause this altough its not clear to me how this could happen. I will give an update soon

zancas:
Moving this to draft until we have a solution in progress.

zancas:
There is now a tracking issue for the observed buggy behavior:  #2195

Oscar-Pepper:
I have done alot of testing and this code is working as intended AFAIC. one limitation of waitsync is that if a sync error is hit it will call the command on the state of sync at the time of error. maybe this is the issue @Tomas-M is facing? I have left a comment here https://github.com/zingolabs/zingolib/issues/2197#issuecomment-3939947408. @Tomas-M if you could provide more info that would be great

Tomas-M:
I tried to reproduce my problem again.
I cloned zingolib repo, switched to "stable" branch, applied this patch and compiled.
Running "zingo-cli --waitsync messages" works fine.
I am not able to reproduce my problem anymore.
It was maybe some error on my side :(

zancas:
This is good to know @Tomas-M .    Next time you notice any issue, please use the github issue system.

If we can get a shared test-wallet, and reference specific PRs in the issue that will also accelerate bug-fixes.

zancas (approved):

PR #2147, opened 2026-02-11
Fix/sapling activation height fallback

Timi16:
Summary
  This replaces a hard expect on sapling activation height with a safe fallback to BlockHeight::from_u32(1), avoiding a panic in
  edge cases.

  Changes

  - Use unwrap_or(BlockHeight::from_u32(1)) for sapling activation height.
  - Keep wallet birthday comparison logic unchanged.


pepper-sync/src/sync.rs
```diff
@@ -1614,7 +1499,7 @@ fn checked_birthday<W: SyncWallet>(
     let wallet_birthday = wallet.get_birthday()?;
     let sapling_activation_height = consensus_parameters
         .activation_height(consensus::NetworkUpgrade::Sapling)
-        .expect("sapling activation height should always return Some");
+        .unwrap_or(BlockHeight::from_u32(1));
```
dorianvp, line 1502: Even though we might want to derive the correct activation height to use from some internal state, I am fine with this as is, especially because it unblocks regtest usage.

zingolib/build.rs
dorianvp, line 1: I don't understand why the Sapling activation height fallback PR would need to edit the `build.rs` file.

zingolib/Cargo.toml
dorianvp, line 1: Same as above.

.DS_Store
dorianvp, line 1: I think this file was accidentally committed. Please remove it.

dorianvp (changes requested):
I see the same changes from #2089, even though this PR is only targeting a subset of those changes. Please remove the out-of-scope changes.

Also, I would recommend cleaning up the git history and force-pushing, so that the `.DS_Store` doesn't show up.

pepper-sync/src/sync.rs
Timi16: Yeah so the change is valid right?

zingolib/build.rs
Timi16: My bad would remove git commit history

.DS_Store
Timi16: Yes i would for sure

PR #2150, opened 2026-02-12
Fix unexpectedeof error on wallet read

Oscar-Pepper:
to test run:
cargo nextest run --run-ignored=all reload_wallet_after_short_sync

This PR adds flushing and file syncing to file write code for improved write safety. It also writes to temp file before replacing the main wallet file to avoid reads midway through a write. This fixes an UnexpectedEOF error that occurs intermittently on wallet reload.

### DIFF
--- a/libtonode-tests/tests/concrete.rs
+++ b/libtonode-tests/tests/concrete.rs
@@ -4570,3 +4570,66 @@ mod send_all {
         ));
     }
 }
+
+mod testnet_test {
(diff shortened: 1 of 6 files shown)


zancas (approved):
Blocking writes to tempfile, atomic rename, passing test.

I have a follow on PR with some ergonomics, but this is a nice bit of bug-fixing!

PR #2154, opened 2026-02-13
Upgrade init height constraint code

zancas:
Supercedes #2143 

Some tests are not yet running, but this is an improvement to our unit test capabilities already.  I think we should get it into dev, and keep iterating on it in conjuction with other PRs.

### DIFF
--- a/pepper-sync/src/wallet.rs
+++ b/pepper-sync/src/wallet.rs
@@ -207,14 +207,14 @@ impl SyncState {
         {
             Some(last_scanned_range.block_range().end - 1)
         } else {
-            self.scan_range_start().map(|birthday| birthday - 1)
+            self.scan_start_height().map(|start| start - 1)
         }
     }
 
     /// Returns the wallet birthday or `None` if `self.scan_ranges` is empty.
     ///
     #[must_use]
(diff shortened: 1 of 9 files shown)


pepper-sync/src/wallet.rs
```diff
@@ -207,15 +207,14 @@ impl SyncState {
...
+            self.scan_start_height().map(|start| start - 1)
         }
     }
 
     /// Returns the wallet birthday or `None` if `self.scan_ranges` is empty.
     ///
-    /// If the wallet birthday is below the sapling activation height, returns the sapling activation height instead.
     #[must_use]
-    pub fn wallet_birthday(&self) -> Option<BlockHeight> {
+    pub fn scan_start_height(&self) -> Option<BlockHeight> {
```
dorianvp, line 217: Non-blocking. I was confused at first, because I thought this function would `scan` something, but later I realized it was an accessor/getter...

dorianvp (approved):
Small non-blocking suggestion.

pepper-sync/src/wallet.rs
zancas: Thanks!  I will fix this.

PR #2172, opened 2026-02-16
Not silently drop Transaction Serialization error details in send.

fluidvanadium:


Oscar-Pepper:
im closing this PR as its actually covered in #2171 where TransmissionError is simplified and these variants are moved to WalletError with std::io fields

PR #2174, opened 2026-02-17
Set to zcash protocol blockheight

### LINKED ISSUE #2173: importing BlockHeight from zcash_primitives is deprecated
It should be imported from the much lighter weight zcash_protocol crate instead.

zancas:
Fixes: #2173

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1021,18 +1021,18 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.58"
+version = "4.5.59"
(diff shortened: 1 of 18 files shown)


dorianvp (commented):
I think we should be using `zcash_primitives` and not `zcash_protocol`, as it's higher level...

See the mermaid diagram [here](https://github.com/zcash/librustzcash) and the walkthrough video attached.

dorianvp (approved):

zancas:
Prefer primitives instead.

PR #2194, opened 2026-02-20
clean up for zingolib v3.0.0 rc

Oscar-Pepper:
- reverts naming to use wallet birthday as the code reads much clearer. to avoid previous confusion with sapling activation heights, sync now returns an error if the birthday returned is lower than the sapling activation height.
- fixed a var name and comments that were misleading or incorrect
- added scan range truncation, preventing truncated wallets to still have `last_known_chain_heights` higher than the current `chain_height`

### DIFF
--- a/pepper-sync/src/error.rs
+++ b/pepper-sync/src/error.rs
@@ -29,6 +29,11 @@ where
     /// Chain error.
     #[error("wallet height {0} is more than {1} blocks ahead of best chain height {2}")]
     ChainError(u32, u32, u32),
+    /// Birthday below sapling error.
+    #[error(
+        "birthday {0} below sapling activation height {1}. pre-sapling wallets are not supported!"
+    )]
+    BirthdayBelowSapling(u32, u32),
     /// Shard tree error.
     #[error("shard tree error. {0}")]
     ShardTreeError(#[from] ShardTreeError<Infallible>),
--- a/pepper-sync/src/sync.rs
+++ b/pepper-sync/src/sync.rs
@@ -1,6 +1,5 @@
 //! Entrypoint for sync engine
 
-use std::cmp;
 use std::collections::{BTreeMap, HashMap};
 use std::ops::Range;
 use std::sync::Arc;
@@ -32,6 +31,7 @@ use crate::keys::transparent::TransparentAddressId;
 use crate::scan::ScanResults;
 use crate::scan::task::{Scanner, ScannerState};
 use crate::scan::transactions::scan_transaction;
+use crate::sync::state::truncate_scan_ranges;
 use crate::wallet::traits::{
     SyncBlocks, SyncNullifiers, SyncOutPoints, SyncShardTrees, SyncTransactions, SyncWallet,
 };
@@ -396,15 +396,14 @@ where
     )
     .await?;
 
-    state::update_scan_ranges(
+    let initial_reorg_detection_start_height = state::update_scan_ranges(
         consensus_parameters,
         last_known_chain_height,
         chain_height,
         wallet_guard
             .get_sync_state_mut()
             .map_err(SyncError::WalletError)?,
-    )
-    .await;
+    );
 
     state::set_initial_state(
         consensus_parameters,
@@ -413,13 +412,6 @@ where
         chain_height,
     )
     .await?;
-
-    let initial_unscanned_height = wallet_guard
-        .get_sync_state()
-        .map_err(SyncError::WalletError)?
-        .highest_scanned_height()
-        .expect("scan ranges must be non-empty")
-        + 1;
 
     expire_transactions(&mut *wallet_guard)?;
 
@@ -451,7 +443,7 @@ where
                     &ufvks,
                     scan_range,
                     scan_results,
-                    initial_unscanned_height,
+                    initial_reorg_detection_start_height,
                     config.performance_level,
                     &mut nullifier_map_limit_exceeded,
                 )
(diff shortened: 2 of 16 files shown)


pepper-sync/src/sync/state.rs
```diff
@@ -80,13 +81,19 @@ pub(super) async fn update_scan_ranges(
     set_chain_tip_scan_range(consensus_parameters, sync_state, chain_height);
     merge_scan_ranges(sync_state, ScanPriority::ChainTip);
 
-    let verification_height = sync_state
+    let reorg_detection_start_height = sync_state
```
zancas, line 84: I really like this name!!!

pepper-sync/src/sync/state.rs
```diff
@@ -146,6 +153,31 @@ async fn create_scan_range(
     sync_state.scan_ranges.push(new_scan_range);
 }
 
+/// Splits the range containing [`truncate_height` + 1] and removes all ranges containing block heights above
+/// `truncate_height`.
+pub(super) fn truncate_scan_ranges(truncate_height: BlockHeight, sync_state: &mut SyncState) {
```
zancas, line 159: Why not define this as a method of a SyncState?

pepper-sync/src/error.rs
```diff
@@ -29,6 +29,11 @@ where
     /// Chain error.
     #[error("wallet height {0} is more than {1} blocks ahead of best chain height {2}")]
     ChainError(u32, u32, u32),
+    /// Birthday below sapling error.
+    #[error(
+        "birthday {0} below sapling activation height {1}. pre-sapling wallets are not supported!"
+    )]
+    BirthdayBelowSapling(u32, u32),
```
zancas, line 36: I am +1 on this.

pepper-sync/src/sync.rs
```diff
@@ -603,10 +595,9 @@ where
 ///
 /// The constants are:
 ///   (1) MAX_REORG_ALLOWANCE:
-///       * the number of blocks AHEAD of proxy_reported_chain_height allowed
-///   (2) Sapling Epoch Height:
+///       * the maximum number of blocks the wallet can truncate during re-org detection
+///   (2) Sapling Activation Height:
```
zancas, line 599: Nice.

pepper-sync/src/sync.rs
```diff
@@ -630,273 +620,35 @@ where
...
-                }))
-                .create_mock_wallet();
-            let res =
-                checked_wallet_height(&mut test_wallet, BlockHeight::from_u32(1), &LOCAL_NETWORK);
-            assert!(matches!(
-                res,
-                Err(SyncError::WalletError(
-                    crate::mocks::MockWalletError::AnErrorVariant(ref s)
-                )) if s == test_error
+        } else if birthday < sapling_activation_height {
```
zancas, line 644: I am +1 on this.

pepper-sync/src/sync.rs
```diff
@@ -630,273 +620,35 @@ where
...
+            return Err(SyncError::BirthdayBelowSapling(
+                u32::from(birthday),
+                u32::from(sapling_activation_height),
             ));
         }
 
-        mod last_known_chain_height {
```
zancas, line 700: Why are all these tests removed?  Are they included elsewhere?

pepper-sync/src/sync.rs
```diff
@@ -2021,21 +1783,6 @@ where
     Ok(())
 }
 
-/// Compares the wallet birthday to sapling activation height and returns the highest block height.
```
zancas, line 2024: Yay!

pepper-sync/src/sync.rs
```diff
@@ -2140,3 +1887,244 @@ fn max_nullifier_map_size(performance_level: PerformanceLevel) -> Option<usize>
         PerformanceLevel::Maximum => None,
     }
 }
+
+#[cfg(test)]
+mod test {
```
zancas, line 1892: Ahh..  OK.

pepper-sync/src/sync.rs
```diff
@@ -2140,3 +1887,244 @@ fn max_nullifier_map_size(performance_level: PerformanceLevel) -> Option<usize>
...
+                    &LOCAL_NETWORK,
+                );
+                assert!(matches!(
+                    res,
+                    Err(SyncError::WalletError(
+                        crate::mocks::MockWalletError::AnErrorVariant(ref s)
+                    )) if s == test_error
+                ));
+            }
+            #[ignore = "in progress"]
```
zancas, line 2062: I will update these.

zancas (changes requested):
I left non-blocking comments, with 1 exception, why is truncate_scan_ranges a free fn that takes a &mut instead of a method of SyncState?

Everything else looks good to me.  I like the renames, and erroring below Sapling.

Oscar-Pepper:
> I left non-blocking comments, with 1 exception, why is truncate_scan_ranges a free fn that takes a &mut instead of a method of SyncState?
> 
> Everything else looks good to me.  I like the renames, and erroring below Sapling.

I'm not sure which changes have been requested for merge. Regarding the truncate being a method on sync state, there was a bunch of trade-offs I considered with this and decided a functional architecture worked best for the library. This allows us to keep SyncState succinct in the 'wallet' module (public APi for use in consumer wallet structs) and keep all the state logic cleanly separated. I also think it would be confusing to have a bunch of SyncState impl spread across the library... I like them to be directly under the struct or in clean submodules so consumers can clearly see all methods/functionality of a type. There is also the fact that many state fns  don't just operate on the SyncState so I think there is more consistency this way. There are possibly other reasons but it's a while since I decided to go this way. If you have a counter argument for a benefit of object-oriented I have missed we can discuss

zancas:
I think the reasoning you offer here is very useful.   I haven't carefully considered tradeoffs in this space, but I think it's very worthwhile to capture our thinking somewhere.

zancas (approved):

PR #2216, opened 2026-02-28
fix zingo common dep

Oscar-Pepper:
zingo_common_components = { git = "https://github.com/zingolabs/zingo-common.git", version = "0.2" }

this means that in development (most cases really) it will use the latest dev of the git repo (if "branch" not specified just uses the default which is dev)... this means that its trying to use version 0.3+ which is current dev where we want it to stil use version 0.2. it *only* uses the crates.io version 0.2 when publishing to crates.io


juanky201271 (approved):
great, man.

PR #2231, opened 2026-03-06
make for_test/regtest code optional

zancas:


Oscar-Pepper (commented):
i think it is important we can test production binaries in regtest so I have removed all regtest features in https://github.com/zingolabs/zingolib/pull/2187. the reason we needed this feature in the first place was becasue we baked in a regtest launcher into zingo-cli which therefore required a zcash-local-net dep. this has been removed already so IMO we should not go this direction

zancas:
Obsolete.

PR #2241, opened 2026-03-12
attempt to port standalone wallet code

zancas:


zingolib_testutils/src/scenarios.rs
```diff
@@ -198,7 +196,7 @@ impl ClientBuilder {
         //! A "faucet" is a lightclient that receives mining rewards
         self.build_client(
             seeds::ABANDON_ART_SEED.to_string(),
-            1,
+            0,
```
zancas: I'm not clear on why this changed.

zancas: It appears to make tests fail, so I am changing it back.

zancas:
Superseded: Offline mode shipped — the `--offline` session flag, the Indexerless library state, and the expiry-retarget rule (ADRs 0001 and 0008) cover what this port attempted, and "standalone wallet" is a retired name for the concept (see the glossary).

PR #2259, opened 2026-03-14
feat: replace `String` with `IndexerInfo`

### LINKED ISSUE #2257: `LightClient::do_info` returns JSON-formatted `String`
Returning a `String` value that is JSON-formatted at the library level seems to be wrong. It should be typed, as consumers will call that method from Rust.

Also, a potential better name could be `get_indexer_info` instead.

dorianvp:
Fixes #2257.


zingolib/src/lightclient.rs
```diff
@@ -187,25 +200,25 @@ impl LightClient {
...
-                    "server_uri" => self.indexer_uri().to_string(),
-                    "vendor" => i.vendor,
-                    "taddr_support" => i.taddr_support,
-                    "chain_name" => i.chain_name,
-                    "sapling_activation_height" => i.sapling_activation_height,
-                    "consensus_branch_id" => i.consensus_branch_id,
-                    "latest_block_height" => i.block_height
-                };
-                o.pretty(2)
+                let branch_id: u32 = i.consensus_branch_id.parse().unwrap();
```
Oscar-Pepper, line 207: can you replace with an expect with a message to explain why this should not panic?

zingolib/src/lightclient.rs
```diff
@@ -60,6 +60,19 @@ pub struct LightClient {
...
+/// Contains information about the indexer this lighclient is connected to.
+pub struct IndexerInfo {
+    pub version: String,
+    pub git_commit: String,
+    pub indexer_uri: http::Uri,
+    pub vendor: String,
+    pub supports_transparent: bool,
+    pub chain_name: String,
+    pub sapling_activation_height: BlockHeight,
+    pub consensus_branch_id: BranchId,
```
Oscar-Pepper, line 72: work is undergoing to remove zcash_protocol from the public API. I think this is ok for now and we can revisit this later when the replacement internal types and modules are in place.

zingolib/src/lightclient.rs
```diff
@@ -187,25 +200,25 @@ impl LightClient {
...
+                let branch_id: u32 = i.consensus_branch_id.parse().unwrap();
+
+                Ok(IndexerInfo {
+                    version: i.version,
+                    git_commit: i.git_commit,
+                    indexer_uri: self.indexer_uri(),
+                    vendor: i.vendor,
+                    supports_transparent: i.taddr_support,
+                    chain_name: i.chain_name,
+                    sapling_activation_height: i.sapling_activation_height.try_into().unwrap(),
```
Oscar-Pepper, line 216: can we replace unwrap with atleast an expect or preferably returning an error now this fn returns a result?

zingolib/src/lightclient.rs
```diff
@@ -187,25 +200,25 @@ impl LightClient {
...
+                    consensus_branch_id: BranchId::try_from(branch_id).unwrap(),
```
Oscar-Pepper, line 217: same as above

zingolib/src/lightclient.rs
```diff
@@ -187,25 +200,25 @@ impl LightClient {
...
+                    latest_block_height: i.block_height.try_into().unwrap(),
```
Oscar-Pepper, line 218: same as above

Oscar-Pepper (changes requested):

zingolib/src/lightclient.rs
```diff
@@ -60,6 +60,19 @@ pub struct LightClient {
...
+    pub version: String,
+    pub git_commit: String,
+    pub indexer_uri: http::Uri,
+    pub vendor: String,
+    pub supports_transparent: bool,
+    pub chain_name: String,
+    pub sapling_activation_height: BlockHeight,
+    pub consensus_branch_id: BranchId,
+    pub latest_block_height: BlockHeight,
+}
```
Oscar-Pepper, line 74: can we add an impl:
impl From<IndexerInfo> for json::JsonValue {
...
}

zingo-cli/src/commands.rs
```diff
@@ -570,7 +570,23 @@ impl Command for InfoCommand {
...
+                    "version" => info.version,
+                    "git_commit" => info.git_commit,
+                    "server_uri" => lightclient.indexer_uri().to_string(),
+                    "vendor" => info.vendor,
+                    "taddr_support" => info.supports_transparent,
+                    "chain_name" => info.chain_name,
+                    "sapling_activation_height" => info.sapling_activation_height.to_string(),
+                    "consensus_branch_id" => u32::from(info.consensus_branch_id).to_string(),
+                    "latest_block_height" => info.latest_block_height.to_string()
+                }
```
Oscar-Pepper, line 585: can we move this code into the impl i mentioned below and call it here?

Oscar-Pepper: i.e. Ok(info) => json::JsonValue::from(info).pretty(2)

Oscar-Pepper (changes requested):

zancas:
Superceded by #2464

PR #2261, opened 2026-03-18
Zingo config rework pt2

Oscar-Pepper:
preliminary work for fixing #2269. continued in PR #2266 

- removes `zcash_protocol` from public API by consealing Parameters impl on ChainType
- consistent `chain_type` naming throughout codebase
- moves non-config consts and fns out of config module and renames to reflect move from LWD to indexer
- `construct_lightwalletd_uri` now returns result for handling uri errors
- new wallet serialization version 40. `ChainType` serialized as u8 instead of string to decouple from fmt::Display and reduce bytes stored.


zancas (approved):
I haven't checked with consumers about how they are affected by pub interface changes.

@juanky201271 ?

PR #2263, opened 2026-03-18
Update zingo-netutils to update Zebra dependencies

pacu:
Fixes:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ zcash_protocol = "0.7" # public API (TODO: remove from public API)
 zcash_transparent = "0.6"
 
 # Zingo-common
-zingo-netutils = "2.0.2"
+zingo-netutils = { git = "https://github.com/zingolabs/zingo-common.git", rev = "54309fbb1a0c3d791b457667582c9c4a4da6c4bc" }
 zingo_common_components = "0.3.0"
 
 # Parallel processing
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -44,8 +44,8 @@ zcash_protocol = "0.7" # public API (TODO: remove from public API)
 zcash_transparent = "0.6"
 
 # Zingo-common
-zingo-netutils = "2.0.2"
-zingo_common_components = "0.3.0"
```
Oscar-Pepper, line 48: no code has changed between latest dev and this crates.io release. please can we stick to crates if possible?

Cargo.toml
```diff
@@ -44,8 +44,8 @@ zcash_protocol = "0.7" # public API (TODO: remove from public API)
 zcash_transparent = "0.6"
 
 # Zingo-common
-zingo-netutils = "2.0.2"
```
Oscar-Pepper, line 47: this crates release if from the git commit you have specified. can you explain why we are moving away from the crates release here?

Cargo.toml
```diff
@@ -117,8 +117,8 @@ zingo-status = "0.2.0"
 pepper-sync = { path = "pepper-sync" } # NOTE: for development between releases
 # pepper-sync = "0.2.0"
 zingolib = { path = "zingolib" }
-zcash_local_net = { git = "https://github.com/zingolabs/infrastructure.git", tag = "zcash_local_net_v0.4.0" }
-zingo_test_vectors = { git = "https://github.com/zingolabs/infrastructure.git", tag = "zcash_local_net_v0.4.0" }
+zcash_local_net = { git = "https://github.com/zingolabs/infrastructure.git", rev = "87cb5ae58204ea8d566cc5f8373cd5765c3360b2" }
+zingo_test_vectors = { git = "https://github.com/zingolabs/infrastructure.git", rev = "87cb5ae58204ea8d566cc5f8373cd5765c3360b2"}
```
Oscar-Pepper, line 121: i need some help understanding the issue this is solving. which type has caused dep conflicts? I have checked and there is no zebra types as far as i can see in the public API of zcash_local_net v0.4.0 other than the out-of-date re-export that is no longer needed. as long as this re-export is not being used there should be no be conflicts related to zebra right? also zingo_test_vectors has zero zebra deps

Oscar-Pepper (changes requested):
Can we get some more detail on what this PR aims to solve please? maybe build errors etc. as im not sure how this can fix dependency conflicts due to the public API work that has already happened in these crates

Cargo.toml
pacu: The problem is that Zaino ends up having two versions of the same thing and doesn't build. :S

PR #2266, opened 2026-03-22
Zingo config rework pt3

### LINKED ISSUE #2269: config modules public API needs to be restricted and config types need reworking
the config module issues:
- the public API includes external types from crates such as zcash_protocol which cause dependency conflicts and zingolib major semver version bumps when these crates change their public API.
- configs hold a lot of unnecessary data in certain cases. for example, if constructing a lightclient by reading a wallet from file, there is no need for wallet settings as they will be read from file. but these are still required to create the lightclient.
- config is held by the lightclient and is overtasked. for example, there are methods such as `backup_existing_wallet` which should be implemented on the lightclient. the config should have a single purpose of constructing a lightclient and all data should be either used during construction or stored in the lightclient or lightwallet fields. The config should not be held by the lightclient and accessed for data after it has been constructed.
- the lightclient and lightwallet configs are not the single source of truth needed to construct these types. there are many constructors with many dead fields which makes the library convoluted and difficult to introduce toml based invocations

Oscar-Pepper:
Fixes: #2269 
ontop of #2261

- renames `ZingoConfig` to `ClientConfig`
- new `WalletConfig` struct in config module replaced function of `WalletBase` which is no longer public and is resolved from the `WalletConfig`
- lightclient now has a single constructor `new` which only takes a `ClientConfig`. all data is encapsulated in the config about how the lightclient is to be constructed with no dead fields i.e. no data needs to be supplied that is not used.
- lightwallet now has a single constructor `new` except for the `read` impl for reading from file. `new` takes the new `WalletConfig` enum which encapsulates all variants of wallet creation with no dead fields.
- lightclient no longer holds the config. the config is used for lightclient construction and the data is stored in relevant fields in lightclient or lightwallet
- finishes stabilising the config modules public API with no external types except for `http::Uri` which appears to be a very stable type.

### DIFF
--- a/darkside-tests/src/chain_generics.rs
+++ b/darkside-tests/src/chain_generics.rs
@@ -39,6 +39,7 @@ pub(crate) mod conduct_chain {
     use incrementalmerkletree::frontier::CommitmentTree;
     use orchard::tree::MerkleHashOrchard;
 
+    use zcash_protocol::consensus::BlockHeight;
     use zingo_netutils::Indexer as _;
 
     use zingolib::lightclient::LightClient;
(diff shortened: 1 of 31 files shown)


zancas:
Needs conflict resolution.

zancas (approved):
I have some intention to encapsulate the Indexer Uri in a proper type as follow up work.

zancas:
I ran local tests and they passed.

Oscar-Pepper:
> I have some intention to encapsulate the Indexer Uri in a proper type as follow up work.

URIs could just Strings in public API and then convert to http::Uri immediately

PR #2268, opened 2026-03-23
Persist the newest shard-tree checkpoint window

### LINKED ISSUE #2267: Wallet save corrupts shard tree state — consecutive sends always fail
## Bug

After a wallet accumulates more than `MAX_REORG_ALLOWANCE` checkpoints in memory, saving and reloading can drop the **recent Sapling checkpoint window** while keeping Orchard current. The next send from a freshly reopened wallet then fails with:

```
Sapling shard store checkpoint not found at anchor height XXXXXXX
```

This is reproducible with unmodified `zingo-cli` and real wallet files on testnet.

## Correct Root Cause

The problem is **not** `mem::replace` corrupting the in-memory tree.

The actual bug is in `pepper-sync/src/wallet/serialization.rs` during `write_shardtree()`:

```rust
store.with_checkpoints(MAX_REORG_ALLOWANCE as usize, |checkpoint_id, checkpoint| {
    checkpoints.push((*checkpoint_id, checkpoint.clone()));
    Ok(())
})
```

`with_checkpoints()` iterates in **ascending checkpoint ID order** (oldest first).
So when a store contains more than `MAX_REORG_ALLOWANCE` checkpoints, serialization writes the **oldest** 100 checkpoints, not the **newest** 100 checkpoints needed for the current anchor/reorg window.

On reload, the wallet gets an out-of-date Sapling checkpoint set from disk.

## What I Observed Live

Using instrumented `zingo-cli` against real testnet wallets:

- Before sync/rebuild, a reopened wallet loaded with ranges like:
  - `sapling: 0..3913376`
  - `orchard: 3913365..3913466`
- A live sync rebuilt Sapling in memory and send #1 succeeded.
- After save + reopen, Sapling again fell back to the old serialized window while Orchard remained current.
- Send #2 then failed with:

```
Sapling shard store checkpoint not found at anchor height 3913465
```

After changing serialization to keep the **newest** checkpoint window, I reran the same live flow and send #1, send #2, and send #3 all succeeded.

## Minimal Fix

Serialize all checkpoints, then keep only the newest `MAX_REORG_ALLOWANCE` before writing:

```rust
let checkpoint_count = store.checkpoint_count().expect("Infallible");
store
    .with_checkpoints(checkpoint_count, |checkpoint_id, checkpoint| {
        checkpoints.push((*checkpoint_id, checkpoint.clone()));
        Ok(())
    })
    .expect("Infallible");

if checkpoints.len() > MAX_REORG_ALLOWANCE as usize {
    let keep_from = checkpoints.len() - MAX_REORG_ALLOWANCE as usize;
    checkpoints.drain(..keep_from);
}
```

## Regression Test

I added a focused round-trip test that creates 151 checkpoints, serializes + deserializes the shard trees, and asserts the reloaded range is `51..150`, not `0..99`.

## Reproduction

```bash
# resync / save a wallet with enough Orchard funds
zingo-cli --data-dir /path/to/wallet --server https://testnet.zec.rocks:443 -c testnet --waitsync balance

# send #1 succeeds after live sync rebuilt Sapling in memory
zingo-cli --data-dir /path/to/wallet --server https://testnet.zec.rocks:443 -c testnet --waitsync quicksend <addr> 1000 "send-1"

# reopen from disk and send #2
# before the fix: fails with missing Sapling checkpoint
# after the fix: succeeds
zingo-cli --data-dir /path/to/wallet --server https://testnet.zec.rocks:443 -c testnet --waitsync quicksend <addr> 1000 "send-2"
```

## Impact

- Consecutive sends from reopened wallets can fail even though balances look correct.
- A rescan or fresh sync can mask the bug temporarily by rebuilding Sapling in memory.
- The failure reappears after save/load because the wrong checkpoint window is persisted.

betterclever:
## Summary

This fixes wallet shard-tree persistence so save/load keeps the newest reorg-safe checkpoint window instead of the oldest one.

## Root Cause

`write_shardtree()` was serializing checkpoints with:

```rust
store.with_checkpoints(MAX_REORG_ALLOWANCE as usize, ...)
```

`with_checkpoints()` iterates in ascending checkpoint ID order, so once a store had more than `MAX_REORG_ALLOWANCE` checkpoints, the wallet file kept the **oldest** 100 checkpoints instead of the **newest** 100.

That let recently rebuilt Sapling checkpoints disappear across save/load, which could cause consecutive sends from reopened wallets to fail with:

```text
Sapling shard store checkpoint not found at anchor height ...
```

## Fix

- collect all checkpoints during serialization
- trim to the newest `MAX_REORG_ALLOWANCE`
- keep the persisted checkpoint window aligned with current anchor/reorg needs

## Test

Added a round-trip regression test that creates 151 checkpoints and verifies the reloaded range is `51..150`, not `0..99`.

## Validation

- `cargo test -p pepper-sync shardtree_roundtrip_keeps_newest_checkpoints -- --nocapture`
- live testnet repro with instrumented `zingo-cli`
  - before fix: the second send after reopen failed with a missing Sapling checkpoint
  - after fix: the first, second, and third sends after reopen all succeeded

Closes #2267.

### DIFF
--- a/pepper-sync/src/wallet/serialization.rs
+++ b/pepper-sync/src/wallet/serialization.rs
@@ -999,7 +999,10 @@ impl ShardTrees {
         const SHARD_HEIGHT: u8,
     >(
         mut reader: R,
-    ) -> std::io::Result<ShardTree<MemoryShardStore<H, C>, DEPTH, SHARD_HEIGHT>> {
+    ) -> std::io::Result<ShardTree<MemoryShardStore<H, C>, DEPTH, SHARD_HEIGHT>>
+    where
+        u32: From<C>,
+    {
         let shards = Vector::read(&mut reader, |r| {
             let level = incrementalmerkletree::Level::from(r.read_u8()?);
             let index = r.read_u64::<LittleEndian>()?;
@@ -1044,7 +1047,6 @@ impl ShardTrees {
                 .expect("Infallible");
         }
         store.put_cap(read_shard(reader)?).expect("Infallible");
-
         Ok(shardtree::ShardTree::new(
             store,
             MAX_REORG_ALLOWANCE as usize,
@@ -1139,12 +1141,17 @@ impl ShardTrees {
 
         // Write checkpoints
         let mut checkpoints = Vec::new();
+        let checkpoint_count = store.checkpoint_count().expect("Infallible");
         store
-            .with_checkpoints(MAX_REORG_ALLOWANCE as usize, |checkpoint_id, checkpoint| {
+            .with_checkpoints(checkpoint_count, |checkpoint_id, checkpoint| {
                 checkpoints.push((*checkpoint_id, checkpoint.clone()));
                 Ok(())
             })
             .expect("Infallible");
+        if checkpoints.len() > MAX_REORG_ALLOWANCE as usize {
+            let keep_from = checkpoints.len() - MAX_REORG_ALLOWANCE as usize;
+            checkpoints.drain(..keep_from);
+        }
         write_with_error_handling!(write_checkpoints, checkpoints);
 
         // Write cap
@@ -1154,5 +1161,67 @@ impl ShardTrees {
         *shardtree = shardtree::ShardTree::new(store, MAX_REORG_ALLOWANCE as usize);
 
         Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn shardtree_roundtrip_keeps_newest_checkpoints() {
+        let mut shard_trees = ShardTrees::new();
+
+        for height in 1..=150 {
+            let height = BlockHeight::from_u32(height);
+            shard_trees
+                .sapling
+                .store_mut()
+                .add_checkpoint(height, Checkpoint::from_parts(TreeState::Empty, BTreeSet::new()))
+                .expect("infallible");
+            shard_trees
+                .orchard
+                .store_mut()
+                .add_checkpoint(height, Checkpoint::from_parts(TreeState::Empty, BTreeSet::new()))
+                .expect("infallible");
+        }
+
+        let mut bytes = Vec::new();
+        shard_trees.write(&mut bytes).expect("write should succeed");
+        let roundtripped = ShardTrees::read(bytes.as_slice()).expect("read should succeed");
+
+        let sapling_store = roundtripped.sapling.store();
+        let orchard_store = roundtripped.orchard.store();
+
+        assert_eq!(sapling_store.checkpoint_count().expect("infallible"), 100);
+        assert_eq!(orchard_store.checkpoint_count().expect("infallible"), 100);
+        assert_eq!(
+            sapling_store.min_checkpoint_id().expect("infallible"),
+            Some(BlockHeight::from_u32(51))
+        );
+        assert_eq!(
+            sapling_store.max_checkpoint_id().expect("infallible"),
+            Some(BlockHeight::from_u32(150))
+        );
+        assert_eq!(
+            orchard_store.min_checkpoint_id().expect("infallible"),
+            Some(BlockHeight::from_u32(51))
+        );
+        assert_eq!(
+            orchard_store.max_checkpoint_id().expect("infallible"),
+            Some(BlockHeight::from_u32(150))
+        );
+        assert!(
+            sapling_store
+                .get_checkpoint(&BlockHeight::from_u32(149))
+                .expect("infallible")
+                .is_some()
+        );
+        assert!(
+            sapling_store
+                .get_checkpoint(&BlockHeight::from_u32(50))
+                .expect("infallible")
+                .is_none()
+        );
     }
 }
--- a/pepper-sync/src/wallet/serialization.rs
+++ b/pepper-sync/src/wallet/serialization.rs
@@ -999,10 +999,7 @@ impl ShardTrees {
         const SHARD_HEIGHT: u8,
     >(
         mut reader: R,
-    ) -> std::io::Result<ShardTree<MemoryShardStore<H, C>, DEPTH, SHARD_HEIGHT>>
-    where
-        u32: From<C>,
-    {
+    ) -> std::io::Result<ShardTree<MemoryShardStore<H, C>, DEPTH, SHARD_HEIGHT>> {
         let shards = Vector::read(&mut reader, |r| {
             let level = incrementalmerkletree::Level::from(r.read_u8()?);
             let index = r.read_u64::<LittleEndian>()?;
@@ -1047,6 +1044,7 @@ impl ShardTrees {
                 .expect("Infallible");
         }
         store.put_cap(read_shard(reader)?).expect("Infallible");
+
         Ok(shardtree::ShardTree::new(
             store,
             MAX_REORG_ALLOWANCE as usize,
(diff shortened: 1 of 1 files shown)


Oscar-Pepper:
hi @betterclever can you create an issue detailing exactly what error has occured for this work to be carried out? the shard store is configured to only ever store a maximum of MAX_REORG_ALLOWANCE checkpoints and is pruned after every checkpoint addition... this means that the wallet does not persist the *first* X checkpoints, it persists all the checkpoints.

Oscar-Pepper:
i have found https://github.com/zingolabs/zingolib/issues/2267 i am reading through it now.

Oscar-Pepper:
thanks for your contribution, the deeper question we need to investigate is why there are more than MAX_REORG_ALLOWANCE checkpoints in the first place? after looking at https://github.com/zingolabs/zingolib/blob/dev/pepper-sync/src/wallet/traits.rs#L237 there is an edge case where the range scanned may not have and orchard or sapling note commitments and will not need to insert any trees, skipping pruning. Unfortunately the pruning method is private so the solution you have proposed is an alternative way to prune. Or an easier fix may be to just increase the limit in `store.with_checkpoints` to a much higher limit i.e. MAX_REORG_ALLOWANCE + 1000 and we just persist the extra checkpoints until it is pruned next time a range is scanned with sapling or orchard note commitments

betterclever:
@Oscar-Pepper I've instrumented the full save/load path and can now show the exact drift.

## Root cause: `with_checkpoints(100)` keeps oldest, not newest

`MemoryShardStore::with_checkpoints` at `shardtree-0.6.1/src/store/memory.rs:130`:

```rust
fn with_checkpoints<F>(&mut self, limit: usize, mut callback: F) -> ...  {
    for (cid, checkpoint) in self.checkpoints.iter().take(limit) {
```

`self.checkpoints` is a `BTreeMap` — `.iter()` is ascending. `.take(limit)` takes the **first** (oldest) entries.

## Instrumented evidence

After rescan + sync to block 3914320:

```
[SERIALIZE] === Sapling ===
[SERIALIZE] total_in_tree=111 range=0..3914320 | serialized=100 range=0..3914309 | DROPPED=11
[SERIALIZE] === Orchard ===
[SERIALIZE] total_in_tree=100 range=3914221..3914320 | serialized=100 range=3914221..3914320 | DROPPED=0
```

Sapling has **111** checkpoints (including one at height 0), but serialization keeps the **oldest 100** (0..3914309), dropping the **11 newest** (3914310..3914320).

Orchard has exactly 100, all recent — nothing dropped.

After reload: Orchard window is 3914221..3914320, Sapling window is 0..3914309. The anchor at ~3914317 exists in Orchard but not Sapling → `CheckpointNotFound`.

## Why >100 checkpoints exist

You asked why there are more than MAX_REORG_ALLOWANCE. `update_shard_trees` (traits.rs:277-305) correctly creates checkpoints for both pools at every height in the range. But `ShardTree::checkpoint()` only prunes when `insert_tree` is called (which triggers `prune_excess_checkpoints`). For Sapling, when blocks have zero Sapling note commitments, `insert_tree` receives an empty located tree — but `add_checkpoint` still adds the copy-forward checkpoint. Over time, Sapling accumulates >100 checkpoints without triggering pruning.

The fix in this PR correctly keeps the **newest** checkpoints by skipping `total - limit` entries before collecting.

Oscar-Pepper:
> @Oscar-Pepper I've instrumented the full save/load path and can now show the exact drift.
> 
> ## Root cause: `with_checkpoints(100)` keeps oldest, not newest
> `MemoryShardStore::with_checkpoints` at `shardtree-0.6.1/src/store/memory.rs:130`:
> 
> ```rust
> fn with_checkpoints<F>(&mut self, limit: usize, mut callback: F) -> ...  {
>     for (cid, checkpoint) in self.checkpoints.iter().take(limit) {
> ```
> 
> `self.checkpoints` is a `BTreeMap` — `.iter()` is ascending. `.take(limit)` takes the **first** (oldest) entries.
> 
> ## Instrumented evidence
> After rescan + sync to block 3914320:
> 
> ```
> [SERIALIZE] === Sapling ===
> [SERIALIZE] total_in_tree=111 range=0..3914320 | serialized=100 range=0..3914309 | DROPPED=11
> [SERIALIZE] === Orchard ===
> [SERIALIZE] total_in_tree=100 range=3914221..3914320 | serialized=100 range=3914221..3914320 | DROPPED=0
> ```
> 
> Sapling has **111** checkpoints (including one at height 0), but serialization keeps the **oldest 100** (0..3914309), dropping the **11 newest** (3914310..3914320).
> 
> Orchard has exactly 100, all recent — nothing dropped.
> 
> After reload: Orchard window is 3914221..3914320, Sapling window is 0..3914309. The anchor at ~3914317 exists in Orchard but not Sapling → `CheckpointNotFound`.
> 
> ## Why >100 checkpoints exist
> You asked why there are more than MAX_REORG_ALLOWANCE. `update_shard_trees` (traits.rs:277-305) correctly creates checkpoints for both pools at every height in the range. But `ShardTree::checkpoint()` only prunes when `insert_tree` is called (which triggers `prune_excess_checkpoints`). For Sapling, when blocks have zero Sapling note commitments, `insert_tree` receives an empty located tree — but `add_checkpoint` still adds the copy-forward checkpoint. Over time, Sapling accumulates >100 checkpoints without triggering pruning.
> 
> The fix in this PR correctly keeps the **newest** checkpoints by skipping `total - limit` entries before collecting.

thanks for showing these details, clearly the ranges with no sapling or orchard note commitments are not triggering pruning as i described above. I am happy to accept this fix instead of the one i proposed. thanks for your contribution!

pepper-sync/src/wallet/serialization.rs
```diff
@@ -999,7 +999,10 @@ impl ShardTrees {
         const SHARD_HEIGHT: u8,
     >(
         mut reader: R,
-    ) -> std::io::Result<ShardTree<MemoryShardStore<H, C>, DEPTH, SHARD_HEIGHT>> {
```
Oscar-Pepper, line 1002: please can you remove unnecessary changes to the read impl. why does this fn need this constraint on C?

pepper-sync/src/wallet/serialization.rs
```diff
@@ -1044,7 +1047,6 @@ impl ShardTrees {
                 .expect("Infallible");
         }
         store.put_cap(read_shard(reader)?).expect("Infallible");
-
```
Oscar-Pepper, line 1047: same as above. this formatting is consistent across the library

Oscar-Pepper (changes requested):
great work on debugging and fixing this issue. thanks again for your contribution! i would like to remove the changes on the read impl but the write impl likes good to go.

Oscar-Pepper:
oh @betterclever it also seems CI is failing on formatting. please run `cargo fmt` too

betterclever:
@Oscar-Pepper Done. Should I remove the test or keep it there?

Oscar-Pepper:
> @Oscar-Pepper Done. Should I remove the test or keep it there?

no thats great thanks

Oscar-Pepper (approved):

Oscar-Pepper:
we need to cut a bug-fix release and I want to include this fix also. I hope you dont mind but im going to port this work to `stable` also.

betterclever:
@Oscar-Pepper Yup that's great. Thank you for the library again. I'm using it for https://github.com/betterclever/zimppy 
Giving you a shoutout there.

Issue #2244, opened 2026-03-12
Possible bug in truncate_scan_ranges(): partition_point predicate looks reversed


Tomas-M:
While I was working on something else with the code using ChatGPT, the AI pointed out a possible error in `pepper-sync/src/sync/state.rs`
and insisted there is a bug
```
let truncated_scan_ranges = sync_state.scan_ranges[..sync_state
    .scan_ranges()
    .partition_point(|range| range.block_range().start > truncate_height)]
    .to_vec();
```

ChatGPT claims the issue is that `partition_point()` expects the predicate results to look like `true true true false false ...`, but for scan ranges ordered by increasing start height, it claims this predicate seems to produce the reverse: `false false false true true`.

It insists this should likely be changed to something like `start <= truncate_height` instead

Could you please verify whether this is actually a bug?

Oscar-Pepper:
You are absolutely right. I have added a failing test and then fixed it here https://github.com/zingolabs/zingolib/pull/2271. Really appreciate you finding this one, this should have been tested. I will look into adding more coverage to at least the state module which contains the majority of the intricate state management of the sync engine and is easy to mock for testing. Thanks!

Tomas-M:
Thank you very much, I am closing this issue.

PR #2271, opened 2026-03-24
write failing test and fix truncate_scan_ranges fn

Oscar-Pepper:
Fixes: https://github.com/zingolabs/zingolib/issues/2244

fixes bug in scan range truncation where the predicate is reversed resulting in clearing the scan ranges instead of truncating to a given height

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9473,7 +9473,7 @@ dependencies = [
 
 [[package]]
 name = "zingolib"
-version = "3.0.0-rc.1"
+version = "3.0.0"
 dependencies = [
  "append-only-vec",
  "bech32",
--- a/pepper-sync/src/sync/state.rs
+++ b/pepper-sync/src/sync/state.rs
@@ -177,7 +177,7 @@ pub(super) fn truncate_scan_ranges(truncate_height: BlockHeight, sync_state: &mu
 
     let truncated_scan_ranges = sync_state.scan_ranges[..sync_state
         .scan_ranges()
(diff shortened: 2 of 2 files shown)


Cargo.lock
```diff
@@ -9473,7 +9473,7 @@ dependencies = [
 
 [[package]]
 name = "zingolib"
-version = "3.0.0-rc.1"
```
zancas, line 9476: Is this the correct version bump?  Should it not be to 3.0.1?

pepper-sync/src/sync/state.rs
```diff
@@ -1067,3 +1067,30 @@ pub(super) fn update_found_note_shard_priority(
         );
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn truncate_scan_ranges() {
```
zancas, line 1076: This seems like a good start, but I am wondering if this test can be expanded, or if we can have a suite of tests, where different boundary conditions are exercised for this fn?

For example, this test passes "250", why not add a suite of tests that hit a range of truncation points?   0,1,98,99,100,101

300 before truncation
300 after truncation

And then verify in separate assert statements with descriptive failure messages that the behavior is expected?

dorianvp (approved):

zancas (changes requested):

Cargo.lock
dorianvp: That `*-rc.1` I think is invalid semver, and that's why cargo removed it automatically.

zancas (approved):

PR #2276, opened 2026-03-24
Sort Cargo.toml from most specific to least specific dependency groups.

fluidvanadium:
It is easier to notice the highest-velocity dependencies if they are at the top!


fluidvanadium:
Meant to open into crosslink!

PR #2277, opened 2026-03-24
Sort Cargo.toml from most specific to least specific dependency groups.

fluidvanadium:
It is easier to notice the highest-velocity dependencies if they are at the top!


Oscar-Pepper (dismissed):
great work!

Oscar-Pepper:
held up on CI fail for doc tests

zancas (approved):

Issue #2286, opened 2026-03-26
zingo-cli needs to support offline mode


zancas:

ala-mode:
Possibly related: https://github.com/zingolabs/zingolib/issues/2226

zancas:
Implemented: zingo-cli has the `--offline` flag (zingo-cli/src/lib.rs:79) committing the session to Offline mode at argument-parse time, with the ratified semantics — no Indexer ever configured, local operations available, sync/transmission/mempool refused typed, and Offline-mode calculate retargeting expiry per ADR 0008.

PR #2288, opened 2026-03-26
Make off- vs on- line mode enum

### LINKED ISSUE #2287: zingo-cli can be run in two mutually exclusive mode online and offline
This MUST be expressed as a Rust enum.

zancas:
Fixes: #2287
Increments on: #2286 

Included in and superceded by #2294

### DIFF
--- a/zingo-cli/src/lib.rs
+++ b/zingo-cli/src/lib.rs
@@ -353,10 +353,31 @@ fn get_mode_of_operation(matches: &clap::ArgMatches) -> ModeOfOperation {
     }
 }
 
+/// Whether the CLI communicates with a remote indexer or operates locally.
+///
+/// Currently always [`Online`](CommunicationMode::Online). The [`Offline`](CommunicationMode::Offline)
+/// variant exists so that offline wallet support has a clean place to land.
+#[derive(Debug, PartialEq)]
+enum CommunicationMode {
+    /// Connected to a remote indexer for sync, send, etc.
+    Online,
+    /// Operating without network access — local-only commands.
+    Offline,
+}
+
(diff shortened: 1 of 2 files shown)


zingo-cli/src/lib.rs
```diff
@@ -369,7 +396,11 @@ pub(crate) struct ConfigTemplate {
 }
 
 impl ConfigTemplate {
-    fn fill(mode: ModeOfOperation, matches: clap::ArgMatches) -> Result<Self, String> {
+    fn fill(
+        mode: ModeOfOperation,
+        communication_mode: CommunicationMode,
```
Oscar-Pepper, line 401: if this is set by an `--offline` flag then shouldnt we be handling the creation of this enum inside `fill` using the `matches` clap data?

zingo-cli/src/lib.rs
```diff
@@ -626,7 +658,9 @@ pub fn help_output(matches: &clap::ArgMatches) -> Option<String> {
 /// The caller (the binary entry point) is responsible for parsing arguments,
 /// handling the help short-circuit, and process-level setup.
 pub fn run_cli(matches: clap::ArgMatches) {
-    match ConfigTemplate::fill(get_mode_of_operation(&matches), matches) {
+    let mode = get_mode_of_operation(&matches);
```
Oscar-Pepper, line 661: shouldnt this be inside `fill` method which already takes `matches` which is what this fn needs to resolve the enum?

Oscar-Pepper (commented):
needs updating to latest config rework changes to pass CI. also left 2 related comments to consider

Issue #2260, opened 2026-03-17
zingocli sync stops regularly, needing manual restart


AloeareV:
Sync is repeatedly stopping with the error `Sync error: Error: server error. server request failed. code: 'Deadline expired before operation could complete', message: "fetch TreeState reply timeout"`. The error message isn't always identical (it isn't always a TreeState reply, just most commonly. I've also seen a fetch Transaction reply timeout). Sometimes the error shows up as a tracing error, but most of the time the error appears when I attempt to run a command (ususally sync status). This error stops sync, and I need to invoke sync run to start it again. 

(edit: This is on a sync of an old pre-peppersync wallet file, one that triggers sync from beginning.)

arch linux
zingolib commit: 37f8fe90cc18fa5b57992350f839f4af9c1697c6 (cargo run --release)
mainnet
No server flag was passed, using whatever the default server is.

zancas:
Attempting to replicate.

zancas:
```BASH
    Finished `release` profile [optimized] target(s) in 4m 20s
     Running `target/release/zingo-cli`
Launching sync task...
Launching save task...
(main) Block:3275838 (type 'help') >> balance
[
    confirmed_orchard_balance: 0
    unconfirmed_orchard_balance: 0
    total_orchard_balance: 0

    confirmed_sapling_balance: 0
    unconfirmed_sapling_balance: 0
    total_sapling_balance: 0

    confirmed_transparent_balance: 0
    unconfirmed_transparent_balance: 0
    total_transparent_balance: 0
]
Sync completed succesfully:
{
    sync start height: 3270182
    sync end height: 3275838
    blocks scanned: 5657
    sapling outputs scanned: 2170
    orchard outputs scanned: 14910
    percentage total outputs scanned: 100
}
(main) Block:3275838 (type 'help') >> info
{
  "version": "v0.4.18",
  "git_commit": "8c4c83aa8d9727c9669fe8e8b80b4ff62d6c95f0",
  "server_uri": "https://zec.rocks:443/",
  "vendor": "ECC LightWalletD",
  "taddr_support": true,
  "chain_name": "main",
  "sapling_activation_height": 419200,
  "consensus_branch_id": "4dec4df0",
  "latest_block_height": 3275838
}
(main) Block:3275838 (type 'help') >> quit
Save task shutdown successfully.
Zingo CLI quit successfully.
1773768758 || shofth nattyb dev[ ]
[...]

PR #2299, opened 2026-03-26
high level pub interface for sync recovery

zancas:
Addresses: #2260 

This PR adds a high-level opinionated interface that makes simple recommendations based on nuanced error internals.

The complete fix will require changes to zingo-cli that I intend to PR separately. 

Note, intend to add integration testing with the zingo-cli PR.


fluidvanadium (approved):
Great

zingolib/src/lightclient/sync.rs
```diff
@@ -152,6 +151,26 @@ impl LightClient {
...
+    /// Polls the sync task and, if it failed, returns the recommended
+    /// recovery action alongside the error description.
+    ///
+    /// This is the primary entry point for consumers (CLI, mobile, PC)
+    /// that need to decide whether to retry, switch servers, or give up
+    /// after a sync failure.
+    ///
+    /// Returns `None` if sync has not been launched, is still running,
+    /// or completed successfully.
+    pub fn poll_sync_recovery(&mut self) -> Option<(SyncRecoveryObservables, String)> {
```
Oscar-Pepper, line 164: i dont understand the use case of this API. consumers are going to be calling `poll_sync` at regular intervals to check if it is started, completed or returned an error. In the case where `poll_sync` returns an error they can no longer call this method as the error has already been returned and there is no longer a sync handle to poll again. In this case they will just call `recovery_recommendation` directly on the error they have already got from `poll_sync`. The only other case is that they use this *instead* of `poll_sync` and get no information for whether the sync has been launched, is still running or has completed which is pointless IMO

PR #2300, opened 2026-03-26
Indexer chain name compatibility

fluidvanadium:
Chaintype is now open to be inferred from Indexer.

### DIFF
--- a/zingolib/src/config.rs
+++ b/zingolib/src/config.rs
@@ -54,8 +54,8 @@ impl TryFrom<&str> for ChainType {
 
     fn try_from(value: &str) -> Result<Self, Self::Error> {
         match value {
-            "mainnet" => Ok(ChainType::Mainnet),
-            "testnet" => Ok(ChainType::Testnet),
+            "mainnet" | "main" => Ok(ChainType::Mainnet),
+            "testnet" | "test" => Ok(ChainType::Testnet),
             "regtest" => Ok(ChainType::Regtest(ActivationHeights::default())),
             _ => Err(InvalidChainType(value.to_string())),
         }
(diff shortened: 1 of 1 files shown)


zingolib/src/config.rs
```diff
@@ -462,7 +462,9 @@ fn wallet_dir_or_default(opt_wallet_dir: Option<PathBuf>, chain: ChainType) -> P
 
 #[cfg(test)]
 mod tests {
-    use crate::config::{ChainType, ClientConfig};
+    use crate::config::ChainType;
+    use crate::config::ClientConfig;
+    use zingo_common_components::protocol::ActivationHeights;
```
zancas, line 467: Is this the type we prefer for this?

fluidvanadium: IDK, this PR didnt change that aspect, it only is imported here to complete a unit test.

zingolib/src/config.rs
```diff
@@ -49,13 +49,18 @@ impl std::fmt::Display for ChainType {
     }
 }
 
+/// Invalid chain type.
+#[derive(thiserror::Error, Debug)]
+#[error("Invalid chain type '{0}'. Expected one of: 'mainnet', 'testnet' or 'regtest'.")]
+pub struct InvalidChainType(String);
```
Oscar-Pepper, line 55: i dont understand why this has moved in between the ChainType impls? please can we keep the impls together

Oscar-Pepper (changes requested):

zingolib/src/config.rs
fluidvanadium: My organization plan was to group `struct Foo` with `impl Foo` and `impl Can for Foo`. This makes it easiest to see how they are used.

> I didnt see that the file was already grouped with `struct` in one group and `impl` in the other. But now that I see there is a pre-existing organization, I can conform to it.

... Is what I would say if the impls *were* together. But they are not. At merge-base, the standard you are setting does not exist.

PR #2303, opened 2026-03-26
fill todo doc comments in zingo-cli

### LINKED ISSUE #2301: zingo-cli has TODO placeholder doc comments
Several public and internal items in `zingo-cli/src/` have `TODO: Add Doc Comment Here!` placeholders instead of real documentation:

**lib.rs:**
- Crate-level module docs (line 1-2)
- `command_loop()` (line 310)
- `ConfigTemplate` struct (line 395)

**commands.rs:**
- `short_help()` trait method (line 40)
- `ShortCircuitedCommand` trait (line 49)
- `exec_without_lc()` method (line 51)
- `HelpCommand` struct (line 498)
- `do_user_command()` function (line 2014)
- Module-level docs (line 1-2, currently reads "upgrade-or-replace")

**version.rs:**
- Module docs (line 1-2)
- `VERSION` const (line 4)

These are all on the `divide_commands` branch (PR #2294).

evanbranigan-dotcom:
Fixes: #2301

Retargeted against `dev` per @zancas feedback on #2301.

Fill all TODO placeholder doc comments across `zingo-cli/src/`:

- **lib.rs**: crate-level docs, `command_loop()`, `ConfigTemplate`
- **commands.rs**: module docs, `short_help()`, `ShortCircuitedCommand`, `HelpCommand`, `get_commands()`, `do_user_command()`
- **version.rs**: module and `VERSION` const docs

Note: `commands.rs` line 1-2 originally read "An interface that passes strings (e.g. from a cli, into zingolib) / upgrade-or-replace" — replaced with clearer module docs. The "upgrade-or-replace" note may have been intentional @zancas; happy to restore if needed.

**AI disclosure**: Doc comments drafted with Claude Code (Anthropic), reviewed and verified against source by human contributor.

### DIFF
--- a/zingo-cli/src/commands.rs
+++ b/zingo-cli/src/commands.rs
@@ -1,5 +1,8 @@
-//! An interface that passes strings (e.g. from a cli, into zingolib)
-//! upgrade-or-replace
+//! Command definitions and dispatch for zingo-cli.
+//!
+//! Each command implements the [`Command`] trait (or [`ShortCircuitedCommand`]
+//! for commands that run without a wallet). All commands are registered in
+//! [`get_commands`] and dispatched by [`do_user_command`].
 
 mod error;
 mod utils;
@@ -37,7 +40,7 @@ pub trait Command {
     /// display command help (in cli)
     fn help(&self) -> &'static str;
 
-    /// TODO: Add Doc Comment for this!
+    /// A one-line summary shown in the two-column command listing.
     fn short_help(&self) -> &'static str;
 
     /// in zingocli, this string is printed to console
@@ -46,9 +49,14 @@ pub trait Command {
     fn exec(&self, _args: &[&str], lightclient: &mut LightClient) -> String;
 }
 
-/// TODO: Add Doc Comment Here!
+/// A command that can execute without an active [`LightClient`].
+///
+/// This is used for commands like `help` that must run before the wallet
+/// is loaded — for example when the user passes `help` as the COMMAND
+/// argument on the command line.
 pub trait ShortCircuitedCommand {
-    /// TODO: Add Doc Comment Here!
+    /// Execute the command without a [`LightClient`], returning the
+    /// output string that will be printed to the console.
     fn exec_without_lc(args: Vec<String>) -> String;
 }
 
@@ -495,7 +503,7 @@ impl Command for ClearCommand {
     }
 }
 
-/// TODO: Add Doc Comment Here!
+/// Lists all available commands or shows detailed help for a specific command.
 pub struct HelpCommand {}
 impl Command for HelpCommand {
     fn help(&self) -> &'static str {
@@ -2011,7 +2019,10 @@ pub fn get_commands() -> HashMap<&'static str, Box<dyn Command>> {
     all
 }
 
-/// TODO: Add Doc Comment Here!
+/// Dispatches a user command by name to the appropriate [`Command`] implementation.
+///
+/// Returns the command's output string, or an "Unknown command" message
+/// if no command with the given name exists.
 pub fn do_user_command(cmd: &str, args: &[&str], lightclient: &mut LightClient) -> String {
     match get_commands().get(cmd.to_ascii_lowercase().as_str()) {
         Some(cmd) => cmd.exec(args, lightclient),
--- a/zingo-cli/src/lib.rs
+++ b/zingo-cli/src/lib.rs
@@ -1,5 +1,12 @@
-//! `ZingoCli`
-//! TODO: Add Crate Description Here!
+//! `ZingoCli` — a command-line interface for the Zingo Zcash light wallet.
+//!
+//! This crate provides the library half of `zingo-cli`. It owns argument
+//! parsing ([`build_clap_app`]), configuration assembly, wallet startup,
+//! the interactive REPL, and single-command dispatch.
+//!
+//! The binary entry point (`main.rs`) is intentionally thin: it handles
+//! process-level concerns (tracing, crypto-provider installation, error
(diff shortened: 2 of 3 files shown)


dorianvp (dismissed):
Nice!

evanbranigan-dotcom:
Rebased on current dev and fixed the doc CI failure — removed the intra-doc link to `ConfigTemplate` (it's `pub(crate)`, so `rustdoc -D warnings` flagged the link as pointing to a private item).

Oscar-Pepper (dismissed):

zingo-cli/src/commands.rs
```diff
@@ -46,9 +49,14 @@ pub trait Command {
     fn exec(&self, _args: &[&str], lightclient: &mut LightClient) -> String;
 }
 
-/// TODO: Add Doc Comment Here!
+/// A command that can execute without an active [`LightClient`].
```
zancas, line 52: There's some implicit confusion in this because both the "active" LightClient and "the wallet" are referenced.

"help" IS the canonical example, but what *exactly* is "not loaded-or-active" should be made completely unambiguous.   Is it the "LightClient" the "wallet" both?  Something else.   A careful and comprehensive review of how these two concepts relate (the lightclient and the wallet) is well worth doing.

Are these concepts coherent?
Are they necessary?
Do they cleanly separate concerns?
Are there other designs that might separate concerns MORE cleanly/coherently?  Why are these concepts used?
What's their history?

Since their original creation, have conditions changed such that they or some parts of them are obsolete?

zingo-cli/src/lib.rs
```diff
@@ -1,5 +1,12 @@
-//! `ZingoCli`
-//! TODO: Add Crate Description Here!
+//! `ZingoCli` — a command-line interface for the Zingo Zcash light wallet.
```
zancas, line 1: Note how this doc-comment implies that the "main.rs" in this context is the thing that is not "wallet logic", does that mean that the architecture of zingo-cli can be studied to reveal insights about the "LightClient" -- vs -- "wallet" design?

zingo-cli/src/lib.rs
```diff
@@ -307,7 +314,11 @@ struct CommandChannel {
     receiver: Receiver<String>,
 }
 
-/// TODO: Add Doc Comment Here!
+/// Spawns a background thread that listens for `(command, args)` messages,
+/// executes each command against the [`LightClient`], and sends the
```
zancas, line 320: Here we reference "LightClient"..

zingo-cli/src/lib.rs
```diff
@@ -392,7 +403,11 @@ fn get_communication_mode(_matches: &clap::ArgMatches) -> CommunicationMode {
     CommunicationMode::Online
 }
 
-/// TODO: Add Doc Comment Here!
+/// All CLI-derived configuration needed to start a wallet session.
```
zancas: Here a "wallet session" is mentioned... 

"wallet session"..  what's that?   Is it defined-and-or used elsewhere?

This apparent coherence with reference to implicitly well-defined concepts, that upon investigation begin to dissolve is where a skilled "AI-Pilot" can really bring value.

To me, these concepts "wallet" "lightclient" "wallet-session" smell suspiciously informal..   like potential loop-holes in a legal framework.   Someone with a Zen Mind, mind discover interesting insights with "What is a BLAH BLAH, again?"  "What does the BLAH BLAH FOO-snargle?" etc.

zancas (dismissed):
This looks like a good start.

evanbranigan-dotcom:
Thanks for the thorough review — these are good questions and I went down the rabbit hole.

**The architecture as it stands:**

`LightClient` and `LightWallet` do separate concerns, but the naming obscures it:

- **`LightWallet`** (`zingolib/src/wallet.rs:112`) — pure in-memory state: keys, addresses, transactions, sync progress, shard trees. No network awareness.
- **`LightClient`** (`zingolib/src/lightclient.rs:89`) — orchestrator/facade that *owns* the wallet (via a private `WalletMeta` wrapper) and manages everything else: gRPC indexer, Tor, sync tasks, persistence. All consumer access to the wallet goes through `LightClient`.
- **`WalletMeta`** (private, `lightclient.rs:49`) — extracts immutable fields (chain_type, birthday, mnemonic, path) outside the `Arc<RwLock<LightWallet>>` for lock-free access.

So the relationship is: `LightClient` owns `WalletMeta` owns `Arc<RwLock<LightWallet>>`. Network sync is handled by `pepper_sync` which talks to `LightWallet` through the `SyncWallet` trait — clean boundary there.

**The leak:** `LightClient::wallet()` (line 191) exposes the raw `Arc<RwLock<LightWallet>>` directly. There's already a TODO acknowledging this should be replaced with pass-through methods. So the facade isn't fully enforced yet.

**"wallet session"** — you're right to flag this. That term isn't used anywhere in the codebase. I was reaching for a concept that doesn't formally exist. What I meant was "the `LightClient` instance and its associated state" — I'll replace it.

**For this PR, I'll revise the doc comments to be precise:**
- `ShortCircuitedCommand`: "without a `LightClient`" (already correct — these literally take no `LightClient` arg)
- `command_loop`: references `LightClient` (correct — it receives one)
- `ConfigTemplate`: replace "wallet session" with something concrete like "create a `LightClient` and start the command loop"
- Crate doc: clarify that `main.rs` handles process-level concerns while `lib.rs` owns the `LightClient` lifecycle — not "wallet logic" in the `LightWallet` sense

The deeper question — whether `LightClient`/`LightWallet` could separate concerns more cleanly — is interesting but probably out of scope for a doc comment PR. Happy to open an issue exploring that if you want.

zancas (approved):

PR #2324, opened 2026-04-10
augmented grace window for zecwalletlite recovery

Oscar-Pepper:
ontop of https://github.com/zingolabs/zingolib/pull/2325


EchoAlice:
I opened a small stacked test PR against this branch here: Oscar-Pepper/zingolib#30

It adds boundary-focused unit coverage for `zip212_enforcement(params, height, extended)` so the extended ZIP-212 grace-window behavior introduced in this PR is pinned at the standard and extended mainnet Canopy boundaries.

PR #2326, opened 2026-04-17
replace zcash protocol BlockHeight with zingo commons type

Oscar-Pepper:
Fixes:


zancas:
Obsolete by ratified decision: mirrored commons types are converted only at the zcash_local_net boundary, so replacing zcash_protocol::BlockHeight in the public API with a commons type runs against the standing discipline.

PR #2329, opened 2026-04-30
test: cover extended ZIP-212 grace window behavior

EchoAlice:
Adds focused unit coverage for the `extended_zip212_grace_period` behavior introduced in zingolabs/zingolib#2324.

The tests pin the ZIP-212 enforcement behavior at the standard and extended mainnet Canopy grace-window boundaries, including the height where standard and extended modes first diverge.

Tested with:

```bash
cargo test -p pepper-sync zip212_enforcement
```

### DIFF
--- a/pepper-sync/CHANGELOG.md
+++ b/pepper-sync/CHANGELOG.md
@@ -20,6 +20,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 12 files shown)


EchoAlice:
Closing this and reopening as a stacked PR against Oscar-Pepper:augmented_grace_window_for_zecwalletlite_recovery so the diff only shows the test commit on top of #2324.

PR #2331, opened 2026-05-02
Update workspace to corez dependency graph

pacu:
## Summary
- update librustzcash dependencies to the released corez-based line
- point zingo-netutils and infrastructure test dependencies at the companion corez migration branches
- use workspace zingo-memo and zingo-status so they inherit the updated dependency graph
- apply initial pepper-sync import and lightwallet protocol updates for the new crate layout

## Companion PRs
- zingolabs/zingo-common#69
- zingolabs/infrastructure#261

## Verification
- cargo tree -i core2 --locked --offline exits with no matching package

## Current status
This is a draft because the dependency graph is core2-free, but the full workspace check still needs the zcash_client_backend 0.22 wallet API migration. Current remaining failures are in zingolib wallet/data-api implementations: missing new WalletRead/WalletWrite trait methods, TransparentKeyOrigin/TransparentOutputFilter updates, TxVersion parameters, and ZIP-321 amount API changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,11 +18,9 @@ resolver = "2"
 ### Workspace
 # pepper-sync = "0.2.0"
 pepper-sync = { path = "pepper-sync" } # NOTE: for development between releases
-# zingo-memo = { path = "zingo-memo" } # NOTE: for development between releases
-zingo-memo = "0.1.0"
+zingo-memo = { path = "zingo-memo" } # NOTE: for development between releases
 zingo-price = { path = "zingo-price" }
-# zingo-status = { path = "zingo-status" } # NOTE: for development between releases
-zingo-status = "0.2.0"
+zingo-status = { path = "zingo-status" } # NOTE: for development between releases
 zingolib = { path = "zingolib" }
 
 ### Tests and Integration Tests
@@ -30,8 +28,8 @@ portpicker = "0.1"
 proptest = "1.6.0"
 tempfile = "3"
 test-log = "0.2"
-zcash_local_net = { git = "https://github.com/zingolabs/infrastructure.git", rev = "5bcbf53d680542c247d01585c12085c1be831adc" }
-zingo_test_vectors = { git = "https://github.com/zingolabs/infrastructure.git", rev = "5bcbf53d680542c247d01585c12085c1be831adc" }
+zcash_local_net = { git = "https://github.com/zingolabs/infrastructure.git", branch = "fix-corez-deps" }
+zingo_test_vectors = { git = "https://github.com/zingolabs/infrastructure.git", branch = "fix-corez-deps" }
 
 ### Documentation
 indoc = "2"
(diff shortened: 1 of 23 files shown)


pacu:
Added a scoped `#[allow(deprecated)]` around the existing `get_block_range_nullifiers` call so this corez / `zcash_client_backend 0.22` dependency migration does not fail CI solely on that deprecated RPC warning.

Tracked the actual replacement work separately in #2332.

Cargo.toml
```diff
@@ -18,20 +18,18 @@ resolver = "2"
...
 zingolib = { path = "zingolib" }
 
 ### Tests and Integration Tests
 portpicker = "0.1"
 proptest = "1.6.0"
 tempfile = "3"
 test-log = "0.2"
-zcash_local_net = { git = "https://github.com/zingolabs/infrastructure.git", rev = "5bcbf53d680542c247d01585c12085c1be831adc" }
-zingo_test_vectors = { git = "https://github.com/zingolabs/infrastructure.git", rev = "5bcbf53d680542c247d01585c12085c1be831adc" }
+zcash_local_net = { git = "https://github.com/zingolabs/infrastructure.git", branch = "fix-corez-deps" }
```
zancas: Do we still need this branch infrastructure?
Infrastructure:

cargo tree -i core2

Nothing exists.

cargo tree -i corez

Single version.

Cargo.toml
```diff
@@ -18,20 +18,18 @@ resolver = "2"
...
+zingo_test_vectors = { git = "https://github.com/zingolabs/infrastructure.git", branch = "fix-corez-deps" }
```
zancas: See above.

Cargo.toml
```diff
@@ -43,32 +41,32 @@ tracing = "0.1"
 tracing-subscriber = "0.3"
 
 ### Zingo-common
-zingo-netutils = { git = "https://github.com/zingolabs/zingo-common", branch = "indexer_trait", features = ["back_compatible", "globally-public-transparent"] }
+zingo-netutils = { git = "https://github.com/zingolabs/zingo-common", branch = "fix-corez-deps", features = ["back_compatible", "globally-public-transparent"] }
```
zancas: I haven't checked zingo-common, but I also expect this is no longer necessary there.

zancas (dismissed):
Non blocking comments, that I will likely fix soon.

idky137:
The following tests are currently failing on this branch for me:

```
        FAIL [  96.794s] libtonode-tests::concrete slow::sapling_incoming_sapling_outgoing
        FAIL [ 116.117s] libtonode-tests::concrete slow::sapling_to_sapling_scan_together
        FAIL [ 110.790s] libtonode-tests::concrete slow::send_to_transparent_and_sapling_maintain_balance
```

-----
Local bins:
- zcashd / zcash-cli: `v6.12.1`
- zebrad: `4.4.0`
- Lightwalletd: `v0.4.19`
- Zainod: `Dev`

pacu:
Tracked down `slow::send_to_transparent_and_sapling_maintain_balance`.

Root cause: with `zcash_client_backend 0.22`, `OvkPolicy::Sender` selects a unified sender OVK based on the input pools. An Orchard-funded send to a Sapling recipient can therefore encrypt the Sapling output with the sender's Orchard OVK bytes. pepper-sync was only trying Sapling OVKs for Sapling outputs, so it missed the external Sapling outgoing note and classified the transaction as `send-to-self`.

Fix pushed in `a18690f8`: pepper-sync now tries each account's unified OVK bytes across Sapling and Orchard outgoing recovery. I also updated the affected test expectations: internal Orchard change is still present as an incoming wallet note, but under the 0.22 sender policy it is not expected to appear as outgoing-note metadata.

Verified:
- `cargo check -p pepper-sync --locked --offline`
- `cargo fmt --check`
- `cargo nextest run --test-threads 1 -p libtonode-tests slow::send_to_transparent_and_sapling_maintain_balance` passes

pacu:
Focused on `slow::sapling_to_sapling_scan_together` next.

Root cause: after the 0.22 OVK behavior change, internal shielded change is no longer represented as outgoing-note metadata. `WalletTransaction::total_value_sent()` was still subtracting every wallet-owned Sapling/Orchard note from outgoing note value. For a Sapling-to-Sapling send, the transaction had an external outgoing Sapling note for 20_000 and a wallet-owned Sapling change note for 70_000, so the old saturating subtraction produced sent value 0.

Fix pushed in `059079dd`: shielded sent value now sums outgoing notes whose output IDs are not also wallet-owned outputs. This keeps external outgoing notes counted and excludes change/self outputs by identity instead of by pool totals.

Verified:
- `cargo check -p pepper-sync --locked --offline`
- `cargo fmt --check`
- `cargo nextest run --test-threads 1 -p libtonode-tests slow::sapling_to_sapling_scan_together` passes
- `cargo nextest run --test-threads 1 -p libtonode-tests slow::send_to_transparent_and_sapling_maintain_balance` still passes

idky137:
AFAICT the PR which is causing these new test failures is: https://github.com/zcash/librustzcash/pull/2058 ( pulled in zcash_client_backend v0.22.0 )

CHANGELOG entry:
```
### Changed
- `zcash_client_backend::wallet::OvkPolicy` has been substantially modified
  to reflect the view that a single outgoing viewing key should be uniformly
  applied to encrypt all external transaction outputs, irrespective of which
  pool those outputs are produced into, and that by default wallet-internal
  outputs should not be decryptable using an OVK (and will therefore only
  be recoverable using the wallet's internal IVK).
  - The semantics of `OvkPolicy::Sender` have changed. In addition to using
    a single OVK for all shielded outputs irrespective of pools, it now
    specifies that wallet-internal change outputs should be treated as though
    the policy for those outputs were `OvkPolicy::None`, rendering them only
    recoverable using the wallet's internal IVK. 
  - The `OvkPolicy::Custom` variant has changed. Instead of pool-specific
    OVKs, this now encapsulates a pair of OVKs, one to be used for all
    shielded external outputs of the transaction, and a second (optional)
    key that will be used to encrypt any wallet-internal change outputs
    that would otherwise only be recoverable using the wallet's internal
    IVK.
```

idky137:
A fix for these broken tests has been implemented here: https://github.com/zingolabs/zingolib/pull/2333

pacu:
Fixed the "Run doc tests" CI failure in `ca778672`.

Cause: two `zingo-status` doctests still imported `BlockHeight` from `zcash_primitives::consensus`; with the updated librustzcash crates it lives at `zcash_protocol::consensus`.

Verified after rebasing on the latest branch head:
- `cargo test -p zingo-status --doc --locked --offline`
- `cargo test --doc --locked --offline`

zancas (approved):

Issue #2347, opened 2026-05-13
incorrect scan range truncation


Oscar-Pepper:
scan range truncation should only happen in the case where the chain_height is lower than the last_known_chain_height. currently scan ranges will also be truncated during *any* re-org scenario which is not intended behaviour.

PR #2348, opened 2026-05-13
fix scan range truncation bug

Oscar-Pepper:
fixes issue where scan range is truncated during re-org when chain height is higher than wallet's last known chain height.

this PR also fixes a re-org test "reorg_changes_incoming_tx_height" and feature gates out the part of sync which calls a broken RPC in darkside to verify this bug has been fixed. this test is now no longer ignored and passes where as when this test is run on current stable branch without the broken RPC call it returns this failure:

        FAIL [   7.630s] darkside-tests::advanced_reorg_tests reorg_changes_incoming_tx_height
  stdout ───

    running 1 test
    test reorg_changes_incoming_tx_height ... FAILED

    failures:

    failures:
        reorg_changes_incoming_tx_height

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 7 filtered out; finished in 7.62s

  stderr ───
    [darkside-tests/tests/advanced_reorg_tests.rs:114:5] &uri = http://127.0.0.1:22003/
    [darkside-tests/tests/advanced_reorg_tests.rs:158:5] &uri = http://127.0.0.1:22003/

    thread 'reorg_changes_incoming_tx_height' panicked at /home/oscar/tmp/zingolib/zingolib/pepper-sync/src/sync/state.rs:829:10:
    scan ranges must be non-empty

    thread 'reorg_changes_incoming_tx_height' panicked at zingolib/src/lightclient/sync.rs:114:47:
    task panicked: JoinError::Panic(Id(135), "scan ranges must be non-empty", ...)

────────────
     Summary [   7.631s] 1 test run: 0 passed, 1 failed, 183 skipped
        FAIL [   7.630s] darkside-tests::advanced_reorg_tests reorg_changes_incoming_tx_height

Fixes: #2347

### DIFF
--- a/.github/workflows/cargo-checkmate.yaml
+++ b/.github/workflows/cargo-checkmate.yaml
@@ -12,7 +12,7 @@ jobs:
         phase: [build, check, clippy, doc, format]
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Setup Rust toolchain
         uses: actions-rust-lang/setup-rust-toolchain@v1
--- a/.github/workflows/ci-nightly.yaml
+++ b/.github/workflows/ci-nightly.yaml
@@ -18,7 +18,7 @@ jobs:
     if: ${{ github.event_name != 'pull_request' || !github.event.pull_request.draft }}
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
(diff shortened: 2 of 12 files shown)


fluidvanadium (approved):
If I understand correctly, the sole functional change is that truncate_scan_ranges is no longer called after "Re-org detected.".

Brings the test back to functionality with a precise modification. 👍

PR #2349, opened 2026-05-13
fix: use exponentiation for saplling diversifier idx

### LINKED ISSUE #2344: Sapling diversifier index limit uses XOR (^) instead of exponentiation
### Summary
In `LightWallet::add_sapling_address `(`zingolib/src/wallet/sync.rs`), the diversifier index bound check uses `2 ^ 16`, which in Rust is bitwise XOR (`2 XOR 16 = 18`), not 2 to the power of 16 (`2^16= 65536`) as the comment states and would be the correct functionality.

### Location
https://github.com/zingolabs/zingolib/blob/dev/zingolib/src/wallet/sync.rs#L91

### Impact
Diversifier indices bigger or equal to 19 are silently dropped from unified address discovery. 

### Fix
Replace the literal with an explicit form, e.g.:
```
const MAX_SAPLING_DIVERSIFIER_INDEX: u128 = 1 << 16;

if u128::from(diversifier_index) > MAX_SAPLING_DIVERSIFIER_INDEX {
    return Ok(());
}
```
Alternatively, use something like `2u128.pow(16)`.

Also: While fixing, we recommend confirming the intended boundary. The current `>` accepts `65536` itself. Since the limit is meant to be "16 bits of index space" (`0..=65535`), switch to `>=`.

dorianvp:
Fixes: #2344.

### DIFF
--- a/zingolib/src/wallet/sync.rs
+++ b/zingolib/src/wallet/sync.rs
@@ -88,7 +88,8 @@ impl SyncWallet for LightWallet {
         address: sapling_crypto::PaymentAddress,
         diversifier_index: DiversifierIndex,
     ) -> Result<(), Self::Error> {
-        if u128::from(diversifier_index) > 2 ^ 16 {
+        const MAX_SAPLING_DIVERSIFIER_INDEX: u128 = 1 << 16;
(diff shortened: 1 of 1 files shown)


dorianvp:
Closing in favor of #2351.

PR #2351, opened 2026-05-14
fix: use exponentiation for sapling diversifier idx

dorianvp:
Fixes: #2344.

### DIFF
--- a/zingolib/src/wallet/sync.rs
+++ b/zingolib/src/wallet/sync.rs
@@ -259,7 +259,7 @@ mod tests {
     fn add_sapling_address_below_limit_proceeds_past_guard() {
         let mut wallet = test_wallet();
         let (_, _, addr) = default_zaddr();
-        let missing_account = zip32::AccountId::try_from(0).unwrap();
+        let account_id = zip32::AccountId::try_from(0).unwrap();
 
         let result = SyncWallet::add_sapling_address(
             &mut wallet,
--- a/zingolib/src/wallet/sync.rs
+++ b/zingolib/src/wallet/sync.rs
@@ -235,9 +235,7 @@ mod tests {
         .unwrap()
     }
 
-    /// Index 65 536 (`2u128.pow(16)`, the exclusive upper bound of the 16-bit index space) must
-    /// early-return `Ok(())` without inserting an address, confirming the limit uses `.pow()`
-    /// and not the XOR expression `2 ^ 16 = 18` that caused the original bug.
+    /// Verifies that diversifier indices equal to or above the limit of 65536 return early and does not insert a new address into the wallet.
     #[test]
     fn add_sapling_address_at_limit_early_returns() {
         let mut wallet = test_wallet();
--- a/zingolib/src/wallet/sync.rs
+++ b/zingolib/src/wallet/sync.rs
@@ -214,6 +214,7 @@ mod tests {
         wallet::{LightWallet, WalletBase, WalletSettings},
     };
     use pepper_sync::config::{PerformanceLevel, SyncConfig, TransparentAddressDiscovery};
+    use sapling_crypto::zip32::DiversifiableFullViewingKey;
     use zip32::DiversifierIndex;
 
     fn test_wallet() -> LightWallet {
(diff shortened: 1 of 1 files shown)


zingolib/src/wallet/sync.rs
```diff
@@ -201,3 +202,74 @@ impl SyncShardTrees for LightWallet {
...
+                },
+                min_confirmations: NonZeroU32::try_from(1).unwrap(),
+            },
+        )
+        .unwrap()
+    }
+
+    /// Index 65 536 (`2u128.pow(16)`, the exclusive upper bound of the 16-bit index space) must
+    /// early-return `Ok(())` without inserting an address, confirming the limit uses `.pow()`
+    /// and not the XOR expression `2 ^ 16 = 18` that caused the original bug.
```
Oscar-Pepper: please remove comments about the bugfix. we can just describe the test here

zingolib/src/wallet/sync.rs
```diff
@@ -201,3 +202,74 @@ impl SyncShardTrees for LightWallet {
...
+        assert!(result.is_ok());
+        assert_eq!(
+            wallet.unified_addresses().len(),
+            before,
+            "early return must not insert an address"
+        );
+    }
+
+    /// Indices the old buggy guard (`2 ^ 16 = 18`, i.e. `> 18`) silently dropped must now
+    /// pass through and attempt real work, witnessed here by `Err(NoAccountKeys)`.
```
Oscar-Pepper: as above, i think its unnecessary to describe the XOR and the test doesn't assert an Err. I think the comment can read:

```suggestion
    /// Verifies that diversifier indices below the limit of 65536 correctly insert a new address into the wallet.
```

zingolib/src/wallet/sync.rs
Oscar-Pepper:
```suggestion
    /// Verifies that diversifier indices equal to or above the limit of 65536 return early and does not insert a new address into the wallet.
```

zingolib/src/wallet/sync.rs
```diff
@@ -201,3 +202,74 @@ impl SyncShardTrees for LightWallet {
...
+    }
+
+    /// Index 65 536 (`2u128.pow(16)`, the exclusive upper bound of the 16-bit index space) must
+    /// early-return `Ok(())` without inserting an address, confirming the limit uses `.pow()`
+    /// and not the XOR expression `2 ^ 16 = 18` that caused the original bug.
+    #[test]
+    fn add_sapling_address_at_limit_early_returns() {
+        let mut wallet = test_wallet();
+        let (_, _, addr) = default_zaddr();
+        let missing_account = zip32::AccountId::try_from(0).unwrap();
```
Oscar-Pepper:
```suggestion
        let account_id = zip32::AccountId::try_from(0).unwrap();
```

zingolib/src/wallet/sync.rs
```diff
@@ -201,3 +202,74 @@ impl SyncShardTrees for LightWallet {
...
+        );
+    }
+
+    /// Indices the old buggy guard (`2 ^ 16 = 18`, i.e. `> 18`) silently dropped must now
+    /// pass through and attempt real work, witnessed here by `Err(NoAccountKeys)`.
+    #[test]
+    fn add_sapling_address_below_limit_proceeds_past_guard() {
+        let mut wallet = test_wallet();
+        let (_, _, addr) = default_zaddr();
+        let missing_account = zip32::AccountId::try_from(0).unwrap();
```
Oscar-Pepper:
```suggestion
        let account_id = zip32::AccountId::try_from(0).unwrap();
```

zingolib/src/wallet/sync.rs
```diff
@@ -201,3 +202,74 @@ impl SyncShardTrees for LightWallet {
...
+
+        let result = SyncWallet::add_sapling_address(
+            &mut wallet,
+            missing_account,
+            addr.clone(),
+            DiversifierIndex::from(1u32),
+        );
+        assert!(result.is_ok(),);
```
Oscar-Pepper: we should assert that the wallet's unified addresses length has incremented similar to how the above test asserts the length is unchanged

Oscar-Pepper (changes requested):
great fix, have some requested changes to the tests

Oscar-Pepper (approved):

Issue #2352, opened 2026-05-15
re-org may result in wallet being unable to construct a witness for spendable notes


Oscar-Pepper:
retention mark removals of shardtree leaves are incorrectly stored in checkpoints corresponding to blockheights the spent note was received instead of the blockheight at which it was spent. this means that the wallet may clear the information needed to build a witness for a mark due to the received note being below the re-org window but the tx the note is spent in may be re-orged, potentially resulting in the note becoming spendable after the witnesss data has been discarded.

PR #2354, opened 2026-05-15
fix premature discarding of witness data

Oscar-Pepper:
fixes mark removal bug and generalizes update_spent_notes fn for shielded protocols

Fixes: https://github.com/zingolabs/zingolib/issues/2352


fluidvanadium (approved):
*remove_mark* now correctly takes *spending_height* instead of *height*

PR #2356, opened 2026-05-22
Fix append subtree roots

Oscar-Pepper:
Fixes: #2357 

added test "add_subtree_roots". failed on without the subtree root index bugfix in the following way:

```
$ cargo nextest run -p libtonode-tests add_subtree_roots --no-capture
   Compiling pepper-sync v0.3.0 (/home/oscar/src/zingolib/pepper-sync)
   Compiling zingolib v3.0.1 (/home/oscar/src/zingolib/zingolib)
   Compiling zingolib_testutils v0.1.0 (/home/oscar/src/zingolib/zingolib_testutils)
   Compiling libtonode-tests v0.2.0 (/home/oscar/src/zingolib/libtonode-tests)
    Finished `test` profile [optimized + debuginfo] target(s) in 47.89s
────────────
 Nextest run ID 134ed0c1-82fb-417c-a91a-56cb6e84f7d6 with nextest profile: default
    Starting 1 test across 6 binaries (83 tests skipped)
       START (1/1) libtonode-tests::sync add_subtree_roots

running 1 test

thread 'add_subtree_roots' panicked at libtonode-tests/tests/sync.rs:107:9:
no of sapling shard roots wallet: 1096, no of sapling shard roots server: 1127
test add_subtree_roots ... FAILED

failures:

failures:
    add_subtree_roots

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 4 filtered out; finished in 8.62s

        FAIL [   8.624s] (1/1) libtonode-tests::sync add_subtree_roots
```

test passes on this branch.

- also fixes an issue where sync polling states sync is running but sync mode states sync is not running.
- correctly builds libtonode-tests without "darkside-tests" feature which disables necessary RPC calls to test pepper-sync

#2348 will need to merge first to fix ci test failure

### DIFF
--- a/zingolib/src/lightclient/sync.rs
+++ b/zingolib/src/lightclient/sync.rs
@@ -21,7 +21,6 @@ impl LightClient {
     /// `sync_handle` field.
     // TODO: add realtime sync updates to zingo-cli when it can handle printing during user input
     pub async fn sync(&mut self) -> Result<(), LightClientError> {
-        dbg!("sync 1");
         if self.sync_mode() != SyncMode::NotRunning {
             return Err(LightClientError::SyncModeError(
                 SyncModeError::SyncAlreadyRunning,
@@ -36,7 +35,6 @@ impl LightClient {
         let wallet = self.wallet.clone();
         let sync_mode = self.sync_mode.clone();
         let sync_handle = tokio::spawn(async move {
-            dbg!("sync 2");
             pepper_sync::sync(client, &network, wallet, sync_mode, sync_config).await
         });
 
@@ -47,7 +45,6 @@ impl LightClient {
             interval.tick().await;
         }
 
-        dbg!("sync 3");
         self.sync_handle = Some(sync_handle);
 
         Ok(())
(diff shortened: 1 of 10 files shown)


zingolib/src/lightclient/sync.rs
```diff
@@ -37,6 +37,14 @@ impl LightClient {
         let sync_handle = tokio::spawn(async move {
             pepper_sync::sync(client, &network, wallet, sync_mode, sync_config).await
         });
+
+        let mut interval = tokio::time::interval(Duration::from_millis(50));
+        interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+        interval.tick().await;
+        while self.sync_mode() == SyncMode::NotRunning {
```
dorianvp: This waits on the `sync_mode` atomic with no timeout and no check on the task handle. If the task
returns before setting `Running` (like from `from_atomic_u8(...)?` at sync.rs:327, which returns
before `Running` is set at 329-330) doesn't this loop hang forever?

Oscar-Pepper: good point, we can't poll sync here as this is a fix to the polling issue and it also could return a SyncResult which would make the API uneccessarily complicated. a timeout is a good idea here.

Oscar-Pepper: i have added a timeout which then polls sync, we should never hit a successful sync result in this case but if it was possible sync would just complete silently which is fine.

dorianvp (dismissed):

Issue #2361, opened 2026-05-26
pending spends with no change do not get updated to confirmed


Oscar-Pepper:
when processing transactions that evaded trial decryption due to not having any decryptable outputs, any transactions already in the wallet are skipped as they were clearly *not* evaded. however, there is an issue where a spending transaction is created with a pending status locally in the wallet which may have no change output. therefore, when the transaction is confirmed on chain it is skipped due to already being in the wallet and the status is never correctly updated to confirmed.

PR #2363, opened 2026-05-26
pending spends with no change update to confirmed

Oscar-Pepper:
Fixes: https://github.com/zingolabs/zingolib/issues/2361


dorianvp (approved):
utACK

Issue #2350, opened 2026-05-14
re-org bug, wallet data truncation error


Oscar-Pepper:
on oscar-pepper/re-org_bug branch run:

cargo nextest run --run-ignored all reorg_expires_incoming_tx --no-capture

shows the wallet tx has not been removed during re-org

Oscar-Pepper:
on investigation this turned out to be a darkside bug, not an issue with re-org

zancas:
Resolved by the current truncation design: `truncate_wallet_transactions` sets every transaction confirmed above the truncation height to `Failed` and clears the spending fields of outputs whose spender failed (pepper-sync traits.rs), and the PR #2495 arc hardened the tree side of the same flow. Note the ratified behavior is marking, not removal — `Failed` is a permanent terminal record and consumers filter it at the display layer — so a reorged-out transaction correctly stops counting toward balance while remaining in history. If removal rather than marking is the desired semantics, that is a design disagreement worth its own issue against the current code.

PR #2364, opened 2026-05-26
Fix darkside test reorg expires incoming tx

Oscar-Pepper:
Fixes: #2350

what was thought to be a possible re-org bug in the related issue turns out to be due to darkside issues where the internal zcash tx has an incorrect txid. this PR fixes this by setting failed txs by the wallet tx's txid instead.


zancas:
Obsolete: the standalone darkside-tests crate was retired with ADR 0005 (its authoritative successor is the mock-indexer darkside module inside zingolib), so the test this branch fixes no longer has a home.

PR #2371, opened 2026-06-01
feat: offline mode

dorianvp:
This PR introduces the Indexerless client state: `ClientConfig` no longer requires an Indexer URI, a client built without one starts offline, and `set_indexer_uri()` is the transition online. Local operations work without a connection; network operations fail with the typed `LightClientError::Offline`. ADR 0001 (`docs/adr/0001-offline-by-default.md`) records the offline-by-default decision; @dorianvp authored the core implementation.

The branch was rebased onto `pre_ironwood` (#2448) and the PR retargeted accordingly, so it composes with the completed zebrad test upgrade. Notable outcomes of the rebase: the synthetic-wallet test constructor `new_for_test` shed its never-contacted placeholder endpoint and is now genuinely Indexerless; `pre_ironwood`'s typed `do_info` (the first #2446 migration) merged cleanly with the optional-indexer plumbing; and the two glossaries that collided at the root are split into per-context files under a root `CONTEXT-MAP.md`, with the library glossary corrected for the post-#2448 world.

The vocabulary and architecture are ratified in the glossary (`zingolib/CONTEXT.md`) and ADR 0006: **Indexerless** names the library state, which may be transitional but may equally last a whole session; **Offline mode** names the zingo-cli session concept (#2286). Proposing is an Indexerless capability — the send path gates on the Indexer at the calculate/transmit stage, not before — and the stored-proposal pattern is scheduled for retirement in follow-up work atop this PR, together with the typed-error expansion (#2446) and the zingo-mobile bridge changes ADR 0006 describes.


zingolib/examples/wallet_from_file.rs
```diff
@@ -0,0 +1,161 @@
...
+//! ```
+
+use std::num::NonZeroU32;
+use std::path::PathBuf;
+
+use zingolib::config::{ClientConfig, WalletConfig};
+use zingolib::lightclient::LightClient;
+use zingolib::wallet::{PerformanceLevel, SyncConfig, TransparentAddressDiscovery, WalletSettings};
+
+/// A 24-word BIP-39 mnemonic used only for this demonstration.
```
zancas, line 47: This mnemonic is used elsewhere.. the comment seems to imply that it's used here EXCLUSIVELY.

zingolib/src/wallet/zcb_traits.rs
```diff
@@ -83,31 +88,103 @@ impl WalletRead for LightWallet {
...
+        Ok(stored_ufvk.encode(&self.chain_type) == derived_ufvk.encode(&self.chain_type))
     }
 
     fn seed_relevance_to_derived_accounts(
         &self,
-        _seed: &secrecy::SecretVec<u8>,
+        seed: &SecretVec<u8>,
     ) -> Result<zcash_client_backend::data_api::SeedRelevance<Self::AccountId>, Self::Error> {
-        unimplemented!()
+        use zcash_client_backend::data_api::SeedRelevance;
```
zancas: Interesting...

zancas:
1. Make the Account trait methods not panic on reachable paths.

My concern is that we may have turned a compile-time gap into a runtime panic: if any zcash_client_backend path we now reach calls one of these accessors, it aborts. 

We MUST guarantee at compile time that the accessor calls can't cause a panic (for example... even if there's a malicious ZCB).

Can we write an adversarial test that causes a panic?

  2. Avoid the shared temp-file race between flush() and save_task.
 
  Both call write_to_path, which always writes to the same wallet_path.with_extension("…tmp"). The wallet().write() lock serializes serialization but not the subsequent file write, so a consumer calling flush() while the background save_task is running can have two writers truncate/interleave the same .tmp before the rename.

Please add an adversarial test that causes interleaving and is fixed by uniquifying the temp filename.

  3. Don't flatten the offline error to a debug string in do_info.
 
  The offline branch returns format!("{e:?}"), which collapses to "Offline" and loses the helpful Display message ("Offline: no indexer configured. Call set_indexer_uri() to connect.").

Make do_info return a structured type, so we can leverage Display properly.

  4. Add an offline → online transition test.
 
  The current tests cover the offline error paths thoroughly but not the transition that's the whole point of the feature:

  construct offline, call set_indexer_uri(), then confirm network ops no longer return Offline.

Even a no-network assertion that indexer_uri().is_some() after set_indexer_uri would lock in the happy path.

zancas (changes requested):
See my comment.

zancas (changes requested):
Add tests proving that neither pepper-sync, nor any other network-requiring functionality, CAN be exercised in offline-mode.

These guarantees MUST be fully auditable at compile-time.

I.E. All paths that exercise network-dependent logic must be conditionally INaccessible in offline-mode, and that condition MUST be enforced at a Single Point.

zancas:
Consider an offline-only feature-gate that compiles only offline functionality.

dorianvp:
Another TODO: feature-gate online capabilities AND make it the default!

zancas:
## Status: the Indexerless foundation is complete; Offline mode is not yet

A full `makers hierarchy-test` run passes on this branch (295 tests; the three libtonode failures from the earlier run were issue #2450, cured by merging #2451's fix into this branch — git will deduplicate when it lands on `pre_ironwood`).

**Complete here:**
- The Indexerless client state: URI-less `ClientConfig`, `Option`al indexer, the typed `Offline` error, and `set_indexer_uri()` as the transition. Creating, restoring, loading, reading stored state, and proposing all work with no Indexer; the send path gates at the calculate/transmit stage, not before proposing.
- The stored-proposal retirement (ADR 0006), in three phases: the public value-taking `send_proposal(&ZingoProposal, resume_sync)` whose borrowed proposal survives an Indexerless attempt by construction; the zingo-cli session holding the pending proposal, threaded functionally through a defaulted `Command` trait method (the ~46 non-flow commands untouched); and the deletion of the wallet slot, `send_stored_proposal`, and their machinery. This is the `feat!` commit: zingo-mobile's `confirm()` calls the deleted API at its current pin, so its pin bump must land the bridge-held slot ADR 0006 describes.
- The glossary split (`CONTEXT-MAP.md`, per-context glossaries) with the ratified vocabulary: **Indexerless** (library state, possibly lasting a whole session) and **Offline mode** (the zingo-cli session concept, #2286).

**Deliberately not in this PR:**
- The `--offline` CLI flag (#2286): the library can be Indexerless, but zingo-cli does not yet pin a session to it.
- Offline signing: calculation still requires an Indexer; the ratified propose-and-sign boundary needs the calculate/broadcast endpoint split, with an explicit expiry policy for transactions signed against a stale tip.
- The typed-error expansion (#2446): zingolib's own surface is now nearly clean (`do_info` is typed on this branch); the remaining work is the FFI layer's, tracked for the mobile pin bump.

zingolib/src/lightclient/send.rs
```diff
@@ -16,71 +13,92 @@ use crate::data::proposal::ZingoProposal;
...
+        Ok(txids)
     }
 
-    async fn shield(
+    /// Transmits previously calculated transactions to the Indexer, in the
+    /// given order — the broadcast half of the offline-signing flow.
+    /// Requires an Indexer; an Indexerless attempt fails with
+    /// [`LightClientError::Offline`] and the Calculated transactions remain
+    /// in the wallet, ready to broadcast once connected.
+    pub async fn broadcast_calculated(
```
Oscar-Pepper: id prefer `transmit` as `broadcast` has other meanings in the new memo format. i.e. broadcasting to groups of people

zingolib/src/wallet/disk.rs
```diff
@@ -51,7 +51,7 @@ impl LightWallet {
     }
 
     /// Serialize into `writer`
-    pub fn write<W: Write>(
+    pub(crate) fn write<W: Write>(
```
Oscar-Pepper, line 54: breaking api change that removes a consumers ability to read/write the wallet

zingolib/src/wallet/disk.rs
```diff
@@ -128,7 +128,7 @@ impl LightWallet {
 
     /// Deserialize into `reader`
     // TODO: update to return WalletError
-    pub fn read<R: Read>(mut reader: R, chain_type: ChainType) -> io::Result<Self> {
+    pub(crate) fn read<R: Read>(mut reader: R, chain_type: ChainType) -> io::Result<Self> {
```
Oscar-Pepper, line 131: breaking api change that removes a consumers ability to read/write the wallet

zingolib/src/wallet/zcb_traits.rs
Oscar-Pepper, line 1: all of the changes in this file should be reverted as the api we use of zcb never calls these implementations so we end up with code we need to maintain and test that is never called

zingolib/src/wallet.rs
```diff
@@ -366,7 +360,7 @@ impl LightWallet {
     ///
     /// Intended to be called from a save task which calls `save` in a loop, awaiting the wallet lock and checking
     /// `self.save_required` status, writing the returned wallet bytes to persistance.
-    pub fn save(&mut self) -> std::io::Result<Option<Vec<u8>>> {
+    pub(crate) fn save(&mut self) -> std::io::Result<Option<Vec<u8>>> {
```
Oscar-Pepper: as mentioned above, this sohuld be made private when the whole walllet struct is made private. otherwise the consumer can still use the wallet without the ability to save

zingolib/src/wallet.rs
```diff
@@ -141,10 +138,8 @@ pub struct LightWallet {
     pub wallet_settings: WalletSettings,
     /// The current and historical daily price of zec.
     pub price_list: PriceList,
-    /// Send proposal
-    send_proposal: Option<ZingoProposal>,
```
Oscar-Pepper, line 145: why are we removing the proposal from the wallet? this breaks the ability to create a proposal and have the user confirm it. the consumer would now have to handle this completely and im not sure what the benefit is. this increases complexity for the consumers and breaks zingo-cli send currently

Oscar-Pepper (changes requested):

zingolib/src/lightclient/send.rs
zancas: Addresses in followup.

zingolib/src/wallet/disk.rs
zancas: reverted

zingolib/src/wallet/disk.rs
zancas: reverted

zingolib/src/wallet/zcb_traits.rs
zancas: reverted

zingolib/src/wallet.rs
zancas: reverted

zingolib/src/wallet.rs
zancas: reverted

zancas:
Superceded by #2457

PR #2372, opened 2026-06-02
remove refund addresses in the case of send failure

Oscar-Pepper:
Fixes: https://github.com/zingolabs/zingolib/issues/2387

This PR adds logic to remove any newly created refund addresses in the event of the first de-shielding tex transaction failing.


zancas:
@fluidvanadium can you review this?

PR #2416, opened 2026-06-17
lightclient::from_bytes added for mobile consumers

juanky201271:
Adds `LightClient::from_bytes(bytes, config)` so consumers can construct a `LightClient`
from a pre-serialized wallet without any filesystem read. Intended for iOS and Android,
where the native layer owns wallet persistence and passes raw bytes across the FFI; the
existing `WalletConfig::Read` variant forces a `File::open(wallet_path)`, which currently
forces our zingo-mobile wrapper into a `std::env::temp_dir()` staging step that fails with
`Permission denied (os error 13)` on Android devices where `TMPDIR` is unset and the app
UID cannot write to `/tmp`. 

The function reuses the existing `LightWallet::read` deserializer (already generic over
`Read`), wrapping the bytes in `std::io::Cursor`. The `config` is still required to build
the `GrpcIndexer` and to propagate chain type + wallet_path/wallet_name (the latter only
used if the caller later runs a save task — `from_bytes` itself never touches the disk).

This essentially re-implements PR #2342 (originally by @dorianvp on top of
`nym_wallet_poc_2_2`), rebased onto current `dev` with the v5 LightClient struct (no
`tor_client` field, async `GrpcIndexer::new`) and a roundtrip unit test added.

### DIFF
--- a/zingolib/CHANGELOG.md
+++ b/zingolib/CHANGELOG.md
@@ -10,6 +10,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Deprecated
 
 ### Added
+- `lightclient::LightClient::from_bytes` constructor — creates a `LightClient` by
+  deserializing wallet bytes from memory via `std::io::Cursor`, without reading any file.
+  Intended for mobile platforms (iOS/Android) where the native layer owns all file I/O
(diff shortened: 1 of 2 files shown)


dorianvp (approved):
utACK

PR #2417, opened 2026-06-22
New Feature swap op_return

juanky201271:
Fixes:


zingolib/src/lightclient/propose.rs
```diff
@@ -65,7 +81,8 @@ impl LightClient {
             .map_err(ProposeSendError::TransactionRequestFailed)?;
         let _ignore_error = self.pause_sync();
         let mut wallet = self.wallet().write().await;
-        let proposal = wallet.create_send_proposal(request, account_id)?;
+        // send-all is a self-spend: route_via_ephemeral has no use case here, so single-hop.
```
Oscar-Pepper: what is meant by "send-all is a self-spend" here? send all in most cases is sending all funds to another recipient right?

juanky201271: Yes, this is confusing, fixed it.

zancas:
Superseded by the owned-spend-pipeline arc: ADR 0010 (PR #2469) routes OP_RETURN through the in-tree proposal types, and the zcb-swap approach this branch took was abandoned with that decision.

PR #2428, opened 2026-07-02
feat(ironwood): implement poc migration flow

dorianvp:
Fixes:


pepper-sync/src/wallet/traits.rs
```diff
@@ -359,6 +386,34 @@ pub trait SyncShardTrees: SyncWallet {
...
+                // ironwood tree without one at the truncation height. When
+                // the tree holds no data, resetting it to empty is lossless
+                // and avoids forcing a full rescan on the first reorg after
+                // upgrade.
+                if shard_trees
+                    .ironwood
+                    .store()
+                    .get_shard_roots()
+                    .expect("infallible")
+                    .is_empty()
```
Oscar-Pepper: this doesnt look correct to me.

1. ironwood should follow the same logic as to syncing sapling in the pre-orchard portion of the chain.
2. on creation, shard trees have a 0 height checkpoint inserted so they are always non-empty
3. `get_shard_roots` is not related to checkpoints. there may be shard roots in the tree due to the initial `get_subtree_roots` RPC before scanning started, and no checkpoints may have been added yet.

Oscar-Pepper: i think we can just copy the orchard path here, unless there is an issue i am missing?

pepper-sync/src/lib.rs
```diff
@@ -129,8 +129,8 @@ use zcash_protocol::ShieldedPool;
 use zcash_protocol::consensus::BlockHeight;
 
 use crate::wallet::{
-    NoteInterface, OrchardNote, OrchardShardStore, SaplingNote, SaplingShardStore,
-    WalletTransaction,
+    Ironwood, IronwoodNote, NoteInterface, Orchard, OrchardNote, OrchardShardStore, Sapling,
+    SaplingNote, SaplingShardStore, WalletTransaction,
 };
 
 pub(crate) trait SyncDomain {
```
Oscar-Pepper: I would prefer to keep SyncDomain and the structs that were created for SyncDomain (Sapling etc.) together.. can we also move SyncDomain to wallet mod?

zingolib/src/wallet/balance.rs
```diff
@@ -545,8 +601,19 @@ impl LightWallet {
             Err(BalanceError::KeyError(KeyError::NoViewCapability)) => Ok(Zatoshis::ZERO),
             Err(e) => Err(e),
         }?;
+        // Zero while ironwood notes carry no positions, which is right
```
Oscar-Pepper, line 604: why would compact block scanning fail to find positions for ironwood?

zingolib/src/wallet/disk.rs
```diff
@@ -45,9 +45,17 @@ use pepper_sync::{
...
     /// Changes in version 41:
     /// `ChainType` serialized as u8 instead of string to decouple from fmt::Display and reduce bytes stored.
+    ///
+    /// Changes in version 42:
+    /// Optional Orchard→Ironwood migration section appended (see
+    /// [`crate::wallet::migration::store`]; the section carries its own inner
+    /// version).
+    ///
+    /// Changes in version 43:
+    /// `allow_v6_transactions` bool appended after `min_confirmations`.
```
Oscar-Pepper: if v42 is never merged to dev, can we just use v42 instead of incrementing again to v43?

zingolib/src/wallet/summary.rs
```diff
@@ -106,6 +106,27 @@ impl LightWallet {
                     })
                     .collect::<Vec<_>>();
 
+                let ironwood_notes = transaction
```
Oscar-Pepper: id prefer this was at the top so the protocols are in order

zingolib/src/wallet/summary.rs
```diff
@@ -154,6 +175,31 @@ impl LightWallet {
                         }
                     })
                     .collect::<Vec<_>>();
+                let outgoing_ironwood_notes = transaction
```
Oscar-Pepper: id prefer this was at the top so the protocols are in order

zingolib/src/wallet/summary.rs
```diff
@@ -219,9 +265,11 @@ impl LightWallet {
                     orchard_notes,
                     sapling_notes,
                     transparent_coins,
+                    ironwood_notes,
```
Oscar-Pepper: id prefer this was at the top so the protocols are in order

zingolib/src/wallet/summary.rs
```diff
@@ -219,9 +265,11 @@ impl LightWallet {
...
                     outgoing_orchard_notes,
                     outgoing_sapling_notes,
                     outgoing_transparent_coins,
+                    outgoing_ironwood_notes,
```
Oscar-Pepper: id prefer this was at the top so the protocols are in order

zingolib/src/wallet/summary.rs
Oscar-Pepper: alternatively, the convention seems to be earlier first so the other three can be reversed

zingolib/src/wallet/zcb_traits.rs
```diff
@@ -802,6 +883,44 @@ impl InputSource for LightWallet {
...
+            .activation_height(NetworkUpgrade::Nu6_3)
+            .is_some();
+        let selected_ironwood_notes: Vec<IronwoodNote> =
+            if ironwood_active {
+                let exclude_ironwood: Vec<OutputId> = exclude
+                    .iter()
+                    .filter(|&note_id| note_id.pool_type() == PoolType::IRONWOOD)
+                    .map(|note_id| OutputId::new(note_id.txid(), note_id.output_index()))
+                    .collect();
+                let include_potentially_spent = matches!(
```
Oscar-Pepper: this is copying code above which was implemented incorrectly, we should not use the include potentially spent bool until we decide on the API for it, but it is not to do with the TargetValue::Allfunds variant

Oscar-Pepper: we should set include potentially spent funds to false for now

zingolib/src/mocks.rs
```diff
@@ -646,3 +646,43 @@ pub mod proposal {
...
+
+    use zcash_protocol::consensus::BlockHeight;
+
+    use crate::wallet::migration::{BroadcastClient, BroadcastError};
+
+    /// Records every submission; fails with a transport error while `fail`
+    /// is set (the raw transaction is then not consumed, mirroring the real
+    /// client's transient-failure contract).
+    #[derive(Default)]
+    pub struct MockBroadcastClient {
```
Oscar-Pepper, line 663: ideally mocks would not be part of crates public API but I guess this is intended for use in integration tests?

zingolib/src/wallet/send.rs
```diff
@@ -100,7 +113,24 @@ impl LightWallet {
         let scan_ranges = self.sync_state.scan_ranges();
 
         match N::SHIELDED_PROTOCOL {
-            ShieldedPool::Ironwood => todo!(), // FIXME: implement ironwood
+            // The Ironwood pool begins at NU6.3 activation (or the wallet
+            // birthday if later). Its shard ranges are empty until the
+            // server serves ironwood subtree roots, in which case the whole
+            // range from the pool's beginning to the anchor must be scanned.
```
Oscar-Pepper: this comment is incorrect / misleading.

we can copy the orchard logic here, `check_note_shards_are_scanned` already accounts for the case where the start block of a shard range containing a note is below the wallets birthday.

zingolib/src/wallet/send.rs
```diff
@@ -100,7 +113,24 @@ impl LightWallet {
...
+            ShieldedPool::Ironwood => {
+                use zcash_protocol::consensus::{NetworkUpgrade, Parameters as _};
+                let ironwood_birthday = self
+                    .chain_type
+                    .activation_height(NetworkUpgrade::Nu6_3)
+                    .map_or(birthday, |activation| activation.max(birthday));
```
Oscar-Pepper: as mentioned above, these checks are unecessary, we can just copy the orchard logic here. we have already had to solve the issue of wallets syncing pre-orchard or having spendable notes near the birthday (the shard range being checked if its scanned contains the birthday). ironwood will be the same.

zingolib/src/wallet/send.rs
Oscar-Pepper: oh on second pass I realise I may have missed a detail here. is this comment suggesting that ironwood may go live without serving the subtree roots until a later date!?

Oscar-Pepper: even in this case I think we dont need to overcomplicate... we should just just make wallets sync the whole chain from birthday to spend an ironwood note until this subtree data is available IMO

Oscar-Pepper (commented):
i have noticed ironwood has not been added to `set_chain_tip_scan_range` in sync::state mod. i will thoroughly check if it has also been missed elsewhere on my next pass

Oscar-Pepper:
> i have noticed ironwood has not been added to `set_chain_tip_scan_range` in sync::state mod. i will thoroughly check if it has also been missed elsewhere on my next pass

a1f4ccf112451f42331b5d9d5dd2ac15b41dd748

zingolib/src/wallet/send.rs
```diff
@@ -111,55 +111,53 @@ impl LightWallet {
...
+            ShieldedPool::Orchard => self.sync_state.orchard_shard_ranges(),
+            ShieldedPool::Sapling => self.sync_state.sapling_shard_ranges(),
+        };
+        check_note_shards_are_scanned(
+            note_height,
+            anchor_height,
+            pool_start,
+            scan_ranges,
+            shard_ranges,
+        )
```
Oscar-Pepper, line 126: nice DRY and also gets the benefit of only checking the pools range is scanned instead of from wallet birthday

pepper-sync/src/lib.rs
```diff
@@ -125,45 +124,4 @@ pub use sync::sync_status;
...
-    const SHIELDED_PROTOCOL: ShieldedPool = ShieldedPool::Orchard;
-
-    type Note = OrchardNote;
-    type ShardStore = OrchardShardStore;
-
-    fn notes_mut(wallet_transaction: &mut WalletTransaction) -> Vec<&mut Self::Note> {
-        wallet_transaction.orchard_notes_mut()
-    }
-}
+pub(crate) use wallet::SyncDomain;
```
Oscar-Pepper, line 127: minor: not sure the re-export here is necessary

zingolib/src/wallet/migration/params.rs
```diff
@@ -0,0 +1,138 @@
+//! Per-network migration constants, versioned so testnet and mainnet
+//! activations can carry different ratified values.
+
+use crate::config::ChainType;
+
+use super::split::{CANONICAL_PART_FEE, MARGINAL_FEE};
+
+/// Number of zatoshis in one ZEC.
+pub(crate) const COIN: u64 = 100_000_000;
```
Oscar-Pepper, line 9: in zingolib we use "coin" to refer to transparent outputs. can we rename this? a possible suggestion is "ZATS_PER_ZEC"

zingolib/src/wallet/migration/params.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// Bumped when ratified constants replace the provisional ones.
+    pub version: u32,
+    /// Canonical denominations in zatoshis, ordered largest first so a greedy
+    /// decomposition walks it directly.
+    pub denominations: Vec<u64>,
+    /// `DENOM_CAP`: the largest permitted denomination.
+    pub denom_cap: u64,
+    /// `DUST_FLOOR`: the smallest permitted denomination. Leftover value
+    /// strictly below this cannot be migrated as a standard note.
+    pub dust_floor: u64,
```
Oscar-Pepper, line 25: this could be confused with dust which seems to be more related to the "sweep_min" field. could we rename this field? a suggestion could be "denom_floor" IIUC

zingolib/src/lightclient/migrate.rs
```diff
@@ -0,0 +1,802 @@
...
+        let params = MigrationParams::provisional(wallet.chain_type());
+        // Note-splitting fees depend on whether the transactions confirm at
+        // or after NU6.3 activation (the Orchard bundle's cross-address rules
+        // change the action count).
+        let post_activation = match (
+            wallet.sync_state.last_known_chain_height(),
+            wallet.chain_type().activation_height(NetworkUpgrade::Nu6_3),
+        ) {
+            (Some(chain_height), Some(activation)) => chain_height + 1 >= activation,
+            _ => false,
```
Oscar-Pepper, line 97: in the case that `last_known_chain_height" is None we must return an error that there is no sync data. we can only plan migration after the wallet has the initial info from the server after first sync. there is an error variant in SyncStatusError called NoSyncData .. this variant could be copied into WalletError for this case (LightClientError has a WalletError variant)

zingolib/src/wallet/migration/split.rs
```diff
@@ -0,0 +1,933 @@
...
+            .collect(),
+    );
+    rounds
+}
+
+impl crate::wallet::LightWallet {
+    /// The values (zatoshis) of the account's spendable pre-Ironwood (V2)
+    /// Orchard notes, the input to [`plan_migration`].
+    #[allow(clippy::result_large_err)]
+    pub fn migration_note_values(
```
Oscar-Pepper, line 346: can we make this more private? pub(crate)?

zingolib/src/wallet/migration/split.rs
```diff
@@ -0,0 +1,933 @@
...
+            .filter(|note| note.note().version() == orchard::note::NoteVersion::V2)
+            .map(|note| note.value())
+            .collect())
+    }
+
+    /// Builds, proves, signs and records one planned note-splitting
+    /// transaction (Orchard→Orchard self-send). Returns its txid. Broadcast
+    /// is the caller's step.
+    #[allow(clippy::result_large_err)]
+    pub fn build_note_split_transaction(
```
Oscar-Pepper, line 372: can we make this more private? pub(crate)?

zingolib/src/wallet/migration/quantize.rs
```diff
@@ -0,0 +1,176 @@
...
+/// The result of decomposing a value into canonical denominations.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Denominations {
+    /// One entry per Ironwood output to create. Every value is a member of
+    /// [`MigrationParams::denominations`]. Ordered largest first.
+    outputs: Vec<Zatoshis>,
+    /// The leftover below the dust floor that has no canonical denomination.
+    /// Always strictly less than [`MigrationParams::dust_floor`]. The
+    /// migration folds this into the fee instead of creating a non-standard
+    /// note.
```
Oscar-Pepper, line 17: the dust floor is set to 0.001 ZEC in MigrationParams::provisional. this means that value up to 0.5 USD are not migrated and folded into fees? this seems like a lot for the user to pay for migration?

zingolib/src/wallet/migration/split.rs
```diff
@@ -0,0 +1,933 @@
...
+        &mut self,
+        account: zip32::AccountId,
+        planned: &NoteSplitTx,
+    ) -> Result<zcash_primitives::transaction::TxId, crate::wallet::error::WalletError> {
+        self.build_migration_transaction_inner(account, &planned.inputs, &planned.outputs)
+    }
+
+    pub(in crate::wallet::migration) fn get_migration_heights(
```
Oscar-Pepper, line 380: I'm not sure why this is needed as it essentially just wraps `get_target_and_anchor_heights` but it also seems unnecessary that we return Result<Option<X>> because the implementaton of `get_target_and_anchor_heights`  is infallible. maybe it should just return a Result where the None case where there is no sync data is returned as a NoSyncdata error?

zingolib/src/lightclient/migrate.rs
```diff
@@ -0,0 +1,802 @@
...
+    /// blocking thread pool without holding the lock (Phase B), and wallet
+    /// writes + submission happen sequentially under the lock again (Phase C).
+    /// The due-part broadcast loop, optionally narrowed to a single part so
+    /// catch-up can sequence sends with spacing.
+    ///
+    /// Proving is parallelised across all due parts via
+    /// [`tokio::task::spawn_blocking`]: wallet reads happen under the write
+    /// lock (Phase A), all Halo2/Groth16 work runs concurrently on the
+    /// blocking thread pool without holding the lock (Phase B), and wallet
+    /// writes + submission happen sequentially under the lock again (Phase C).
```
Oscar-Pepper, line 231: comment duplicated

zingolib/src/wallet/migration/parts.rs
```diff
@@ -0,0 +1,958 @@
...
+        let activation = self
+            .chain_type
+            .activation_height(NetworkUpgrade::Nu6_3)
+            .ok_or_else(|| {
+                WalletError::MigrationBuild("NU6.3 has no activation height".to_string())
+            })?;
+        if boundary < activation {
+            return Err(WalletError::MigrationBuild(format!(
+                "part boundary {boundary} is below the NU6.3 activation height {activation}"
+            )));
```
Oscar-Pepper, line 455: open question: do we need to add protection to stop any migration from being run until nu6.3 is active? if migration is run when we release pre-nu6.3 will it keep hitting this error?

zingolib/src/wallet/migration/parts.rs
```diff
@@ -0,0 +1,958 @@
...
+            WalletError::MigrationStateCorrupt("bound note has no tree position".to_string()),
+        )?;
+
+        let Some(root) = self.shard_trees.orchard.root_at_checkpoint_id(&boundary)? else {
+            return Ok(None);
+        };
+        let Some(witness) = self
+            .shard_trees
+            .orchard
+            .witness_at_checkpoint_id_caching(position, &boundary)?
```
Oscar-Pepper, line 372: i think we will need to increase our shard_tree checkpoint count to the bucket modulus otherwise we will not be able to construct a witness for the boundary if its 100 later into the bucket

Oscar-Pepper: otherwise we could just defer to teh `MissedBoundary` skip reason and let it be re-assigned, but this may slow down migration alot if we dont have background tasks running properly

Oscar-Pepper: on second thought storing all this extra data forever for the sake of a specific single migration seems like a mistake. 100 blocks gives well over an hour window for a part migration when the bucket boundary is hit

zingolib/src/lightclient/migrate.rs
```diff
@@ -0,0 +1,802 @@
...
+                    })();
+                    wallet.migration = Some(state);
+                    wallet.save_required = true;
+                    bind?;
+                    wallet.refresh_part_witnesses()?;
+                }
+
+                let client = self.migration_broadcast_client();
+                let sent = self.broadcast_due_parts_with(&client).await?;
+                self.await_migration_confirmations(&sent).await?;
```
Oscar-Pepper, line 699: this fn is written as if it is tracking note-splitting confirmations but the surrounding code is in the context of the sending of parts

Oscar-Pepper (approved):

PR #2431, opened 2026-07-03
build: point infrastructure deps at remove_unnecessary_deps branch

zancas:
Pin zcash_local_net and zingo_test_vectors to the head of zingolabs/infrastructure#276, which prunes ~130 transitive dependencies (zebra-*, rocksdb, jsonrpsee, bindgen chains) and introduces the zingo-consensus crate.

That branch drops the zcash_protocol::PoolType re-export and speaks its own zingo_consensus::{MinerPool, ActivationHeights} types, which are distinct from the zingo_common_components types zingolib uses. Keep the scenario API on the wallet-side vocabulary and convert only at the zcash_local_net boundary: zingolib_testutils::scenarios gains private converters plus a public validator_activation_heights() helper, now used at every call site in zingolib_testutils and libtonode-tests. The reverse heights conversion asserts nu6_3 is unset, since the wallet-side type cannot represent it.

Add CONTEXT.md capturing the MinerPool/PoolType and dual ActivationHeights terminology boundary.

Re-pin to a release tag once infrastructure#276 merges.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -1,0 +1,28 @@
+# Context
+
+Glossary of terms as used in this repository.
+
+## MinerPool
+
+The pool a test validator mines block rewards to (transparent, Sapling, or
+Orchard). An infrastructure-side concept owned by the `zingo-consensus` crate
+(zingolabs/infrastructure); it describes validator configuration, not wallet
+state.
+
+Not to be confused with **PoolType**.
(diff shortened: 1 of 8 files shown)


zancas:
The good news is that all the tests pass!   The bad news is that none of them tests against zebrad.

PR #2435, opened 2026-07-06
chore(makers): add test-set front door and workbench tooling crate

zancas:
Non-invasive, ergonomic test tooling.

Run:

`makers test help`

- tools/workbench: standalone std-only dev-tooling crate (zaino pattern, own workspace, one binary per src/bin file). get-rust-version reads and validates rust-toolchain.toml's channel as the single source of truth for RUST_VERSION, replacing the sed script in Makefile.toml.
- Makefile.toml: `makers test [packages|live|libtonode|zingo-cli]` front door (default: packages). The live-package list (libtonode-tests, zingo-cli) is defined once; `packages` derives its exclusions and `live` its selections from it. All sets delegate to the container-test engine and forward extra nextest args.


dorianvp (approved):
utACK!

PR #2438, opened 2026-07-06
define tests against upgraded infrastructure, NU6.3 enabled zainod, and NU6.3 enabled zebrad

### LINKED ISSUE #2439: In order to support ironwood in zaino, it was helpful to remove dependencies on zcash* and zebra* from infrastructure
This is because those crates are also directly depended on by zaino.   Depending both directly and indirectly on the same crates created a significant maintenance burden, and worse created the opportunity for different version of the same dep to clash in the dependency tree.

The solution was to eliminate the depedencies from infrastructure where they were minimally useful to begin with.

This means that zingolib must update to the new narrow-attack-surface infrastructure crate.

zancas:
Builds atop, and requires #2437 merge that first.

Fixes: #2439 

230 / 247 ran on dev when I started


zancas:
230 / 247 pass at this point.

Rather than debugging these in place, I am going to invest in extending the mock framework.

Issue #2442, opened 2026-07-08
silent spend bug


zancas:
## Symptom

Sends fail with an orchard double-spend rejection from the validator:

```
duplicate nullifier: Nullifier(0x...), in finalized state: true
```

Every retry builds a new transaction around the same input note and is rejected again, while history accumulates unmineable "sent" entries. The wallet's balance silently includes a note that was spent on chain long ago. Only a from-birthday rescan recovers. Observed in the wild from zingo-mobile on mainnet (2026-07-07).

## Root cause

A note's `spending_transaction` field is the wallet's **only durable record** of an on-chain spend. Spend detection *consumes* the chain-side observation (`detect_shielded_spends` removes the matched nullifier-map entry), cleanup prunes the map behind the fully-scanned frontier, and `Scanned` ranges are never revisited. So once the spending transaction's block has been scanned, that one field cannot be reconstructed by any forward sync.

`set_transactions_failed` nevertheless permitted the `Confirmed -> Failed` transition and unconditionally called `reset_spends`, which nulls the field. The result is a **phantom unspent note**: the wallet re-offers a spent note as spendable, forever.

The damage is permanent **iff** sync has raced ahead of the spending transaction — its block behind the fully-scanned frontier, which is exactly when its record is `Confirmed`. If the reset lands before that block is scanned, the next scan re-observes the nullifier and heals the wallet.

The false verdicts that trigger the reset are structural: every "this transaction failed" decision (transmit retry exhaustion, expiry check) is made from local evidence only — the wallet never asks the server whether the txid is actually in the mempool or a block. A delivered transaction whose response was lost is judged failed by its own retries.

## Fix

`set_transactions_failed` now refuses txids whose record is `Confirmed`, logging a warning. Reorg invalidation retains the old behavior via `pub(crate) set_transactions_failed_unchecked`, sound there because truncation reopens the affected scan ranges. Landed in 4f87dd658 on `root_cause_and_fix_silent_spend_bug`, pinned by three unit tests in `sync::test::spend_reset_lifecycle`:

- `failing_a_confirmed_transaction_must_not_destroy_the_spend_observation` (the enforced invariant)
- `reset_before_scan_heals_on_spend_detection` (the self-healing pre-scan sequence)
- `scanned_transaction_overwrites_failed_record` (the record replacement that completes healing)

## Follow-up (out of scope for the guard)

[...]

PR #2443, opened 2026-07-08
Fix silent spend bug

zancas:
Fixes:  #2442

Note that this PR is not into dev, rather its atop the test upgrade branch so I could leverage the new tooling.

I am likely to pull more bug-fixes in here as I go.

### DIFF
--- a/pepper-sync/src/sync.rs
+++ b/pepper-sync/src/sync.rs
@@ -962,7 +962,40 @@ pub fn reset_spends(
 /// Sets transactions associated with list of `failed_txids` in `wallet_transactions` to `Failed` status.
 ///
 /// Sets the `spending_transaction` fields of any outputs spent in these transactions to `None`.
+///
+/// Transactions with `Confirmed` status are skipped with a warning. A mined transaction
+/// cannot fail — only a reorg can un-mine it, and reorgs are handled by truncation, which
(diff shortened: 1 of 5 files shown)


pepper-sync/src/sync.rs
```diff
@@ -962,9 +962,42 @@ pub fn reset_spends(
...
+/// `spending_transaction` field is the wallet's only durable record of the on-chain spend
+/// (the nullifier-map entry was consumed at detection and its block is behind the
+/// fully-scanned frontier), so resetting it here would create a permanent phantom unspent
+/// note that no forward sync can correct.
 pub fn set_transactions_failed(
     wallet_transactions: &mut HashMap<TxId, WalletTransaction>,
     failed_txids: Vec<TxId>,
+) {
+    let (confirmed_txids, failable_txids): (Vec<TxId>, Vec<TxId>) =
+        failed_txids.into_iter().partition(|txid| {
```
Oscar-Pepper, line 978: it we are using partition here we will need to sort the failed txs to have confirmed txs first as this isnt enforced and this logic depends on it. its maybe safer to just filter for unconfirmed txs

zancas: That partition predicate partitions on:

 "transaction.status().is_confirmed()" 

Doesn't the result in a "Confirmed" and "everything else" partition?

Issue #2447, opened 2026-07-09
pepper-sync's subtractive darkside_test feature disables transparent-address discovery for every co-built crate


zancas:
## Symptom

`zingolib lightclient::mock_chain_tests::from_t_z_o_tz_to_zo_tzo_to_orchard` fails deterministically in any multi-package nextest invocation whose build graph includes `darkside-tests` (`makers test packages`, the hierarchy-test packages phase, `cargo nextest run --workspace`), while passing in every single-package invocation (`-p zingolib`, host or container, any feature set). The failure is a zero transparent balance after syncing a chain whose funding block pays the wallet's transparent address.

## Mechanism

`darkside-tests` depends on `zingolib` with `features = ["darkside_tests"]`, and `zingolib`'s `darkside_tests` feature enables `pepper-sync/darkside_test`. Cargo unifies features across a single build invocation, so co-building `darkside-tests` enables `pepper-sync/darkside_test` for **every** crate in that build — including the `zingolib` unit-test binary.

The feature is **subtractive**: `pepper-sync/src/sync.rs:383` wraps the entire `transparent::update_addresses_and_scan_targets` call in `#[cfg(not(feature = "darkside_test"))]`, and the same cfg removes the `transparent` module, witness code paths (`witness.rs`, five sites), transparent key derivation (`keys/transparent.rs`, five sites), client fetch paths (`client.rs`, `client/fetch.rs`), scan-state handling (`sync/state.rs`), and swaps a scanner branch (`scan/transactions.rs:100/107`). With the feature unified on, sync completes "successfully" without ever requesting transparent-address transactions, so purely-transparent receipts are silently never detected.

Cargo features are required to be additive; this one removes functionality, so its enablement by an unrelated sibling package silently degrades wallet behavior everywhere in the co-built graph. The mock-indexer twin was merely the first co-built test that depends on transparent discovery — a bare `makers test` (whole workspace in one invocation) would strip transparent discovery from the libtonode live suite the same way.

## Evidence

Both directions reproduced on one host, same tree:

```
cargo nextest run -p zingolib -E 'test(from_t_z_o_tz_to_zo_tzo_to_orchard)'
# PASS (113s); mock taddr-request ledger POPULATED, wallet transactions: 1

cargo nextest run -p zingolib -p darkside-tests -E 'test(from_t_z_o_tz_to_zo_tzo_to_orchard)'
[...]

Issue #2440, opened 2026-07-06
add_subtree_roots: container runs saw a truncated sapling subtree-root stream (wallet 1096 vs server 1128) — evidence points to a stuck backend behind zec.rocks


zancas:
## Symptom

Two consecutive `makers test all` container runs failed `libtonode-tests::sync add_subtree_roots` with **identical** numbers:

```
no of sapling shard roots wallet: 1096, no of sapling shard roots server: 1128
```

The wallet (pepper-sync's own subtree-root fetch, over the lightclient's connection) ended 32 sapling shards short of what a parallel fresh connection saw from the same URI (`https://zec.rocks:443`), in the same test process.

## Isolation matrix

| Code | Environment | Result |
|---|---|---|
| new branch (2 runs) | container | FAIL — wallet 1096 vs server 1128, identical both runs |
| `dev` | host | PASS (1128/1128) |
| same new branch | host | PASS (1128/1128) |
| same new branch, ~1 day later | container | PASS (1128/1128, probe: 3 connections all healthy) |

So: not a code regression (same commit passes on host and, later, in the container), and not a persistent environment property.

## Why the determinism matters

The wallet landed on **exactly 1096 twice**. That rules out:

- **Timing/throughput-based stream cuts** (request timeout, slow container network): those slice at a *time* boundary, so the count would jitter between runs.
- **A fixed byte/message cap on the container's network path** (proxy buffer, frame limits): a static cap would also have fired in the later container run, which pulled all 1128 roots (~96 KB) through the same path cleanly.

What survives: a backend behind the zec.rocks load balancer serving a **stable, internally consistent, stale view** — a node *stuck* (not progressing) at 1096 finalized sapling shards. This explains the clean stream end (it sent everything it had), the exact repetition (its state wasn't moving; connection routing pinned the wallet's channel to it both times), the parallel fresh connection seeing 1128 (healthy backend), and the later all-clear (node caught up or left the pool). The gap itself is diagnostic: 32 shards ≈ 2.1M note commitments ≈ months of chain — worth reporting to the zec.rocks operators.

A contributing client-side weakness made the anomaly silent: pepper-sync's `get_subtree_roots` accepts a clean-but-premature stream end as complete (its retry logic only fires on `DeadlineExceeded` / EOF-decode errors), so a short stream becomes wrong wallet state instead of an error.

## Changes

[...]

PR #2448, opened 2026-07-09
Complete test upgrade against, zebrad, ready for ironwood-migration merge.

zancas:


libtonode-tests/tests/sync.rs
```diff
@@ -326,56 +321,123 @@ async fn sync_test() {
...
+    // the chain is longer than the cycles alone and heights below
+    // tip − MAX_REORG_ALLOWANCE are legitimately checkpoint-free
+    // (pruned; that pruning having run is part of what this exercises).
+    let tip = local_net.validator().get_chain_height().await;
+    // Retention observed: the last MAX_REORG_ALLOWANCE checkpoints
+    // INCLUDING the tip (tip−99..=tip). Whether the finalized fork
+    // point at tip−100 also needs its checkpoint (a maximal truncation
+    // lands there) is an open question for the pepper-sync owners; if
+    // the answer is yes, extend this window down by one and the test
+    // correctly goes red until retention grows.
```
Oscar-Pepper, line 437: the answer to this is no. the shard tree only stores MAX_CHECKPOINTS which is set to 100 currently which is tip-99..=tip

zancas: OK... is that bound a function of any other parameter?

Oscar-Pepper: although this should be looked at. it seems we should be storing 1 more checkpoint than the maximum re-org i.e. 101

libtonode-tests/tests/sync.rs
```diff
@@ -404,3 +466,167 @@ async fn store_all_checkpoints_in_verification_window() {
...
+/// Prints, for three fresh connections, how many sapling subtree roots the
+/// stream yields and the chain height at which it ends — then the count
+/// pepper-sync's own fetch path lands in the wallet. If a short stream's last
+/// completing height is months old, the connection reached a lagging backend
+/// behind the load balancer; if counts differ between connections to a
+/// current backend, the stream is being cut mid-flight.
+///
+/// Run in the environment under investigation with output visible, e.g.:
+///   makers test -p libtonode-tests \
+///     -E 'test(diagnose_subtree_root_stream)' --run-ignored all --no-capture
```
Oscar-Pepper, line 498: i dont seem to get the incorrect behaviour mentioned in this comment when i run this test

pepper-sync/src/sync.rs
```diff
@@ -523,18 +530,38 @@ where
...
+                        }
+                        if unprocessed_mempool_transactions_count.load(atomic::Ordering::Acquire)
+                            > 0
+                            || shutdown_poll_started.elapsed()
+                                >= std::time::Duration::from_secs(1)
+                        {
+                            break false;
+                        }
+                        tokio::time::sleep(std::time::Duration::from_millis(25)).await;
+                    };
```
Oscar-Pepper, line 561: CHANGE REQUEST: i think this change introduces 2 issues:
1. in the case the chain is already fully synced the mempool will close immediately... this 1 sec wait that was removed is specifically to prevent this by allowing time for the mempool connection and for the indexer to start serving mempool txs. the comment mentions it has been streaming since session start but in this case there has not been multiple pollls as suggested. it can be set to shutdown on the first sync loop.
2. this change introduces a 1 second timeout on processing *every* mempool transaction that has been received. the purpose of the unprocessed_mempool_transactions_count is to make sure sync continued running until all mempool txs have been processed and shutdown should wait until this is complete

Oscar-Pepper: im not sure what the issue with the original code is that is trying to be solved here

zingolib/src/lightclient/mock_chain_tests.rs
```diff
@@ -0,0 +1,819 @@
...
+/// feature deletes transparent-address discovery at compile time, and
+/// cargo feature unification enables it for every crate co-built with
+/// darkside-tests (`makers test packages`, `--workspace`), so this
+/// test's purely-transparent step-1 funding is silently never detected
+/// in those invocation shapes while passing in `-p zingolib` ones. Runs
+/// green via `--run-ignored` in a single-package invocation. Un-ignore
+/// when the feature becomes runtime configuration.
+#[ignore = "zingolabs/zingolib#2447: pepper-sync's subtractive darkside_test feature, when \
+            unified into multi-package builds, compiles out the transparent-address discovery \
+            this test's funding depends on"]
```
Oscar-Pepper, line 633: if darkside is being replaced by the mockchain, maybe we should just delete the darkside test crate and remove the darkside feature so we dont get these feature unification issues. This comment suggests that makers is setup to compile zingolib in the context of the workspace *including* darkside which is really bad. i solved this with the `run_workspace_tests.sh` script which excludes darkside in the test build but runs darkside in a seperate compilation

zingolib/Cargo.toml
```diff
@@ -67,6 +76,8 @@ zingo-status = { workspace = true }
 zingo_common_components = { workspace = true }
 zingo_test_vectors = { workspace = true, optional = true }
 zip32.workspace = true
+zaino-proto = { git = "https://github.com/zingolabs/zaino.git", rev = "2e2816cb3755e7ff4d18bed1a96b95dc07c1549c", version = "0.2.0", default-features = false, features = ["grpc_proxy_server"], optional = true }
```
Oscar-Pepper, line 79: just for context, what does zaino-proto offer that lightwallet-protocol doesnt?

zingolib_testutils/src/scenarios.rs
```diff
@@ -243,72 +527,91 @@ pub async fn unfunded_client(
...
 
-    let mut faucet = client_builder
-        .build_faucet(true, configured_activation_heights)
-        .await;
+    let mut faucet = client_builder.build_faucet(true).await;
 
-    if matches!(DefaultValidator::PROCESS, ProcessId::Zebrad) {
+    // A replayed chain already contains the shielded-funds transactions;
+    // the freshly built faucet wallet recovers them by sync below.
+    if !replayed && matches!(DefaultValidator::PROCESS, ProcessId::Zebrad) {
```
Oscar-Pepper, line 575: CHANGE REQUEST: this zebrad_shielded_funds should be removed now that zebra supports mining to shielded

zingolib_testutils/Cargo.toml
```diff
@@ -17,8 +19,19 @@ zingo_common_components.workspace = true
...
+zcash_primitives.workspace = true
+zcash_protocol.workspace = true
+
 # utilties
 portpicker.workspace = true
 tempfile.workspace = true
 http.workspace = true
 zip32.workspace = true
+tokio = { workspace = true, features = ["net", "time", "rt", "io-util", "sync", "macros"] }
+zingo-netutils = { version = "5.0.1", path = "../zingo-netutils" }
```
Oscar-Pepper: i often avoid using this feature of cargo.toml to list a crates.io and path as its been misunderstood in the past. im happy to keep this but to be clear, this is building with path locally but with crates.io during publishing. i.e. the published build is different to the local build.

Oscar-Pepper (changes requested):
this is amazing!

pepper-sync/src/sync.rs
Oscar-Pepper: i expect the 1 second wait for each test was expensive... however i want to make sure that (a) we dont miss mempool txs when the wallet is already fully synced (this will be solved by continuous sync which should be implemented soon) and (b) we dont shutdown sync before the revceived mempool txs have been fully processed and scanned

libtonode-tests/tests/sync.rs
zancas: When this was originally committed the error state was reliably toggled "on" by invoking with podman.

That's "the environment under investigation".. I will update the doc-comment.

zancas: Doc-comment updated as promised: #2452, against `pre_ironwood` (this PR's branch), so it folds into this diff once merged.

The short version, since my earlier reply deserved sharpening: the container-only record stands — two podman runs failed at exactly 1096 sapling shards while host runs of the same commit passed — but that records *where* it happened, not a mechanism. Within the same failing container process, a parallel fresh connection to the same URI saw all 1128 roots (#2440's isolation matrix), so the healthy backend was reachable from inside the container, and a backend's state is client-independent. Why the wallet's channel pinned to the stuck backend both times is #2440's open question.

Also relevant to your reproduction attempt: a day-later container run pulled all 1128 cleanly, and the 9d40495b9 resume defense means a cut stream now self-repairs — so not observing the incorrect behaviour today is the expected result, whichever environment you run in.

pepper-sync/src/sync.rs
zancas: See the exhaustive "drain_verdict" logic in the follow-up "address_review_comments" PR:

https://github.com/zingolabs/zingolib/pull/2452/changes#diff-6935e3d62907ae6bf5e4c8e925989d176c7ba9d2362b4b8fa06341c4a76f76e4R1054

zingolib/src/lightclient/mock_chain_tests.rs
zancas: Yes!   I have a draft PR that sketches out exactly this.  I thought that removing full crates should be held out into a separate PR, even though, as you say it's a logical follow-up that immediately simplifies the feature interaction graph.

zingolib_testutils/src/scenarios.rs
zancas: Agreed — this function needed to go in its current form. The name claims a shielded-funding workaround, and zebrad mining coinbase directly to orchard makes that claim stale. The minimal correction turned out to be a rename rather than a removal, because two jobs survive the workaround this function once was: it offloads a fixed amount from the faucet so the spendable balance stays predictable for test assertions, and it first mines two separation blocks to clear the NU6.1/NU6.2 activation ladder and keep note selection off the tip note — both constraints observed as live zebra mempool rejections ("incorrect consensus branch id", "could not validate orchard proof"). Removing it outright would break the funded-faucet balance expectations and reintroduce those hazards. Commit 84e20ea52 (on the follow-up PR #2452) renames it to `normalize_shielded_faucet_balance` so the name states the purpose. One side note for precision: zebrad mines to orchard but not to sapling, so "supports mining to shielded" holds per-pool.

zingolib/src/lightclient/mock_chain_tests.rs
zancas: The draft PR is now visible: #2453, the follow-up package-simplification work. It retires the `darkside-tests` and `zingo-testutils` crates outright, leaves the orphaned `darkside_test`/`darkside_tests` features declared but unenabled so the feature-unification poison from #2447 is cured, and records the decision in ADR 0005. The reorg protections the crate nominally provided now run offline in seconds against the mock indexer.

zingolib/Cargo.toml
zancas: Nothing..  as follow-up work we need to unify the provenance chain for these two.

zingolib_testutils/Cargo.toml
zancas: Nice catch!  Fixed.

libtonode-tests/tests/sync.rs
zancas: Agreed, and now tracked: #2454. The short version: what this thread calls `MAX_CHECKPOINTS` is pepper-sync's `MAX_REORG_ALLOWANCE` (100) doing double duty as shard-tree checkpoint retention, and the two jobs differ by exactly one — tolerating a reorg of depth *d* requires *d + 1* retained checkpoints, so retention must be computed from the reorg bound rather than restated as the same literal. The authoritative upstream anchor is `zcash_client_backend`'s `PRUNING_DEPTH` (100) — a library pepper-sync already depends on, whose docs define the pruning floor as `chain_tip - (PRUNING_DEPTH - 1)`. Both numbers must derive uniquely from it: retention `= PRUNING_DEPTH`, and `MAX_REORG_ALLOWANCE = PRUNING_DEPTH - 1`. Validator-side constants are not the anchor: pepper-sync's dependency set must not grow, and zebra's `MAX_BLOCK_REORG_HEIGHT` is local node policy, already widened to 1000 on zebra main. `PRUNING_DEPTH` is `pub(crate)` today, so the derivation starts as a documented pin. We plan to close #2454 in follow-up work after this PR and its satellites land.

Oscar-Pepper (approved):

PR #2451, opened 2026-07-11
fix: treat duplicate-in-mempool rejection as transmission success

### LINKED ISSUE #2450: send: retry loop converts duplicate-in-mempool rejection into set_transactions_failed on a live transaction
## Summary

`transmit_transactions` retries a timed-out `send_transaction` by
re-submitting the identical raw transaction, but treats the node's
duplicate rejection — `RPC error -1: transaction already exists in
mempool` — as just another failure. After `MAX_RETRIES` the wallet
calls `set_transactions_failed` on a transaction that is live in the
mempool and will confirm. The retry loop thereby converts a successful
transmission into a false failed state whenever the first submission
lands but its response is slow.

Observed on `pre_ironwood` at `10c2403b5` (zainod 0.6.0-rc.1, zebrad
6.0.0-rc.0, regtest) in a containerized run of 2026-07-11: three tests
failed identically under parallel load
(`concrete::test_scanning_in_watch_only_mode`,
`chain_generics::generate_a_range_of_value_transfers`,
`chain_generics::send_shield_cycle`).

## Mechanism

The loop at `zingolib/src/lightclient/send.rs:162-199`:

1. First `send_transaction` exceeds `DEFAULT_REQUEST_TIMEOUT` (10 s,
   `lightclient.rs:53`) while the transaction has in fact entered the
   validator's mempool.
2. The loop re-submits the identical bytes. Every retry is now
   *guaranteed* to fail: the node correctly answers
   `-1 transaction already exists in mempool`.
3. `MAX_RETRIES = 3` exhausts; the loop calls
   `set_transactions_failed` (`send.rs:187`) and returns
   `TransmissionError::TransmissionFailed`.

The observatory record (pipeline traffic proxies; the instrument of
zingolabs/zaino#1386) shows the full sequence for
`test_scanning_in_watch_only_mode` — original submission unanswered
within the recorded window, then three retries each answered with the
duplicate rejection within ~70 ms:

```
53.722s conn4   ->  8192B sendrawtransaction   <- original: accepted, no reply in window
59.746s conn929 ->  8192B sendrawtransaction   <- retry 1 after the 10s timeout
59.815s conn929 <-   245B                      <- -1 "transaction already exists in mempool"
60.821s conn3   ->  8192B sendrawtransaction   <- retry 2
60.875s conn3   <-   245B
61.885s conn929 ->  8192B sendrawtransaction   <- retry 3
```

## Why this is worse than a test flake

`set_transactions_failed` marks failed a transaction that confirms
shortly after. The wallet's spend tracking then believes the
transaction's inputs are unspent: a subsequent send can select those
notes and be rejected as a double-spend, and balances report a state
the chain does not have. A blind timeout-resubmit is only sound if the
duplicate outcome is treated as what it is — positive confirmation
that an earlier submission succeeded.

## Fix

In the retry loop, recognize the node's duplicate rejection
("transaction already exists in mempool", and its sibling "already in
block chain") and `break Ok(*txid)` instead of counting it as a
failure. The discrimination must currently be a substring match on the
error text because zainod surfaces the rejection as an untyped
`Internal` catch-all — that error-surface gap is filed as
zingolabs/zaino#1392, and the match should upgrade to a typed check
when it lands.

Regression test: the mock indexer
(`zingolib/src/testutils/mock_indexer.rs`) can fabricate the exact
sequence — a `send_transaction` answering with the duplicate rejection
while the mock mempool holds the transaction — asserting
`transmit_transactions` returns `Ok` and the transaction is not marked
failed. Fast-tier; no LocalNet required.

## Related issues in other repos

- zingolabs/zaino#1392 — the same event's error surface: zainod maps
  the routine duplicate rejection through the `unhandled rpc-specific`
  `Internal` catch-all, forcing the string match above.
- zingolabs/zaino#1386 — the observatory instrument these records come
  from, and the established pattern for attaching them to upstream
  reports.

The trigger itself — the validator completing mempool insertion within
~6 s yet the original `sendrawtransaction` going unanswered in the
recorded window — is load-correlated and remains observational; the
byte-level record cannot distinguish a withheld response from the
client having dropped the connection after its own timeout, so no
zebra issue is filed on this evidence.

zancas:
Closes #2450.

## The defect

When a send's first submission lands in the validator's mempool but its
response is lost, `transmit_transactions` re-submits the identical
bytes and the node correctly answers `RPC error -1: transaction already
exists in mempool`. The retry loop counted that rejection as a failure
and, after `MAX_RETRIES`, called `set_transactions_failed` on a
transaction that was live and about to confirm — a spend-tracking lie.
The live shape was three load-dependent libtonode flakes in the
2026-07-11 container run (`test_scanning_in_watch_only_mode` and two
`chain_generics` tests), diagnosed from the observatory traffic records
in #2450.

## The fix

The duplicate rejection is positive confirmation that an earlier
submission succeeded, so the retry loop now breaks `Ok` with the
calculated txid when it sees it ("transaction already exists in
mempool", or its mined sibling "already in block chain"). The
discrimination is a substring match because zainod surfaces the
rejection through an untyped `Internal` catch-all — filed as
zingolabs/zaino#1392 — and the comment in `send.rs` carries that
pointer so the match upgrades to a typed check when it lands.

## The test

The mock chain grows the two validator semantics needed to reproduce
the incident deterministically, as real behaviors rather than test
hooks: resubmitted bytes are rejected with zainod 0.6.0-rc.1's verbatim
message, and a one-shot `lose_next_send_response` fault accepts bytes
into the mempool while failing the reply — exactly the
accepted-but-unanswered submission the observatory recorded.

`mock_chain_tests::send_survives_lost_response_and_duplicate_rejection`
funds a wallet, injects the fault, sends, and asserts the send returns
`Ok`, the transaction is never marked `Failed`, and after mining the
mempool the balance carries the same ZIP-317 arithmetic as the
untampered send test. It was confirmed red against the unfixed loop —
failing with the byte-for-byte production error — and goes green with
this fix. Deterministic, ~14 s, no LocalNet.

### DIFF
--- a/zingolib/src/lightclient/mock_chain_tests.rs
+++ b/zingolib/src/lightclient/mock_chain_tests.rs
@@ -830,6 +830,7 @@ async fn from_t_z_o_tz_to_zo_tzo_to_orchard() {
 /// arithmetic.
 #[tokio::test]
 async fn send_survives_lost_response_and_duplicate_rejection() {
+    use crate::testutils::mock_indexer::LostSendDestination;
     use zingo_status::confirmation_status::ConfirmationStatus;
 
     let mut net = MockNet::launch().await;
@@ -844,7 +845,7 @@ async fn send_survives_lost_response_and_duplicate_rejection() {
     recipient.sync_and_await().await.unwrap();
     check_client_balances!(recipient, o: 100_000 s: 0 t: 0);
 
-    net.chain.write().await.lose_next_send_response = true;
+    net.chain.write().await.lose_next_send_response = Some(LostSendDestination::Mempool);
 
     let txids = from_inputs::quick_send(
         &mut recipient,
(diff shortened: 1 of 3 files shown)


zancas:
@Oscar-Pepper this one is yours to review and merge into `pre_ironwood` when you are satisfied. Context: it closes #2450, the upstream error-surface report is zingolabs/zaino#1392, and the full containerized hierarchy suite ran green (284/284) on 2026-07-11 with both commits in the tree.

Oscar-Pepper (approved):

PR #2457, opened 2026-07-13
Feat/offline mode2

zancas:
Same feature, but the PR is explicitly against DEV.


zancas:
Same as #2371 but explicitly against dev.

Oscar-Pepper (approved):

Issue #2455, opened 2026-07-13
Offline signing needs the no-expiry sentinel: upstream expiry control, pepper-sync sentinel handling, and an offline-to-online audit of Calculated transactions


zancas:
## Decision

Transactions calculated by an Indexerless client (offline signing, ADR 0006, PR #2371's `calculate_proposal`/`transmit_calculated` split) use the **no-expiry sentinel**: `nExpiryHeight = 0`. A transaction signed against the wallet's stale chain view must not be dead on arrival merely because the session stayed offline longer than the 40-block default window.

(Vocabulary note: the client's send attempt is **Transmission**, per the glossary; "broadcast" is reserved for a distinct future concept — a memo multiple recipients can decrypt — and earlier references here to a "broadcast" step mean Transmission.)

## What blocks it

Expiry is fixed inside `zcash_primitives`' `Builder::new` (`target_height + DEFAULT_TX_EXPIRY_DELTA`), and `zcash_client_backend::data_api::wallet::create_proposed_transactions` — the call path `LightWallet::calculate_transactions` uses — exposes no expiry control. Implementing the sentinel therefore requires an upstream librustzcash change (an expiry policy parameter on the create path), or bypassing `zcash_client_backend`'s builder orchestration, which we do not want. Until then, offline-calculated transactions inherit the stale-tip expiry, and `calculate_proposal`'s doc comment says so.

## What the sentinel requires in this workspace once unblocked

1. **pepper-sync must treat `nExpiryHeight = 0` as never-expiring.** The sync engine marks unconfirmed transactions `Failed` once `last_known_chain_height >= expiry_height` (`pepper-sync/src/sync.rs`, the expiry check near `set_transactions_failed`). Zero compares below every chain height, so today a sentinel transaction would be failed on the FIRST sync after connecting — precisely the offline-to-online transition the feature exists for.
2. **The offline-to-online transition must audit Calculated transactions.** A never-expiring signed transaction escapes the expiry-based cleanup that bounds every other unconfirmed transaction, so when a session leaves the Indexerless state (`set_indexer_uri`, or the first sync after), wallet transactions with `Calculated` status need a check: inputs already spent elsewhere (conflict — surface or fail the record), or still transmittable (surface for a Transmission decision). Nothing may silently linger forever.
[...]

PR #2458, opened 2026-07-13
feat: offline signing with no expiry height, exposed as zingo-cli calculate and transmit

zancas:
Extends Offline mode with the calculate/transmit split that was requested for zingo-cli testing: `calculate` signs the stored proposal with no Indexer, and `transmit` sends the resulting Calculated transactions once an Indexer is available.

## No expiry height

Transactions built by `calculate` deliberately carry **no expiry height** — the ZIP 203 sentinel, `nExpiryHeight = 0`. A transaction signed against an Indexerless wallet's stale chain view must not be expired by the time an Indexer is available to transmit it (#2455). The consequence is framed plainly in the command help, the `calculate` output, and the API docs: these transactions **never expire**; each remains valid to transmit at any later time, until it is transmitted or its inputs are spent by another transaction, so a Calculated record is live value in flight.

Two supporting changes make the sentinel safe:

- **`[patch.crates-io]` to `zingolabs/librustzcash@zingo_expiry_override_zp_0_28_0`.** Expiry is fixed inside `zcash_primitives`' `Builder` and `zcash_client_backend` exposes no parameter for it. The fork branch is the `zcash_client_backend-0.23.0` release tag — every patched crate version matches its crates.io release exactly — plus one additive seam: a guard-scoped expiry-height override read by non-coinbase `Builder::new`, inert unless a guard is live. zingolib installs the guard only inside `create_proposed_transactions`, around the synchronous construction call, under the wallet write lock that serializes builds. The sibling crates are patched to the same commit so the family keeps a single type identity.
- **pepper-sync excludes the sentinel from its expiry check.** Naively compared, every chain height exceeds zero, and a never-expiring transaction would have been marked `Failed` on the first sync after calculation — exactly the offline-to-online transition offline signing exists for.

## CLI surface

- `calculate` — consumes the stored proposal (from `send`, `send_all`, or `shield`), signs offline, stores the transactions with `Calculated` status, and reports the txids alongside an explicit `expiry_height: none` field.
- `transmit [txid ...]` — transmits Calculated transactions: all of them ordered by target height by default, or specific txids in caller order (multi-step proposals, such as TEX sends, must transmit step one first).

The offline signing test now pins `nExpiryHeight = 0`, and parse tests cover both commands. 257 tests pass across zingolib, zingo-cli, and pepper-sync.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -952,8 +952,7 @@ dependencies = [
 [[package]]
 name = "equihash"
 version = "0.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "306286e8dcc39ab3dfceb74c792ce8baffdab90591321d3ffaae64829734c37f"
+source = "git+https://github.com/zingolabs/librustzcash?branch=zingo_expiry_override_zp_0_28_0#701d1e787d1b62f1ab4b167942d3ec11f22936e7"
 dependencies = [
  "blake2b_simd",
  "corez",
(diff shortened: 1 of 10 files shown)


zancas:
Superceded by #2460

PR #2459, opened 2026-07-13
cargo auto-update

zancas:
Fixes:


Oscar-Pepper:
can we not merge this until ironwood migration is complete to avoid dep conflicts setting us back?

PR #2464, opened 2026-07-14
Remove stringly typed errors from zingo

### LINKED ISSUE #2446: Stop returning error prose inside String: typed Results across the public API
## Problem

A few public APIs still report failures *in-band*: they return `String`, and error conditions travel inside that string as prose conventionally prefixed with `Error:`. Downstream consumers (zingo-cli, and zingo-mobile's FFI crate) can then only classify an outcome by sniffing the content of the value — there is no type-level distinction between "here is your data" and "something went wrong."

Concrete example: `LightClient::do_info` returns a `String` that is either server info or error prose; zingo-mobile's `info_server` pipes it verbatim across the FFI, where the platform layers detect failure by checking for the `error` prefix, case-insensitively.

## Why this is a correctness problem

- **The success channel is unparseable in principle.** A `String` that may hold JSON data, human status prose, or error text cannot be interpreted without content-sniffing, and every consumer must re-implement the sniff identically.
- **Valid data can look like an error.** Any payload whose serialization legitimately begins with the sentinel is misclassified. Base64 and JSON payloads can and eventually do start with `error` under case-insensitive comparison; when the misclassified value is wallet data, the consumer wrongly discards a healthy result — a data-loss bug class, not a cosmetic one.
- **The sentinel is fragile.** Anything that rewrites the message — localization, wrapping, reformatting — silently breaks detection, turning real failures into apparent successes.
- **Typed errors are already the norm here.** Most of the API surface returns `Result<T, E>` with rich error enums; the remaining string-returning functions are the inconsistency, and they erase exactly the error structure the library already has.

## Ask

Inventory the public API surface for functions that can return error prose inside a `String` (starting with `do_info`), and migrate each to `Result<T, TypedError>` where `T` is a data type (serde-serializable where consumers want JSON). The goal state: no zingolib API requires a caller to inspect a value's content to learn whether the call succeeded.

## Relationship to zingo-mobile

zingo-mobile's FFI layer is the primary consumer of these strings and has a paired issue to carry typed errors the rest of the way (throwing UniFFI functions, promise rejection instead of resolved error strings). That work can eliminate the sentinel convention at every layer above zingolib, but only if zingolib's own surface stops minting in-band error strings first: zingolabs/zingo-mobile#1151

zancas:
Fixes: #2446

### DIFF
--- a/.agent-plans/dry-commands-and-expiry.md
+++ b/.agent-plans/dry-commands-and-expiry.md
@@ -21,7 +21,12 @@ dev's retarget work superseded). Live PR: #2464.
   derived from it); upstream issue filed asking zcash_protocol to
   export a named constant, to be adopted at a normal version bump.
 
-## In progress (2026-07-14)
+## Done (2026-07-14): sniff elimination and integer ratio
+
+Note: this section's description of `is_complete()` ("preserves the
+refetching-nullifiers nuance the old 99% override encoded") records the
+first implementation, which the review follow-up below superseded —
+completion is now defined by scan-range state alone.
 
 - Eliminate the REPL's in-band error sniffing in `zingo-cli/src/lib.rs`
   (the internal instance of issue #2446's problem; same family as the
@@ -43,7 +48,7 @@ dev's retarget work superseded). Live PR: #2464.
   dev's retarget implementation (c007996aa) and ADR 0008; the ADR 0006
   pointer fix and this file were replayed as the rebase remnant.
 
-## In progress (2026-07-14, #2446 close-out; DRY deferred)
+## Done (2026-07-14): #2446 close-out (DRY deferred)
 
 Executing the three ratified steps on PR #2464 (audit evidence in
 issue #2465): (1) `do_delete` error typed in place (`io::Error`);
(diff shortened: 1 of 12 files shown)


zancas:
**`SyncStatus::is_complete` now defines completion by scan ranges, not by the output ratio.**

The first version compared the output counters, guarded by `total_outputs > 0`. That had two defects: the guard conflates "sync never started" with "nothing to scan", so a wallet with no shielded outputs between birthday and chain height could never report complete; and the denominator is fixed at session start, so mid-session chain growth could let the numerator reach a stale target and fake completion while ranges remained unscanned.

The fix uses the sync task's own terminal condition instead: sync has started (`sync_start_height != 0`, the existing sentinel) and every scan range is `Scanned`. That resolves both defects, and the refetching-nullifiers nuance comes along for free, since `ScannedWithoutMapping` and `RefetchingNullifiers` are not `Scanned`. The output ratio is still what the prompt displays; it just no longer decides completion.

The longhand scan-priority predicates also gained named helpers on `ScanPriority` (`is_scanned()`, `awaits_nullifier_retrieval()`) with no behavior change, and new pepper-sync unit tests pin the completion contract with all counters held at zero, proving scan ranges alone decide it.

zancas:
Merged after extensive agent review.

PR #2466, opened 2026-07-15
feat: add pool info on VTs

dorianvp:
Adds pool info for `ValueTransfers`.


zancas (approved):
Three failing tests also fail in the target.

PR #2419, opened 2026-06-30
feat: ironwood support

dorianvp:
This branch makes the wallet ironwood-capable from the network edge to the command line. The pepper-sync crate scans the third shielded pool — ironwood compact-block actions, a third shard tree, and a third nullifier map — counts it in sync progress and batching budgets, and persists it behind version-gated wallet-format bumps (SyncState v4, NullifierMap v2, TreeBounds, WalletTransaction, and ShardTrees v1), each with migration tests over old blobs. Following ZIP 326, receivers and viewing keys are scoped to the Orchard protocol rather than to a pool, so ironwood shares orchard key material and unified addresses gain no new receiver. On top of that, zingolib extends note selection, balances, summaries, and the `zcash_client_backend` trait implementations to the new pool, riding the ironwood dependency train (`zcash_client_backend` 0.24.0-rc.1, `zcash_primitives` 0.29.0, `orchard` 0.15.0) plus a `[patch.crates-io]` git pin of `lightwallet-protocol` that carries the ironwood proto fields until upstream regenerates and publishes them.

The largest single addition is the consent-bound orchard-to-ironwood migration engine of ZIP 318, in the new `zingolib::wallet::migration` module: deterministic denomination sizing and note splitting (`split`), part records with an explicit state machine (`parts`), bucket-scheduled broadcasts with jitter (`schedule`), versioned persistence (`store`), restart reconciliation (`reconcile`), and an immediate drain path (`drain`). `LightClient` methods orchestrate consent, rounds, catch-up, and status reporting, and `zingo-cli` exposes the corresponding `migrate` commands, with `shield` now targeting the ironwood pool. This work arrived through PRs #2428 and #2429, the summary and value-transfer extensions of #2463 and #2466, the sweep-floor fix of #2468, the zebrad-testing line of #2461, the stable backport of #2473, and #2475, which also carried #2477's scan-progress accounting fix and its regression test.

[...]

PR #2469, opened 2026-07-16
OP_RETURN via an owned spend pipeline: ADR 0010 scoping, dependency trims, and the owned proposal types

zancas:
## LOC breakdown

Diff against the `feat/ironwood` merge-base, split at each file's `#[cfg(test)]` boundary (test modules trail production code in every affected file); `testutils/`, `mocks.rs`, and `tests/` trees count as test wholesale. Comparative migration scaffolding added and deleted within the branch nets out of these figures.

| Package | Production | Tests | Docs & manifests |
|---|---|---|---|
| zingolib | +2,622 / −1,576 | +713 / −380 | +3 / −5 |
| zingo-cli | +20 / −15 | — | +0 / −1 |
| libtonode-tests | — | +92 / −31 | +1 / −1 |
| pepper-sync | +5 / −1 | — | — |
| workspace root | — | — | +281 / −3 |
| **Total** | **+2,647 / −1,592** | **+805 / −411** | **+285 / −10** |

The production additions are the owned pipeline (`zingolib/src/wallet/spend/`: types, plan, build, apply, fee sizing); the production deletions are the zcb trait impls (`zcb_traits.rs`, 1,095 lines), `data/proposal.rs`, and the old propose/calculate plumbing. pepper-sync's +5 is a re-export comment block, no dependency change.

## What this PR does

zingolib takes ownership of its spend pipeline as an in-tree module and removes `zcash_client_backend` from its dependency tree (ADR 0010, in-repo), giving OP_RETURN Data — the swap memo channel for THORChain/MAYAChain deposits — a native seam instead of waiting on an upstream Proposal-API feature that never landed (zcash/librustzcash#2441 closed unmerged). `cargo tree -i zcash_client_backend` now resolves only through pepper-sync, whose serialization-only dependency is explicitly out of scope.

The pipeline is layered functionally, effects at the edges: a pure plan layer (`&LightWallet` → owned `Proposal`), a wallet-pure build layer (proposal + keys + witnesses + provers → signed transactions), and a single apply mutation site. `propose_send`, `propose_send_all`, and `quick_send` gain `Option<OpReturnData>` (a validated ≤80-byte newtype); the payload rides the final transaction — the exposure step of a ZIP-320 TEX pair, or the single step otherwise — and a Shield can never carry it, by construction. The ADR 0008 expiry retarget reduces to a pure field update. Refund addresses are derived at plan time and reserved only at apply, so an abandoned TEX proposal no longer burns an ephemeral index.

Fee correctness is enforced twice: the plan feeds the upstream ZIP-317 `FeeRule` the serialized size of every output including the null-data output, and the upstream builder's value-balance assertion refuses to build any transaction whose planned fee is off by a zatoshi. Before the old path was deleted, a comparative suite proved the planner fee-for-fee and input-for-input equal to zcb 0.24's `propose_transfer`/`propose_shielding`, and the built transactions structurally equal to `create_proposed_transactions`' output (version, expiry, output sets, bundle shapes), across transfer, TEX two-step, and shield. That suite exposed two real counting rules now pinned by surviving invariant tests: the Ironwood V6 bundle counts `max(spends, outputs)` actions while the legacy Orchard bundle is branch-dependent — cross-address pairing before NU6.3, one action per spend and output after.

All 206 zingolib tests pass. The end-to-end regtest round trip (a mined transaction carrying the payload) is `op_return_data_confirms_on_chain` in the libtonode suite, pending a container run.

## Sequencing

Based on `feat/ironwood`; draft until #2419 lands. The branch carries the landed #2464 typed-error regime, which the pipeline's per-layer error enums (`PlanError`, `BuildError`, `OpReturnDataError`) extend.


zingolib/src/lightclient/propose.rs
```diff
@@ -102,11 +109,10 @@ impl LightClient {
         let request = transaction_request_from_receivers(receivers)
             .map_err(ProposeSendError::TransactionRequestFailed)?;
         let mut wallet = self.wallet().write().await;
-        let proposal = wallet.create_send_proposal(request, account_id)?;
-        wallet.store_proposal(ZingoProposal::Send {
-            proposal: proposal.clone(),
-            sending_account: account_id,
-        });
+        // A send-all is a self-spend at heart: no vault reads its origin,
```
Oscar-Pepper: i understand self-spend to mean a send-to-self, which a send-all is not. can we remove/correct this comment?

zingolib/src/wallet/spend/build.rs
```diff
@@ -0,0 +1,922 @@
...
+/// One sapling spend, self-contained.
+pub struct SaplingSpendMaterial {
+    scope: zip32::Scope,
+    note: sapling_crypto::Note,
+    merkle_path: sapling_crypto::MerklePath,
+}
+
+/// One orchard-family (Orchard or Ironwood) spend, self-contained.
+pub struct OrchardSpendMaterial {
+    note: orchard::Note,
```
Oscar-Pepper, line 112: i notice sapling has a `scope` but orchard doesnt? i will revisit this with more context

zingolib/src/wallet/spend/build.rs
```diff
@@ -0,0 +1,922 @@
...
+            }
+            PoolType::Shielded(ShieldedPool::Orchard) => {
+                let fvk = orchard_fvk
+                    .cloned()
+                    .ok_or(BuildError::MissingKey("orchard"))?;
+                let change_address = fvk.address_at(0u32, orchard::keys::Scope::Internal);
+                if ironwood_active {
+                    // Post-NU6.3 the Orchard bundle forbids ordinary
+                    // outputs; change returns to a spent note's own
+                    // address via the dedicated change-output API.
```
Oscar-Pepper: can you clarify this comment and code please? it is mentioning limitation's to orchard bundles, but my understanding is that - post-ironwood - the orchard internal key will derive addresses where fund will be directed to the ironwood internal pool. they will be in the ironwood bundle *not* the orchard bundle if i am correct, and therefore the limitations will not apply.

Oscar-Pepper: oh i need to read `add_orchard_change_output` for context. i see now that the change is in the `orchard` scope not `ironwood`, even if ironwood is active

Oscar-Pepper (commented):
not tested and i need to go through the logic in zingolib/src/wallet/spend/plan.rs thoroughly but on first-pass this PR looks really good. i will also look into more detail regarding some of my comments as I read through. im super excited to not be depending on another backend!

zingolib/src/wallet/spend/build.rs
juanky201271: Sapling's builder takes an FVK and the internal one is a separate derivation, so the scope picks between them. One Orchard FVK spends either scope, so there's nothing to carry. Now said in both structs' doc comments.

zingolib/src/lightclient/propose.rs
juanky201271: You are right — removed in 2fd574a. A send-all is not a send-to-self. The `false` argument is just `route_via_ephemeral`, which the signature already says.

zingolib/src/wallet/spend/build.rs
juanky201271: Not an oversight, and now documented on both structs in 2fd574a: Sapling's builder takes a full viewing key and the internal one is a separate derivation (`fvk()` vs `to_internal_fvk()`), so the scope picks between them. One Orchard FVK spends notes of either scope, so there is nothing for the scope to select.

zingolib/src/wallet/spend/build.rs
juanky201271: Comment rewritten in 2fd574a. It now says up front that this arm is Orchard-pool change landing in the legacy Orchard bundle (the arm below carries Ironwood-pool change), and names the real constraint: post-NU6.3 that bundle disables cross-address transfers, and `add_orchard_change_output` is the only way to retain value in one because it pairs the output with a fabricated zero-valued spend at the same address inside a single action.

juanky201271 (approved):
good enough for me.

dorianvp:
This PR moves a huge chunk of zcb to zingolib, but does not port the same tests. Closing in favor of #2753.

PR #2470, opened 2026-07-16
feat: transmit and fetch price over the Nym mixnet (ADR 0011)

zancas:
# How to test by hand

 Build the wallet with the mixnet:

`cargo build -p zingo-cli --features nym`

Place the proxy beside it:

`makers bundle-nym-proxy`

Start a connected session:

`./target/debug/zingo-cli --data-dir <your-wallet-dir>`

The mixnet turns on by itself; run:

`nym status`

until it reports ready.

Prove the tunnel without spending, run:

`updatecurrentprice`.

Then send a small amount to your own address with `quicksend`; the transaction travels the mixnet to one randomly chosen indexer, and only the fee leaves your wallet.

#### What this is:

Implements ADR 0011: the wallet transmits transactions and fetches price over the Nym mixnet by default, hiding the client IP from the indexers it broadcasts to. The design was grilled to ratification in `.agent-plans/nym-transmission.md`; this PR carries the complete implementation arc (increments 1–12).

## What the wallet now does

A connected session forces Mixnet Mode on at startup (`--no-mixnet` opts out per session; the off state is never persisted, and `--offline` sessions skip the mixnet entirely). Mixnet Mode is tri-state — off, bootstrapping, ready — and fail-closed: while the mode is on, a transport failure refuses the send rather than silently dropping to clearnet. Clearnet is reachable only by the user's deliberate per-session toggle-off.

Sends practice **Witness Rotation**: each Transmission goes to one Broadcast Indexer picked at random from a curated, operator-diverse list kept separate from the sync-server list, so no single operator accumulates a record of the user's sends. When delivery cannot be confirmed, the send escalates through serially gated rounds of fresh random picks (1, then 2, then 3 arms; at most six distinct witnesses), which routes around a censoring indexer while keeping the happy path single-witness. Every arm reuses the same regression-tested `resilient_transmit` policy as the clearnet path. The price fetch rides the same mixnet through a SOCKS5-aware reqwest client, resolving DNS at the proxy.

## Architecture

nym-sdk cannot share a lockfile with the ironwood zcash stack (its post-quantum path needs stable `crypto-common ^0.2` while `zcash_primitives` pins `=0.2.0-rc.1`; the analysis in the plan file shows the pin is durable). zingo-netutils therefore became its own workspace and resolution unit: the heavy mixnet client lives there, shipped as a spawnable `nym-proxy` binary that announces a local SOCKS5 endpoint, while the wallet side stays in the main lock and dials that endpoint through a light `socks5-transmit` feature. A shared pure racing planner (`arm_race.rs`) drives both the hedged bootstrap across providers and the send fan-out, with every per-attempt outcome retained and surfaced (`nym status` narrates bootstrap progress). The whole surface sits behind off-by-default `nym` features; the default build is byte-for-byte behavior-preserving.

The CLI gains `nym on|off|status`, startup flags `--no-mixnet` / `--nym-proxy <path>`, and zero-config discovery of a `nym-proxy` binary sitting beside the wallet executable; `makers bundle-nym-proxy` builds and places it. The Broadcast Indexer list was populated from a 130-candidate discovery sweep probed live on 2026-07-21, which found fourteen distinct operators; the list carries the eleven reachable over the mixnet, one endpoint each (the other three were excluded because they answer only on port 9067, which the mixnet exit gateways cannot reliably relay). Provenance and the sybil caveat are recorded in the module docs.

## Verification

Two new CI jobs cover what `--workspace` cannot: `netutils-standalone` (fmt, clippy `-D warnings`, tests, all features, own lockfile) and `nym-feature` (clippy + the nym unit tests in the main lock). The pure planner, fan-out, supervisor, route, and transmit policies are unit-tested against injected transports, seeded RNGs, and paused-time clocks — including the ratified falsifiers (round gating, escalation, caps, fail-closed, duplicate-equals-success). Stage 1 of the live smoke ladder (netutils' ignored tests over the real mixnet) passed after the per-attempt-timeout fix; stages 2 and 3 (bundled binary, full wallet session) remain before merge.

## Follow-ups tracked separately

Operational vetting of the broadcast list, fan-out hardening (per-arm retry tuning, independent delivery confirmation), and distribution UX (`--waitmixnet`, release packaging) are filed as their own issues.


zancas:
**Update: increments 13–17 (post-description).** The original PR built the Nym mixnet send arc through increment 12. Five further increments landed while validating it against a live mainnet mixnet, driven by what the live runs exposed. All are behind the off-by-default `nym` feature; the default build and the clearnet send path remain byte-for-byte unchanged, and the netutils split-lockfile boundary is untouched. The arc is now proven end to end: a real 10,000-zat transaction went out over the mixnet and returned its txid.

**Increment 13 — transmit heartbeat (`dfdd90fb8`).** A mixnet send can legitimately run for minutes, and the CLI previously sat silent for the whole duration. Every transmitting command (`confirm`, `quicksend`, `quickshield`, `transmit`, `migrate`, `migration start`/`catchup`) now prints a progress line roughly every thirty seconds, fed by a `TransmitProgressHandle` side channel that mirrors the existing `DrainProgressHandle` pattern. The narration is ungated, so clearnet retries benefit too; a send that finishes before the first tick stays silent. Review attention: the handle is cleared through a drop scope on every exit path, so no stale line can outlive a transmission.

**Increment 14 — instrumentation, paired probes, persistent history (`82d0661af`).** The first live send failed with six identical opaque "transport error" arms. SOCKS5 failures are now typed by connection phase — proxy-dial, tunnel establishment, post-tunnel transport, rejection — each carrying its elapsed time and the full error chain tonic otherwise hides. A new `nym probe [uri]` runs `GetLightdInfo` against every Broadcast Indexer over both clearnet and the mixnet, side by side, so a mixnet-specific failure reads at a glance; its clearnet leg contacts indexers from the real IP, so it is a user-invoked diagnostic only, never an automatic path. Every send arm and probe leg appends to an append-only, dependency-free TSV beside the wallet, readable via `nym history`, recording only host, route, timing, and error text — never wallet or address data.

**Increment 15 — connect timeout and port-443 prune (`1383922d6`).** The live probe exposed a bug: a witness that completed the SOCKS5 tunnel but stalled the TLS handshake hung for over five minutes, because the per-phase socket timeouts did not cover the handshake tonic layers on top. A `connect_timeout` on the endpoint bounds it, covering both the send and probe paths through the shared dialer. The probe also showed a clean split — every port-9067 witness fails over the mixnet while every port-443 witness succeeds, because the exit gateways relay the standard port but mishandle the non-standard one — so the three port-9067 entries left the broadcast list, and a regression test now requires every entry to be port 443.

**Increment 16 — proxy lifetime coupling (`9960fa823`, refined in `9db772ab3`).** A user's Ctrl-C killed the proxy child out from under the session, and because the supervisor stopped watching stdout at ready, Mixnet Mode stayed stale-ready against a dead proxy. The fix couples the child's lifetime to the parent rather than insulating it from signals: the proxy spawns in its own process group so a terminal interrupt no longer reaches it, and the supervisor holds the child's stdin pipe open so any parent exit — clean, panic, or SIGKILL — closes the pipe and the child reads EOF and exits. A fourth Mixnet Mode state, `Died`, records an unexpected proxy exit; it is distinct from `Off` because it is unconsented, so the route resolver refuses on `Died` rather than falling back to clearnet. Review attention: this is the fail-closed invariant — please confirm `resolve_route` never yields clearnet for any state but the deliberate `Off`, including `Died` with a stale address on hand.

**Increment 17 — health-gated readiness and https-only (`7be0f1b02`).** Two separate live sessions drew mixnet paths where the SOCKS5 listener came up but carried no data end to end; the proxy announced readiness anyway, so the wallet marked itself ready against a dead path. The existing connectivity check could not catch this, because it only opened a tunnel and never moved data. The proxy binary now health-gates readiness with a real `GetLightdInfo` round trip through its own tunnel before announcing, redrawing fresh gateways on failure up to three distinct paths, and exiting non-zero if all fail so the supervisor records `Died` and sends refuse fail-closed. Separately, mixnet transmission is now https-only: the tunnel terminates at an untrusted exit gateway, so the dialer refuses any non-https indexer with a typed error rather than falling back to plaintext, and the broadcast-list tests and `nym probe` enforce it too.

**Testing.** Each increment carries falsifying unit tests — narration order and scope-clearing, phase attribution against a dead port and a handshake-breaking fake proxy, the port-443 and https invariants, the `Died`-refuses-clearnet invariant, the post-ready death transition, and the paused-clock heartbeat cadence. The redraw/retry loop and the real mixnet round trips are exercised by the live smoke ladder, which passed all three stages, culminating in the confirmed mainnet send.

PR #2478, opened 2026-07-17
Six edges tested: verification and fixes for the #2419 review findings

zancas:
An external review of #2419 raised six potential correctness violations. This branch subjects each one to an independent verification: for every finding, a separate investigation re-derived the claimed mechanism from the source, the git history, and the pinned zcash_client_backend 0.24.0-rc.1, and then authored a test that demonstrates the defect. Every confirmed finding's test was observed failing against the unfixed tree with exactly the predicted failure before the fix was written, so each test is a true detector rather than an assertion of the fix's own behavior.

The branch carries one commit per finding, each combining the demonstrating test with its fix. Two findings needed no fix and carry pinning tests instead: the ironwood OutputRef label is correct as written (the contradicting comment was a relic of the patched 0.23 era and is replaced by the true invariant), and the open-window relaunch concern is refuted because `broadcast_due_parts` selects the currently open bucket from persisted state. The finding 3 fix routes the drain flow and the note-splitting rounds through one shared `build_and_transmit` engine, so both enforce the same cleanup contract. Test scaffolds that repeated across modules moved into reusable fixtures, notably `testutils::synthetic_wallet::inject_confirmed_orchard_notes` for already-synced mock-net wallets with really derived nullifiers. The full pepper-sync and zingolib unit suites pass (67 and 205 tests), and clippy and rustfmt are clean. The 6a demonstration is 32-bit only and passes vacuously on x86_64; running it under a 32-bit target such as i686-unknown-linux-gnu shows the failure the fix removes. The finding 3 test costs about fifty seconds, dominated by the wallet's fixed transmit-probe stalls.

| Potential violation | Test | Verdict | Commit | Fix |
| --- | --- | --- | --- | --- |
| 1. `TargetValue::AllFunds` aborts the process in `select_spendable_notes` | `select_spendable_notes_all_funds_returns_err_not_panic` | Confirmed | 569215aa8 | The arm returns the new `WalletError::AllFundsSelectionUnsupported` through the trait's error channel. |
| 2. Self-contradicting pool label on selected ironwood inputs | `ironwood_noteref_pool_label_round_trips_through_exclude` | Contradiction confirmed; the code is right, the comment was stale | 88f2b1e19 | The comment now states the real invariant: the label must match the `ReceivedNotes` vector because it round-trips through the `exclude` split. |
| 3. A failed note-split round strands `Calculated` transactions with notes marked spent | `failed_split_round_transmit_strands_calculated_transactions` | Confirmed (the paused-sync leg is unreachable today: `pause_sync` never engages under `migrate_to_ironwood`'s exclusive borrow) | 3fc286cae | Drain and split rounds share `build_and_transmit`, which fails already-built transactions on a build error and all unsent ones on a transmit error. |
| 4. Chain-tip prioritization floods the wallet range while ironwood shard ranges are empty | `chain_tip_priority_confined_to_tip_when_ironwood_shards_are_empty` | Confirmed | ba59f3b6f | The empty-shard fallback start is clamped to the pool's activation height (Sapling, Nu5, or Nu6.3). |
| 5. Open-window parts unreachable after a relaunch | `open_window_part_is_broadcast_at_relaunch` | Refuted: `broadcast_due_parts` selects `bucket_index == current_bucket` from persisted state, driven by `auto_broadcast_if_due` | 3f8cc5e9b | None needed; the test pins all three legs of the mechanism. |
| 6a. `u64` to `usize` truncation reading `max_actions_per_split_tx` | `max_actions_read_never_silently_truncates` | Confirmed, 32-bit targets only; no 64-bit misbehavior exists | c8a46a210 | `usize::try_from` with an `InvalidData` error, matching the reader's other malformed-field paths. |
| 6b. Ironwood tree-size check cannot distinguish a non-serving server from corruption | `nonzero_ironwood_tree_size_mismatch_is_rejected` | Confirmed | a8d1dc0a6 | A nonzero-versus-nonzero mismatch returns `ScanError::IncorrectTreeSize`; the warning remains for the genuine zero-sided cases. |
| 6c. Restore-from-backup's unknown txid treated as failed, re-planning an in-flight split | `unknown_txid_after_restore_is_not_classified_failed` | Confirmed | 6b0b0203c | `transaction_failed` uses `is_some_and`, so an unknown txid awaits confirmation instead of triggering `RetrySplit`. |

### DIFF
--- a/.agent-plans/pr2419-findings-verification.md
+++ b/.agent-plans/pr2419-findings-verification.md
@@ -1,0 +1,41 @@
+# Claim: independent verification of the #2419 review findings (2026-07-16)
+
+Session working in the `feat_ironwood` worktree on branch `feat_ironwood`.
+The user handed over six correctness findings from a reviewer of PR #2419
+and asked for an independent review plus an attempted failing test that
+demonstrates each finding. The deliverable is test code and a verdict per
+finding, not fixes. Any product-source change waits for the user's
+ratification.
+
+## Scope
+
+The work began as tests only; the user then ratified fixing each confirmed
+finding, with test and fix combined per commit on the `six_edges_tested`
+branch (PR against feat/ironwood, Oscar reviewing). Verdicts: findings 1,
+3, 4, 6a, 6b, and 6c confirmed and fixed; finding 2 resolved in the code's
+favor (the comment was stale) and pinned by test; finding 5 refuted and
+pinned by test. Fixtures added while DRYing the tests:
+`testutils::synthetic_wallet::inject_confirmed_orchard_notes`, the
+migrate.rs test-module scaffold helpers, and `planned_state` in the store
+tests.
+
+## File claims (test-only additions)
+
+- `zingolib/src/wallet/zcb_traits.rs` (findings 1 and 2: AllFunds panic,
+  ironwood OutputRef pool label)
+- `zingolib/src/lightclient/migrate.rs` (finding 3: paused sync leak on the
+  note-split round)
+- `pepper-sync/src/sync/state.rs` (finding 4: chain-tip range fallback with
+  empty ironwood shard ranges)
+- `zingolib/src/wallet/migration/schedule.rs` and
+  `zingolib/src/wallet/migration/reconcile.rs` (finding 5: open-window part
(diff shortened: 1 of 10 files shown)


zancas:
Superceded by #2481

PR #2479, opened 2026-07-17
Fix faucet normalization under pool-preference selection; drop two stale pins

zancas:
## Summary

Two pin cleanups from the 2026-07-17 dependency audit, and the root-cause fix
for the three mining-balance tests that failed on freshly mined chains.

## The normalization fix (the important part)

`mine_to_ironwood`, `send_orchard_back_and_forth`, and
`send_mined_ironwood_to_ironwood` failed on freshly mined chains with
balances exactly one block reward (618,780,000 zatoshis) short, while cached
replays passed. The failures were never zebrad's: fresh chains fail
identically under 6.0.0 and 6.0.0-rc.0, and A/B experiments show both
validators produce value-identical coinbases and accept the migration
transaction. The cause is wallet commit 731b2b761 ("repair note selection's
stale remainder and honor pool preference", merged with #2468 after the
chain caches were mined): once selection leads with the payment's own pool,
`normalize_shielded_faucet_balance`'s self-send to an ironwood receiver
stops draining orchard and leaves one block reward un-migrated, which the
ironwood-balance assertions misread as a missing reward. Cached chains kept
passing because replay resubmits wallet-built blocks that predate the
selection change — chain caches mask changes by any author of chain bytes,
the wallet included.

The normalization now performs an orchard-restricted drain via
`drain_orchard_to_ironwood`, asserting `stranded == 0`. The drain spends
only orchard by construction and cannot touch the tip block's note, and its
fee cancels exactly as the offload's does (the faucet collects it back in
the confirming block's coinbase), so `funded_faucet_ironwood_balance` and
every downstream expectation are unchanged. The unused
`IRONWOOD_MIGRATION_AMOUNT` constant is deleted. The full suite runs green
with regenerated caches (`ZINGO_REGENERATE_CHAIN_CACHE=1`) — the
environment that exposed the bug.

## The pin cleanups

The `time` `[patch.crates-io]` entry is dropped: crates.io has carried the
exact version the patch resolved (0.3.47) since 2026-02-05, and the lockfile
re-sources `time`/`time-core`/`time-macros` at identical versions, so the
dependency graph changes source only. `zaino-proto` moves from the March
git rev to crates.io 0.2.0 (the zaino workspace's 2026-07-13 debut), with
compile gates across `zingolib` (default and `testutils`), `zingo-cli`,
`zingolib_testutils`, and `libtonode-tests` bounding rev-to-release drift at
zero for our usage surface. One duplicate remains by necessity:
`zingo_grpc_proxy` pins `zaino-proto` through its own git manifest, so the
lockfile carries both copies; the gates prove the types never cross, and the
duplicate resolves when `zingo_grpc_proxy` re-pins or publishes.

Held pins, re-verified during the audit: `lightwallet-protocol` (upstream
has released nothing containing the ironwood protos; the pinned rev is
their HEAD), `zcash_client_backend` (no 0.24.0 final exists),
`bip32` (0.6.0-pre.1 is what the librustzcash train requires), the
infrastructure pin (no tag covers the pinned content), and the zainod image
(zingolabs/zaino#1404 remains open; the 0.6.0 tags contain the regression).

### DIFF
--- a/.agent-plans/zebrad-6-0-0-bump.md
+++ b/.agent-plans/zebrad-6-0-0-bump.md
@@ -79,6 +79,19 @@ test-logs/observatory/concrete__*.log. Root-cause analysis and a
 patched/rc dependency audit are delegated to background agents this
 session.
 
+## Resolution (2026-07-17): PR #2479
+
+Root cause of the three mining-balance reds: wallet commit 731b2b761
+(pool-preference selection, merged with #2468 after the caches were
+mined) defeated the normalization self-send; zebrad exonerated by a
+fresh-chain rc.0 A/B. Fixed by draining orchard via
+drain_orchard_to_ironwood in normalize_shielded_faucet_balance
+(d7196e8a4); user-confirmed green with regenerated caches. Shipped as
+PR #2479 (chore/drop-time-patch → feat/ironwood) together with the
+audit's two pin cleanups: the time [patch.crates-io] drop (a46f5e632)
+and the zaino-proto git→crates.io 0.2.0 migration (4c36bcd2e; a
+benign duplicate copy remains via zingo_grpc_proxy's own git pin).
+
 ## Merge state (2026-07-17)
 
 PR #2475 (with #2477's two commits inside) is MERGED into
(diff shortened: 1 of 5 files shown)


zancas:
Superceded by #2481

PR #2480, opened 2026-07-17
fix: AllFunds selection as a pure function (was a process-aborting panic)

zancas:
## Summary

This PR closes the last verified finding from the #2419 review: the
`TargetValue::AllFunds` arm of `select_spendable_notes` was a `panic!`, so a
request any upstream `zcash_client_backend` sweep proposal can legitimately
make would abort the whole process. It lands as a TDD pair — the contract
tests are committed red first, failing on that exact panic, and the fix
turns them green.

## The contract

`MaxSpendable` selects every spend-worthy note: all pools participate
regardless of source order (matching budgeted selection's unconditional
fallback), migration-reserved notes are included because all funds means
all, and the dust discipline is the house rule — a note at or below
`MARGINAL_FEE` costs more to spend than it contributes, exactly as budgeted
selection and the migration drain already decide.

`Everything` is refused with a new typed error,
`WalletError::AllFundsEverythingUnsupported`, rather than half-implemented:
its upstream contract — fail if *any* unspendable funds exist — requires a
whole-wallet audit this selector does not yet perform, and a wrong success
would silently strand funds. The typed refusal keeps the process alive and
fails the proposal cleanly; the doc comment carries the reasoning for
whoever implements the audit.

## The shape

Functional core, effects at the edge: the match arm only gathers the strict
(guaranteed-unspent) spendable candidates per pool through the existing
read-only accessor, and the decision is `all_funds_selection`, a
module-level pure function — no wallet access, no mutation, same inputs to
same selection — so the policy is testable without a wallet. The tests
drive the real trait method end to end through `SyntheticWalletBuilder`:
one spendable note per shielded pool plus a dust-line orchard note,
asserting the three-note selection totals 220,000 zatoshis with the dust
left behind, and that `Everything` yields the dedicated error.

The only file-overlap risk is #2478's touch to the same file in the
`OutputRef` label region; the hunks are disjoint.

### DIFF
--- a/.agent-plans/allfunds-pure-selection.md
+++ b/.agent-plans/allfunds-pure-selection.md
@@ -1,0 +1,23 @@
+# Claim: AllFunds selection — TDD fix (grilling session, 2026-07-17)
+
+The #2419 review verified that `select_spendable_notes` panics on
+`TargetValue::AllFunds` (zingolib/src/wallet/zcb_traits.rs), the sole
+verified finding without an owner. The user ratified a TDD pair on
+branch `fix/allfunds-pure-selection` (based on feat/ironwood,
+independent of PR #2479): first a red test pinning the contract, then
+the fix implemented as a side-effect-free pure function
+(functional-core/imperative-shell — the wallet reads gather
+candidates at the edge; a pure `all_funds_selection` decides).
+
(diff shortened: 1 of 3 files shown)


zancas:
Superceded by #2481

PR #2486, opened 2026-07-18
fix: restore migration state on every broadcast-path exit

zancas:
The due-part broadcast path takes the migration state out of the wallet so it can mutate the state and the wallet through independent `&mut` borrows, and an early `?`-return on `WalletError::NoSyncData` fired before the restore, silently destroying the consented ZIP 318 migration schedule; the next save persisted the loss. The state is production-reachable: `LightWallet::clear_all` empties the scan ranges — the sole source of `last_known_chain_height` — while keeping the migration, so a rescan followed by any broadcast driver (`auto_broadcast_if_due`, `broadcast_due_parts`, `catch_up_migration`, or `migrate_to_ironwood`) lost the schedule. This is finding B1 of the #2419 review.

The fix makes the leak structurally impossible rather than patching the one site. A new `LightWallet::with_migration_state` combinator owns the take/use/restore bracket: the body is a function argument, the state is restored on every exit path, and because the body is synchronous the state can never be out of the wallet across an `.await` point — which also closes a second instance of the same class, discovered during the conversion: broadcast Phase C held the state outside the wallet across its per-part network submits, so a cancelled future (a timeout wrapper, a `select!`) would strand it with no error at all. Phase C is restructured as a bracketed recording step followed by a submit loop whose awaits all run with the state inside the wallet. `reconcile_migration`, the catch-up shift block, and `refresh_part_witnesses` convert to the same combinator; the take-and-drop in `cancel_ironwood_migration` is intentional destruction and stays.

The regression test is a DRY pair inside `mod no_sync_data_preserves_migration_state`: one shared scenario function parameterized only by how the height becomes `None`, with `via_clear_all` pinning that the production rescan route really produces a live migration with no height (its guards fail loudly if `clear_all` ever changes semantics), and `via_empty_sync_state` pinning the contract on the state itself, independent of any route into it. One test red with the other green names the layer that changed.

Verification: `cargo nextest run -p zingolib` passes (209 tests) on the host, clippy is silent, and fmt is applied. The chain-bound libtonode suites are container-only in this tree and were not run here.

### DIFF
--- a/.env.testing-artifacts
+++ b/.env.testing-artifacts
@@ -3,9 +3,17 @@
 
 # Zaino image tag used to provide zainod (https://hub.docker.com/r/zingodevops/zainod)
 ZAINO_IMAGE_TAG=0.6.0-rc.1-no-tls
+# Manifest digest the tag resolved to when pinned (2026-07-17). Tags are
+# mutable pointers; the digest is content. docker-ci consumes
+# tag@digest, so an upstream repush of the tag fails the build loudly
+# instead of silently changing which zainod the tests run. Note the
+# binary inside this image self-identifies as 0.4.3-ironwood.1.
+ZAINO_IMAGE_DIGEST=sha256:e48b133dbf53dbed77b872de74d65aa8a45357a3844b658ea472a340949feb7f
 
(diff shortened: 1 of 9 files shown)


dorianvp (approved):
utACK!

PR #2489, opened 2026-07-20
chore: expand API for mobile

dorianvp:


zingolib/src/lightclient/migrate.rs
```diff
@@ -668,17 +668,44 @@ impl LightClient {
     /// This function is idempotent over wallet state: a call that fails partway leaves the notes
     /// of every unsent transaction spendable. Calling it again re-plans and
     /// sends the remainder.
+    ///
+    /// Syncs the wallet before draining. Consumers that own the sync
+    /// lifecycle and keep a background sync running should call
```
zancas, line 673: Can we enforce this with a side-effectless fn that acts as a contract binding the correcting behavior in compiled code for each of the cases?

dorianvp: About it being synced the answer is probably no at compile time. About the bracketing, probably yes but that could be it's own unit or PR instead of here... are you thinking about introducing a guard or similar?

zancas (approved):
Follow-up with improvements, in-flight.

PR #2490, opened 2026-07-20
fix: sync quiescence for the drain and the stored proposal, behind shipped APIs

zancas:
## What this does

The review of #2489 found that `drain_orchard_to_ironwood_presynced` planned against the wallet while the caller's background sync could still be mutating it — the pause happened only later, inside the build stage. This PR fixes that race and, while pinning it with tests, fixes three latent defects of the same pause discipline in the shipped propose/send family — **without changing any shipped public signature**.

## The shape of the fix

**`SyncQuiescence`** is an RAII witness: `quiesce_sync()` pauses a running engine (a paused or not-running engine is already quiescent), and dropping the witness restores the prior sync mode. A `stop_sync` issued while the witness is held wins over the restore.

The witness appears in a public signature in exactly one place: `drain_orchard_to_ironwood_presynced` takes `&SyncQuiescence`, so the plan-under-running-sync call shape does not compile (a `compile_fail` doctest pins this). That entry point is new in #2489 with no released consumer, so nothing shipped breaks.

The shipped propose/send family keeps its API — `resume_sync: bool` and all — and holds a witness *internally*, stored beside the stored proposal (process-lifetime state, ADR 0006): minted before a proposing call's first wallet read, released when the proposal is consumed (send or calculate), cleared (`clear_proposal`), or fails to come into existence.

## Red-first

The first commit lands four `quiescence_contract` tests that fail against `feat/ironwood`, one per defect:

1. `propose_send` pauses, then error-returns past every resume — a failed proposal leaves a consumer's background sync silently suspended.
2. `propose_shield` never paused: a shield proposal is created under a running engine.
3. `quick_shield` never paused: it builds and stores signed transactions under a running engine.
4. `propose_send_all` runs its send-all sizing (the `max_send_value` trial proposals) before its pause.

They stay red through the witness and drain commits and turn green at the implementation commit, behind unchanged signatures. Three further tests pin the corrected protocol: a stored proposal holds its pause; `clear_proposal` (the decline path) restores the engine; an Indexerless send failure, which consumes nothing, releases nothing.

## Behavioral notes for consumers

- No call-site changes anywhere: zingo-cli, the test suites, and mobile's shipped integration compile and behave as before on the happy path.
- One deliberate refinement: `send_stored_proposal(resume_sync: true)` now restores the engine's *prior* mode instead of unconditionally resuming, so a pause the caller established before proposing is preserved rather than overridden.
- `clear_proposal` now ends the proposal's pause. Previously a declined proposal left sync suspended until some later send opted into resuming.

## Deferred

Converting the shipped propose/send span to a caller-held witness (the API-breaking form) is deferred to a coordinated move with mobile.

Verified: 224-test zingolib suite green, drain `compile_fail` doctest green, the four contract tests verified red at each pre-implementation commit, workspace check/clippy/fmt clean. Chain-bound libtonode call sites are compile-verified (container-only on this host).

### DIFF
--- a/zingolib/CONTEXT.md
+++ b/zingolib/CONTEXT.md
@@ -115,6 +115,8 @@
 
 **SyncMode** — The current state of the sync engine: `NotRunning`, `Running`, `Paused`, or `Shutdown`.
 
+**SyncQuiescence** — A held value that witnesses the sync engine is quiescent — paused or not running — and keeps it so for as long as the value lives. `LightClient::quiesce_sync` is the sole constructor: it pauses a running engine, and dropping the witness restores the prior sync mode. The Orchard→Ironwood drain demands the witness as a parameter, so planning under a running sync is unrepresentable for that path; the Two-phase Send holds one internally beside the stored Proposal, giving the shipped propose/send protocol the same guarantee without a signature change.
+
 **SyncConfig** — Configuration for the sync engine: performance level and transparent address discovery settings.
 
 **Performance Level** — Controls batch size and nullifier map scope: `Low`, `Medium`, `High`, or `Maximum`.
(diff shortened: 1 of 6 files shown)


dorianvp (approved):
I will approve but will change the name in a follow-up PR. The name is technically correct but it's working against readability.

Also, here's a quick agent pass on it: 

> The type is an RAII guard — it pauses on construction and resumes on drop. But it's named after an abstract state ("quiescence") rather than its job (a scoped pause handle). The doc comment itself reaches for "witness" and "evidence" to explain what it is, which is a sign the name isn't carrying its weight. Readers either already know the word or stop to look it up, and it buys nothing over plainer options.

Issue #2493, opened 2026-07-21
Ironwood hardening backlog: open findings from the #2419 review


zancas:
A six-slice review of PR #2419 (feat/ironwood, reviewed at head f1b81dd6c) produced the findings below. Two findings are being remediated directly on the branch and are excluded here: the bit-rotted `unit_test_twins` balance macros (the two live originals are deleted; their offline twins in `mock_chain_tests.rs` were correct and remain the sole record), and the pre-ironwood wallet-file wipe on first reorg truncation (fixed by a pure truncation contract in `pepper-sync/src/sync/truncate.rs` with a red-green test). Everything else is pinned here for follow-up. File and line references are against the review head.

## High severity

1. **The scheduled migration flow strands unsplit wallets.** No production code constructs `MigrationPhase::NoteSplitting` (it appears only in `store.rs` serialization and in tests), so `start_ironwood_migration` on a fragmented wallet persists `Planned` forever. Reconcile keeps recommending `ContinueNoteSplitting`, which nothing implements, and the stuck state blocks the drain path with `AlreadyInProgress`.

2. **One bad part aborts an entire broadcast pass.** `prepare_part` returns hard `Err` for per-part recoverable conditions — bound note already spent, nullifier mismatch, pre-activation boundary (`zingolib/src/wallet/migration/parts.rs:453-508`) — and Phase A propagates it with `?` (`zingolib/src/lightclient/migrate.rs:413`), so every other due part in the window goes unsent until a reconcile happens to run. These conditions belong in `PrepareResult::Skip`.

3. **`migrate_to_ironwood` bypasses recorded consent and collapses the schedule.** When a scheduled migration exists, the one-call path reuses it without re-checking the plan hash or `state.account` (`zingolib/src/lightclient/migrate.rs:997-1041`) and force-shifts every pending part into the current bucket — the correlated immediate broadcast the recorded consent rejected.

4. **Re-running `migrate` after completion burns fees and migrates nothing.** Only `cancel` clears a `Complete` state, so a rerun with newly received orchard funds executes split rounds (paying fees), skips binding because stale confirmed parts make `state.parts` non-empty, broadcasts nothing, overwrites `Complete` with `PartsScheduled`, and returns a success summary with empty `part_txids` (`zingolib/src/lightclient/migrate.rs:987-1075`).

[...]

PR #2495, opened 2026-07-21
fix: stop migrated-wallet wipes and migration consent bypasses (#2419 review highs)

zancas:
The #2419 review's high-severity findings, fixed at the root — and where a finding exposed a bug class, hardened until the class itself is gone. The remaining medium and low findings stay pinned in #2493.

**Deleted the bit-rotted twin ledgers.** The `unit_test_twins` bump-and-check macros bound an `i:` argument but expanded to `i: 0`, silently zeroing every ironwood expectation their call sites recorded — invisible in default CI because the suite sits behind an off-by-default feature. The offline twins in `mock_chain_tests.rs` were correct all along and are now the sole record; the equivalence doc says so with a dated amendment.

**Stopped reorg truncation from wiping migrated wallets.** A pre-ironwood (v0) `ShardTrees` blob deserializes to an empty ironwood tree, and the first routine reorg after the upgrade read its missing checkpoint as corruption and cleared all wallet data. Truncation decisions now live in one pure function, `sync::truncate::plan_truncation`, which maps wallet state, per-pool checkpoint facts, and the truncation target to the correct post-truncation state of every store. Red-green: the regression test reproduced the wipe before the fix and pins it after. The root cause — an `Ok(bool)` conflating "nothing to roll back" with "cannot roll back" — is then retired as a class: a crate-boundary wrapper (`shardtree_ext`) converts every shardtree boolean into a closed domain enum with a source-walk test forbidding raw calls, plan variants carry the evidence that justifies them, and the pure planner is pinned by an exhaustive sweep plus proptest properties over generated checkpoint histories (dev-dependency only, no new crate in the lockfile).

**Gated the migration lifecycle on mode and consent.** `MigrationState` now records which flow created it (`Scheduled`/`Immediate`, persisted as migration-section v2 with v1 blobs conservatively reading as `Scheduled`). The one-call `migrate_to_ironwood` refuses to collapse a consented schedule into a correlated immediate broadcast, refuses another account's migration, clears a completed migration so a rerun binds fresh funds instead of burning split fees and returning an empty success, and still resumes its own interrupted runs. `start_ironwood_migration` refuses an unsplit plan with a typed error instead of stranding consent in a `Planned` state no driver advances while blocking the drain.

**Stopped one bad part from aborting the broadcast pass.** A spent bound note, a nullifier divergence, or a pre-activation boundary now skip that part (with evidence-carrying `SkipReason` variants, falling to reconciliation) instead of hard-erroring out of the whole pass and leaving every other due part unsent.

**Pinned pre-ironwood semantics for the mobileclient scenarios.** The externally consumed `*_mobileclient` scenarios changed behavior transitively when the workspace default flipped to ironwood-era. They now use their own activation schedule with NU6.3 deferred, restoring orchard-pool funding on a never-activating chain, with a unit test forbidding the flip from leaking back.

Verification: pepper-sync 85/85 and zingolib 231/231 fast tests, workspace clippy, fmt, and `cargo check --workspace --tests` all green.


dorianvp:
Here's a probably-too-verbose Fable review:

**1. The twins deletion rests on a false claim — the offline twins do *not* "carry the same ledgers."** The bit-rot itself is real: I verified in the pre-image that both `bump_and_check!` macros bind `$i:tt` but expand `i: 0` (old `unit_test_twins.rs:651-653, 702-705`), so the deletion rationale's premise holds. But the justification for deletion — "the offline twins... carry the same ledgers, so they are now the sole record" — is contradicted by the values. For `from_t_z_o_tz_to_zo_tzo_to_orchard`: the twin in `mock_chain_tests.rs` has **17** bump calls vs the original's 16, and the ledgers diverge from step 4 onward in *every* column, not just the zeroed ironwood one — original records `i: 45_000` where the twin asserts `i: 55_000`; original `s: 5_000 t: 5_000` vs twin `s: 10_000 t: 10_000`; original `i: 485_000` vs twin `i: 500_000`, drifting through the end (`s: 340_000` vs `s: 395_000`). For `list_value_transfers_check_fees`: original recorded `i: 5_000` where the twin asserts `i: 15_000`. So either the originals' recorded values were stale (plausible — the whole file sat behind an off-by-default feature and may never have run post-migration) or the mock chain's fee/ledger model diverges from a live chain; the PR doesn't know which, and deleting the originals destroys the only artifact that would force the question. The doc's prior "EQUIVALENT (assertion-identical, full 16-step ledger)" for row 8 was already false, and the amendment perpetuates it. Note also that the fix for the actual defect was five characters (`i: 0` → `i: $i`); by this PR's own evidence-first standards, the right move was fix the macro, run the live originals once, and let the chain adjudicate the ledger — then delete if warranted. As written, request changes on this commit.

**2. The truncation planner's soundness premise is false in this codebase: subtree roots enter shard trees with no checkpoint.** The module doc (`truncate.rs`) justifies `Untouched` with "frontier and subtree insertion both attach checkpoints — so the newest checkpoint bounds the chain state the tree knows about." Frontier insertion does (`sync.rs:2146ff`, `Retention::Checkpoint`), but subtree insertion does not: `witness::add_subtree_roots` (`witness.rs:118-125`) writes shards via `store_mut().put_shard(...)` with `RetentionFlags::EPHEMERAL` and no checkpoint. Worse, `update_subtree_roots` resumes fetching from `get_shard_roots().len()`, so a once-inserted root is never re-fetched or corrected. Concrete hazard: a subtree completes near the tip, the wallet fetches its root, then a reorg within `MAX_REORG_ALLOWANCE` crosses the completing block. If the affected pool's tree has no checkpoint at the truncation target — which is *exactly* the migrated ironwood tree this PR protects (newest checkpoint = 0) — the planner says `Untouched` and the now-stale root stays in the store permanently, feeding wrong witnesses/anchors later, silently. The old code's brutal wipe was accidentally safe against this; the successful-rollback path was also safe because `truncate_to_checkpoint` prunes above the checkpoint position and the count-based fetch then re-requests. `Untouched` is the one path that knowingly retains a tree whose above-target contents the facts can't see. The window is narrow (root fetched between subtree completion and the reorg), but it's low-probability/high-impact/undetectable — the same profile as the wipe this PR fixes. At minimum the module doc's premise needs correcting; a real fix is either shard-root extent in `TreeTruncationFacts` or re-validating the newest stored root against the server each session (which retires the staleness class beyond truncation).

**3. "The boolean-blindness class can no longer be reintroduced silently" is only true inside pepper-sync — and it already survives outside it.** `zingolib/src/wallet/legacy.rs:131, 139, 197, 205` call raw `sap_tree.checkpoint(height).expect("infallible")` and discard the boolean entirely — no assert, in the legacy wallet *import* path, where a `NotAboveNewest` (duplicate or out-of-order checkpoint in an old wallet file) is far more plausible than in `ShardTrees::new`, and a silently skipped checkpoint there is precisely the condition the PR itself says "feeds RequiresRescan." The source-walk test only covers `pepper-sync/src`, and `ShardTreeExt` is `pub(crate)`, so downstream crates *cannot* route through the wrapper even if they want to. Either make the trait `pub` and extend the confinement story, or at least assert in `legacy.rs`; as is, the "class retired" framing in the PR description overstates the guarantee.

**4. The immediate-path entry gate is entry-only, and the per-round resume trusts whatever state exists.** `migrate_to_ironwood` runs the gate once under a write lock, drops it, then per round takes a *fresh* lock and does `wallet.migration.take()` followed by force-shifting every `Assigned` part into the current bucket — with no check that `state.mode == Immediate` or `state.account == account` at that point. Today this isn't exploitable because every creator (`start_ironwood_migration`, the loop itself) takes `&mut self`, so one handle serializes them — but that means the consent guarantee rests on receiver discipline at the API surface, not on the state machine. The moment a scheduled-flow split driver (explicitly planned in #2493) or a second client handle over the shared `Arc<RwLock<LightWallet>>` exists, the exact consent collapse of finding 3 reopens through the resume path. One assert inside the per-round closure closes it permanently and cheaply.

**5. `BoundaryBeforeActivation` can make the immediate loop spin to a misleading error.** If NU6.3 activation falls inside the current bucket window, every part's boundary is below activation, every broadcast pass skips everything, `await_migration_confirmations(&[])` returns immediately (vacuous `all`), and the loop burns all 64 `MAX_ROUNDS` back-to-back before failing with `SplitDidNotConverge` — wrong diagnosis for "wait for the next bucket." The old hard error was at least accurate. Relatedly, the `SkipReason` doc's claim that "Reconciliation reassigns the part once post-activation buckets open" is optimistic: `classify` has no arm for this condition, so the part only heals once its window closes into `Overdue` → `PromptCatchUp`, which requires user consent.

**6. Gate ordering nits.** `DifferentAccount` is matched before the Completed-clears arm, so a *completed* migration for account A permanently blocks account B's immediate path (nothing else clears it). And the gate destroys the `Complete { residual }` record before `plan_ironwood_migration` gets a chance to fail — a failed precondition shouldn't erase history it never used.

Things I checked adversarially that **hold up**: the planner itself is correct and its exhaustive sweep genuinely covers its input space (`checkpoint_at_target ∈ {None, Some(target)}` matches everything `gather` can produce); the `ClearAll`/`NoOp` region boundaries are behavior-identical to the old birthday-clamp logic, including the fresh-wallet edge where `highest_scanned < birthday`; the error path's `clear_wallet_data` recursion terminates through the `ClearAll` branch and does clear the trees; the v1→v2 mode-byte persistence and its append-only inverse test are sound; `Scheduled`-as-default for v1 blobs is the right conservative reading; the source-walk patterns don't false-positive on `checkpoint_classified`/`checkpoint_count`; the `Cargo.lock` "+1 line, no new crate" claim is accurate (it's the `proptest` entry in pepper-sync's dep list, already in the graph); and the mobileclient deferral is correctly rooted at `unfunded_mobileclient` with `normalize_shielded_faucet_balance` genuinely inert when `nu6_3()` is `None`. One nit there: the alignment test claims "everything below NU6.3 stays aligned" but checks only nu5–nu6.2 of the nine sub-ironwood heights, and the builder copies from the regtest fixture rather than the workspace default it's asserted against.

zancas:
All six points land, and point 1 changes the PR. Agreed: the deletion's "same ledgers" claim was false, and the ledgers diverge exactly as you showed. Both originals are restored verbatim with the five-character macro repair (d2c491071), and the equivalence record now carries rows 5 and 8 as LEDGER DISPUTED until the next extra-credit run lets the chain adjudicate which record is true. Points 2 and 3 are confirmed against source and fixed in 73c751c9e: sessions refetch the newest still-bare subtree root (`put_shard` replaces it if the chain moved, making the count-based resume self-healing), the truncation doc states the corrected premise, `ShardTreeExt` is exported, the legacy-import sites classify their outcomes, and zingolib enforces the confinement with its own source-walk. Points 4 through 6 are fixed on the follow-up branch `fix/2493-migration-lifecycle-mediums` (5a6ff69b1): the per-round resume verifies mode and account, pre-anchorable immediate rounds refuse with a retry-after height, and the gate clears completed history only after a successful plan, ordered so one account's history never blocks another. One note on point 5: the activation floor now in the schedule (same branch) confines `BoundaryBeforeActivation` to schedules persisted before the floor existed, and its doc now says so.

pepper-sync/src/shardtree_ext.rs
```diff
@@ -0,0 +1,166 @@
...
+use zcash_protocol::consensus::BlockHeight;
+
+/// Outcome of adding a checkpoint at the tree's rightmost leaf state.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CheckpointOutcome {
+    /// The checkpoint was added.
+    Added,
+    /// The id is not above the store's newest checkpoint; the tree is
+    /// unchanged and no checkpoint was added.
+    NotAboveNewest,
```
Oscar-Pepper, line 39: we should not limit checkpoint addition to only being the newest... we need to store the top MAX_REORG_ALLOWANCE checkpoints of the chain and they may be scanned in any order

Oscar-Pepper: we should always add checkpoints and allow the shard tree crate to handle pruning

pepper-sync/src/sync/truncate.rs
```diff
@@ -0,0 +1,619 @@
...
+/// evidence in its facts.
+fn plan_pool_truncation(
+    facts: TreeTruncationFacts,
+    truncate_height: BlockHeight,
+) -> PoolTruncation {
+    if let Some(checkpoint) = facts.checkpoint_at_target {
+        return PoolTruncation::ToCheckpoint { checkpoint };
+    }
+    match facts.newest_checkpoint {
+        Some(newest_checkpoint) if newest_checkpoint > truncate_height => {
```
Oscar-Pepper, line 169: CHANGE REQUEST: why is a rescan required if there are checkpoints higher than the truncate height. this will always be the case for truncation. i dont understand what was wrong with the initial logic. all top MAX_REORG_ALLOWANCE blocks are guaranteed to have checkpoints so if a checkpoint is not found a rescan is triggered... i feel like this is overengineered and complicated for no reason

pepper-sync/src/wallet/traits.rs
```diff
@@ -345,65 +348,106 @@ pub trait SyncShardTrees: SyncWallet {
...
+    /// at `truncate_height`, stays untouched because it records nothing
+    /// above that height, or — holding state it cannot roll back —
+    /// aborts with [`SyncError::TruncationError`] so the caller can fall
+    /// back to the clear-and-rescan recovery. No decision is made here;
+    /// the plan is authoritative, and a planned rollback the tree store
+    /// unexpectedly refuses is reported as the same truncation error.
     fn truncate_shard_trees(
         &mut self,
         truncate_height: BlockHeight,
+        plan: &ShardTreeTruncationPlan,
```
Oscar-Pepper: minor: we should have all the data we need to just truncate to a given height, why is it necessary to have to do pre-processing on a plan whyen all the logic can be contained in this method?

pepper-sync/src/witness.rs
```diff
@@ -94,6 +94,40 @@ where
...
+    let roots = shard_tree.store().get_shard_roots().expect("infallible");
+    let Some(newest) = roots.last() else {
+        return 0;
+    };
+    let bare = shard_tree
+        .store()
+        .get_shard(*newest)
+        .expect("infallible")
+        .is_some_and(|shard| shard.root().is_leaf());
+    (roots.len() - usize::from(bare)) as u32
```
Oscar-Pepper, line 146: CHANGE REQUEST: add_shard_ranges has not been updated to account for this refetching of the highest server-fetched root. this will result in duplicate shard ranges which is the fundamental data needed for scan range splitting

zingolib/src/lightclient/migrate.rs
```diff
@@ -268,6 +271,9 @@ impl LightClient {
         if hash != consented_plan_hash {
             return Err(MigrationError::ConsentStale.into());
         }
+        if !plan.is_split() {
+            return Err(MigrationError::NoteSplittingRequired.into());
+        }
```
Oscar-Pepper, line 276: what about the case where:
a) notes have already been split
b) notes happen to be in the exact denomation + part fee, so no splitting is required

Oscar-Pepper: ah there seems to be an issue here actually. in the context of how zingo-cli "migration" command works, there is no note-splitting sub-command. shouldnt note-splitting be part of the start of migration?

Oscar-Pepper (changes requested):

pepper-sync/src/shardtree_ext.rs
Oscar-Pepper: looking at shard tree this may actually be a bug/limitation of the library so its not necessarily incorrect here

zancas: Per your first point this enum is only used to express the result of appending a checkpoint to the rightmost leaf.

I am updating the extension trait method name.

PR #2520, opened 2026-07-23
feat: match ZIP 318 so our migration patterns collide with every conforming wallet's

zancas:
This draft tracks the work of matching our Orchard→Ironwood migration exactly to ZIP 318 (`draft-schell-ironwood-migration`).

The guiding principle of this branch inverts the usual engineering instinct: the more our policy and patterns collide with everyone else's, the better. The migration's privacy rests entirely on indistinguishability. Every conforming wallet quantizes into the same canonical Denominations, emits Parts of identical shape (one Orchard spend, one Ironwood output, no change, the canonical fee), and anchors at the same network-wide Boundaries. A wallet is protected precisely to the extent that its migration transactions are indistinguishable from — that is, collide with — every other wallet's. Any place where our policy diverges from the ZIP, however locally clever, is a fingerprint that shrinks the anonymity set for our users and for everyone else. Conformance here is not compliance overhead; it is the security property.

Accordingly, the commits that land here will audit and tighten our quantization, scheduling, and part-construction policy against the draft ZIP, and will treat any observable deviation from the population-wide pattern as a bug.

This branch stacks on #2509 (`fix/ironwood-split-mobile`); until that merges, the diff shown here duplicates it. The ZIP-318 conformance commits will appear on top, and the shared portion of the diff collapses once #2509 lands.


zancas:
Superceded by #2545

PR #2529, opened 2026-07-25
fix!: read both orphaned wallet layouts; salvage any unreadable wallet file

zancas:
Supersedes #2528, which targeted `fix/ironwood-split-mobile`. This defect is present on `dev` itself, so the fix belongs here.

## The defect is live on dev

Both commits that created the orphaned layouts are ancestors of `dev`: `32261bb5f` removed the `allow_v6_transactions` byte and bumped the wallet version to 43, and `4158e20c2` re-pinned the version to 42 without the byte, ruling that the byte-carrying revision "never shipped." Version 42 therefore names two incompatible layouts, and version 43 names a third file population.

`dev` reads `32..=42` with no disambiguation, so it misparses a byte-carrying 42 file one byte into the price list and rejects a 43 file outright. Anyone who built and ran from `dev` between those commits has a wallet that current `dev` cannot open — which is exactly what a tester reported against a mobile build.

## The fix

`LightWallet::read` accepts version 43 as the final 42 layout (the layouts are byte-identical) and disambiguates the two revisions of 42 by an end-of-file-anchored parse of the buffered tail. The canonical layout is preferred, so files written by current code never depend on the fallback, and a tail that parses cleanly *both* ways is refused rather than guessed — accepting the wrong reading would silently substitute a different price list and migration state. Ambiguity cannot arise for a migration-free tail, whose length is always odd while the two readings differ by exactly one byte; the refusal guards the migration-bearing case, where no such parity argument holds.

Beneath that sits a salvage floor: `LightWallet::read_recovery_info` recovers the seed phrase, birthday, and account count from the stable version-32+ prefix of any file whose full parse fails, and `zingo-cli` falls back to it when the wallet will not load but the user asked for `recovery_info`. No format break can strand a wallet.

Version 43 is burned and must never be reassigned; the next format bump is 44.

## Policy

ADR 0015 records the ruling that motivated this: we support users who build and run from `dev`, so no release gate protects them, and a wallet format is shipped the moment it lands in `dev`. Shipped layouts stay readable, used version numbers are burned, and a format change lands only together with its read-side compatibility and regression tests. The `zingolib` glossary gains Wallet Version, Shipped Format, and Recovery Salvage.

## Tests

Covering version 43, both bool values of the byte-carrying layout, prefix salvage from a corrupt file, all four outcomes of the ambiguity decision, and a sweep of a local `data_wallets` corpus requiring every real wallet file to parse or salvage (gitignored, prints no seed material, passes empty in CI).

The layout tests assert on the file *tail*, not on recovery info: recovery info is read from the prefix and survives a misparsed tail intact, so a prefix-only assertion would pass against a reader that dropped the entire migration section. Both defenses were confirmed by mutation — disabling the trailing-byte check fails the tests through the refusal, and disabling both fails them on the tail comparison.

Disk suite passes 20/20 on this branch; `zingolib` and `zingo-cli` are clippy- and fmt-clean.

### DIFF
--- a/zingolib/CONTEXT.md
+++ b/zingolib/CONTEXT.md
@@ -215,6 +215,12 @@
 
 **Dirty Flag** — An internal boolean on `LightWallet` that records whether unsaved changes exist. Set automatically by all mutating operations; cleared after a successful write. Exposed to external code via `LightWallet::mark_dirty()`.
 
+**Wallet Version** — The number leading a Wallet File that identifies its layout. A Wallet Version, once used by any revision that reached users, is burned: it is never reassigned to a different layout. (Version 43 is burned; version 42 names two layouts that the reader tells apart.)
+
+**Shipped Format** — A Wallet File layout that has landed in dev. Landing in dev is the shipping event — there is no later release gate protecting dev builders — and every Shipped Format remains readable, and the wallet writable, forever after. See ADR 0015.
+
+**Recovery Salvage** — The last-resort read that recovers the seed phrase, birthday, and account count from the stable prefix of a Wallet File whose full parse fails. A backstop beneath the Shipped Format guarantee, not a substitute for it: restoring from seed forfeits local transaction metadata and forces a rescan.
+
 ---
 
 ## Consumers
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ libtonode-tests/store_all_checkpoints_test
 .zingo-cli/
 .DS_Store
 .agent-plans/
+
+# Local corpus of real wallet files (live seed material — never commit)
+/data_wallets/
--- a/docs/adr/0015-landing-in-dev-ships-the-wallet-file-format.md
+++ b/docs/adr/0015-landing-in-dev-ships-the-wallet-file-format.md
@@ -19,9 +19,13 @@ reassigned; and a format change may land in dev only together with the
 read-side compatibility that keeps every previously shipped file
 loading, proven by regression tests in the same change. Version 42's
 two layouts are both read, disambiguated by an end-of-file-anchored
-dual parse of the file tail, which is deterministic because the
-revisions differ by exactly one byte. Version 43 is burned; the next
-format bump is 44.
+dual parse of the file tail: the canonical layout is preferred, so a
+file written by current code never depends on the fallback, and a tail
+that parses cleanly *both* ways is refused rather than guessed. The
+refusal covers a case that cannot arise for a migration-free tail,
+whose length is always odd while the two readings differ by exactly
+one byte, but for which no such parity argument holds once a migration
+section is present. Version 43 is burned; the next format bump is 44.
 
 Beneath the compatibility guarantee sits a salvage floor for files
 outside it — corrupt files, files from abandoned side branches, files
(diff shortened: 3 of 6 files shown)


zingolib/CONTEXT.md
dorianvp, line 1: Nice!

zancas:
Local test run against b235403c4ae0e5b21d15118470a1f52b5cb511a2:


```
────────────
     Summary [ 367.191s] 40 tests run: 40 passed (25 slow), 6 skipped
[cargo-make][2] INFO - Build Done in 368.65 seconds.
[cargo-make][1] INFO - Build Done in 369.05 seconds.

====================== test summary ==========================
  packages:     356 run,  356 passed,    0 failed,    0 timed out,   14 skipped
  zingo-cli:     97 run,   97 passed,    0 failed,    0 timed out,    0 skipped
  libtonode:     40 run,   40 passed,    0 failed,    0 timed out,    6 skipped
  TOTAL:        493 run,  493 passed,    0 failed,    0 timed out,   20 skipped
==============================================================
  note: skipped counts include the other phases' tests — phases share one --workspace build scope, filtered per phase.
[cargo-make] INFO - Build Done in 676.84 seconds.
```

zancas:
## Review — approve

I verified the load-bearing claims against the code rather than taking the description's word.

The parity argument on `resolve_v42_tail` holds: `PriceList::write` emits one version byte, two `Optional` flag bytes, a `CompactSize` (widths 1, 3, 5, 9 — odd, growing by even increments), and 8-byte prices, and the migration `Optional` adds one flag byte, so a migration-free tail is always odd-length while the two readings differ by exactly one byte. The salvage prefix parse in `read_recovery_info` mirrors `read_v32` byte-for-byte, correctly skips chain-type validation, and correctly rejects empty-seed view-only files. The layout tests assert on the re-serialized tail, whose bytes are the true suffix of the file, so `tail_offset()` is sound and the deliberately non-trivial tail gives the assertion teeth. Version 43 flows through the sequential branch as the final 42 layout and re-saves as 42, which is the intended burn. The CLI fallback opens the same path startup uses.

Four minor observations, none blocking. First, because `PriceList::serialized_version()` is 0, every canonical tail begins with a byte in {0, 1}, so the pre-release parse always runs; a canonical migration-bearing file could in principle parse both ways and hit the ambiguity refusal despite unambiguous provenance — the odds are minuscule and salvage covers it, but the refusal can theoretically bite a file the fallback was never needed for. Second, `print_salvaged_recovery_info` drops the startup error when `File::open` itself fails, so a missing wallet file reports a bare OS error without context. Third, `resolve_v42_tail` discards the pre-release parse error in the both-fail case, which could point diagnosis of a corrupt genuinely-pre-release file the wrong way. Fourth, the version-32+ prefix is now load-bearing policy for salvage, but only ADR 0015 says so; a sentence in the `serialized_version` docs would put the constraint where the next format-changer will look.

The disambiguation is principled rather than heuristic, the refusal is argued from the encoding rather than asserted, and the tests were mutation-checked against the exact failure mode that would have made them vacuous. Approved.

🤖 Review by [Claude Code](https://claude.com/claude-code)

zancas (commented):
AI Approved.

PR #2548, opened 2026-07-27
fix: restore price fetch

dorianvp:


zancas (approved):
We ought to think about different price sources... but I guess that's out of scope.

PR #2561, opened 2026-07-27
Fielded network-failure taxonomy: typed stages, targets, and cause chains end to end

zancas:
@Oscar-Pepper I would build the cargo doc for this new package and read the doc from lib.rs.   That should give you clear examples of how this tool is relevant and useful.


zancas:
## Addendum: two commits added after review

The review of this PR surfaced one substantive finding: the fetch paths
returned the price but never stored it. `PriceList::current_price` sat
behind a public getter that nothing ever wrote, and the wallet marked
itself dirty after an operation that changed no state. Commit bf06a3a47
closes that gap. `PriceList` gains `record_current_price` as the storage
half of an update, `update_current_price` records what it fetches
(making its `&mut self` honest), and `LightWallet::record_price_update`
now takes the fetched `Price` and stores it under the briefly-held lock,
which makes the `save_required` flag truthful. The field already
round-trips through the price list's serialization, so this is a value
change, not a format change; a new zingo-price test pins the round trip
of a recorded price. The same commit fences the surviving lock-holding
path: `LightWallet::update_current_price` is now deprecated, because it
holds the wallet lock across the network wait — exactly the polling
blackout this PR removes — and it remains only as the unit under test
for the dead-proxy and black-holed-proxy falsifiers, which allow the
lint deliberately.

Commit f701b1053 gives `zingo-net-diag` its front-page howto. The crate
doc now opens with why the taxonomy exists (prose failures can be read
but never dispatched on; the Connection Doctor was reduced to
substring-matching sentences), explains the two roles the crate serves
(producers classify at the seam that owns the error; consumers dispatch
on typed fields), and walks both roles in one doctest that also pins the
`Display` contract. The prose cites
`zingolib::nym::socks5_transmit_stage` as the production producer and
`LightClient::mixnet_death_detail` as the consumer-visible payoff.
`README.md` is generated from the crate doc with
`cargo rdme --intralinks-strip-links`, the pepper-sync convention;
`Cargo.toml` names the readme and records the command.

Neither commit carries `mobile-adaptation-required`: the surfaces the
mobile pin reads are unchanged (`record_price_update` is crate-private,
the deprecation sits on a wallet-level method mobile does not call, and
the zingo-price addition is purely additive). Verified with
`cargo fmt --check`, clippy in both feature configurations, the
zingo-net-diag and zingo-price suites (including the new round-trip test
and the doctest), the touched zingolib subset (price, probe, supervisor,
fan-out, and death-detail tests, 23 passing under the `nym` feature),
`cargo doc` with `-D warnings`, and `cargo rdme --check` confirming the
README matches the crate doc.

zancas:
TODO analyze pros-and-cons of making all of this logic be a zingolib mod, instead of a separate package.

zancas:
## Addendum: the two review follow-ups now ride this PR

Commit 2d48add89 closes #2562. The provider connect race in
`NymProxy::start` was the last nym-touching path reporting failures as
flattened prose. It now mirrors the fan-out's shape: the driver retains
each arm's failure as a typed record, and an exhausted race surfaces
`NymProxyError::AttemptsExhausted { attempts, failures: Vec<zingo_net_diag::NetOpFailure> }`,
with the joined-prose account surviving only in `Display`.
Classification happens at the seam that owns the error type: a pure,
total `provider_attempt_stage` match maps gateway registration to
`remote-connect`, the per-attempt bound to `timed-out(20000ms)`, and
the port draw to `route-resolution`, with the shortened provider name
as the target. The netutils workspace takes the optional path
dependency on `zingo-net-diag` that the design document anticipated
(gated on `nym`; the crate's zero-dep std-only design is what lets it
resolve in that workspace's own lockfile). `Build` and `Connect` also
move their causes into `source()` so a chain walk sees one text per
layer, and `RaceState::failure_summary` is removed with its last
caller.

Commit 756c7479a closes #2563: two supervisor tests pin the
`death_detail` mode gate — a died transport surfaces its typed cause
whole, every other mode surfaces `None` even over a stale record, and
a spawned child's death (a closed stdout pipe, no typed cause)
surfaces `None` rather than a fabricated record.

Verified with `cargo fmt --check`, clippy in all four
workspace/feature configurations, `cargo nextest run` (netutils: 27
default / 45 under `nym`; zingolib: 15 supervisor tests under `nym`),
and `cargo doc` with `-D warnings`. Neither commit carries
`mobile-adaptation-required`: `AttemptsExhausted` is consumed
in-process only (the spawned binary renders it to stderr), and the
supervisor change is test-only.

Oscar-Pepper (dismissed):
local tests pass but not tested nym feature manually via zingo-cli

zancas:
Superceded by #2579 which specifically aims to make zingo-cli / nym more testable (and relevant to other consumers).

PR #2572, opened 2026-07-28
error when ironwood data is missing from compact block scanning

Oscar-Pepper:
currently, the user may be left with unspendable ironwood notes when a server only paritally serves ironwood data. this PR prefers an error is returned so the user cannot progress in scanning until the server is switched to serve all the necessary ironwood data


zancas (approved):

zancas:
Nice.

Issue #2573, opened 2026-07-28
Decide the fate of the migration sizing slack: fee-fold or residual change note


zancas:
Oscar's review of #2428 questioned the economics of the migration dust floor: value below the floor "not migrated and folded into fees … seems like a lot for the user to pay for migration" (https://github.com/zingolabs/zingolib/pull/2428#discussion_r3537094086).

Most of that concern has since been resolved in Oscar's favor. The floor is no longer local policy: it is the ZIP 318 `MAX_RESIDUAL_VALUE` (0.01 ZEC), imported from the `zcash_pool_migration` reference crate and pinned by the `zip318_conformance_tripwires` tests. Notes worth at most `sweep_min` are never selected; they remain in the pre-Ironwood pool as residual, tracked in `MigrationPlan::residual` and covered by the consent hash. A pool that cannot fund even the smallest denomination is likewise left entirely residual, and the planner unwinds any consolidation rounds that would have served it.

One case still folds value into fees: the sizing slack. When the pooled value funds at least one denomination, the fixed-point sizing in `plan_migration` selects denominations whose sum plus overhead is at most the pooled total, and the final note-split transaction emits exactly the sized targets, so the input surplus — strictly less than the smallest denomination, hence bounded below 0.01 ZEC per migration — becomes that transaction's implicit fee. The consented plan determines the slack arithmetically, but the plan does not display it as a line item.

This issue tracks the open decision:

1. Keep the fee-fold, and (optionally) surface the slack explicitly in the plan presented for consent; or
2. Emit the slack as a pre-Ironwood V2 change note that the residual policy holds and a later replan can sweep, at the cost of an extra output per migration and of value stranded at pool retirement if never swept.

A sub-`MAX_RESIDUAL_VALUE` value cannot be a standard Ironwood note in either design, so the choice is between paying the slack forward as fee at consent time and parking it in the retiring pool.

PR #2574, opened 2026-07-28
refactor!: answer the #2428 review threads that were still live in code

zancas:
Your review of #2428 left thirteen threads unresolved when the PR merged. This PR is the collected answer. It changes the code where a comment still described it, and it accounts for every other thread below, so each can be resolved from here.

## Changed in this PR

**`get_migration_heights` returns the heights directly** ([thread](https://github.com/zingolabs/zingolib/pull/2428#discussion_r3537628868)). Agreed, and done. The helper wrapped an infallible call yet returned `Result<Option<_>>`, and all five callers unwrapped the `None` case with the same `ok_or(WalletError::NoSyncData)`. The function now returns the heights directly and maps the `None` case to `NoSyncData` itself, which deleted the identical `ok_or` at all five call sites. The wrapper survives for one reason, which its doc now states: it pins the wallet's own `min_confirmations` and types the no-sync-data case once.

**The confirmation-wait errors drop their `Split` names** ([thread](https://github.com/zingolabs/zingolib/pull/2428#discussion_r3550815924)). Agreed. The function is generic over migration transactions, yet its errors said "split". This PR renames the two shared variants to `MigrationTransactionFailed` and `MigrationConfirmationTimeout`; `SplitDidNotConverge` keeps its name, being genuinely split-only. The rename is breaking on the public `MigrationError` enum.

**The duplicated comment is gone** ([thread](https://github.com/zingolabs/zingolib/pull/2428#discussion_r3550543915)). The doc on `broadcast_due_parts_selected` carried its two paragraphs twice, verbatim. Fixed.

**The ironwood balance comment now answers your question** ([thread](https://github.com/zingolabs/zingolib/pull/2428#discussion_r3529898930)). You asked why compact block scanning would fail to find positions for ironwood. It does not fail: scanning finds ironwood positions, and pepper-sync carries ironwood located trees and a nullifier-to-position map. A note lacks a position only until a scan locates its block in the commitment tree, and the wallet records its own outputs at broadcast, before any scan, so those notes hold no position and cannot be witnessed. The comment in `shielded_spendable_balance` now states this reason instead of asserting the conclusion.

## Resolved by work that landed since the review

**The `COIN` constant** ([thread](https://github.com/zingolabs/zingolib/pull/2428#discussion_r3535820553)). Agreed: in this codebase "coin" names a transparent output, and the constant collided with that sense. The local constant is gone. `params.rs` now imports `zcash_protocol::value::COIN`, so the workspace defines no such name of its own.

**The `dust_floor` field** ([thread](https://github.com/zingolabs/zingolib/pull/2428#discussion_r3535837660)). The field is renamed, though not to `denom_floor`. Your name has the virtue of saying what the field does: it floors the denomination ladder. We chose `max_residual_value` instead, because ZIP 318 names the quantity `MAX_RESIDUAL_VALUE` and the reference crate exports it under that name; a field that mirrors the specification reads without translation.

**An error when sync data is missing** ([thread](https://github.com/zingolabs/zingolib/pull/2428#discussion_r3535920358)). Done as you suggested. `WalletError::NoSyncData` now exists, and every planning and broadcast path returns it when `last_known_chain_height` is `None`. This PR folds the same mapping into `get_migration_heights`, so the migration height reads are typed as well.

**Visibility of `migration_note_values`** ([thread](https://github.com/zingolabs/zingolib/pull/2428#discussion_r3535972815)). The function is `pub(crate)` now.

**Visibility of `build_note_split_transaction`** ([thread](https://github.com/zingolabs/zingolib/pull/2428#discussion_r3535976179)). The function is `pub(crate)` now.

**Protection before NU6.3 activation** ([thread](https://github.com/zingolabs/zingolib/pull/2428#discussion_r3550585693)). Later work added the protection you asked about. The start path refuses with `IronwoodEraTooYoung` until the first window boundary after NU6.3 activation, and the scheduler floors every anchor at activation, so a pre-activation release fails once with a typed error instead of striking the build error repeatedly.

**Checkpoint count versus bucket modulus** ([thread](https://github.com/zingolabs/zingolib/pull/2428#discussion_r3550606149)). The code landed where your third comment arrived. Raising the checkpoint retention would have guaranteed every boundary a witness, but, as you said, it would store extra data forever for the sake of a single migration. Instead, `refresh_part_witnesses` caches each part's witness while its boundary checkpoint is still retained, and a part whose boundary is pruned takes the `MissedBoundary` skip and is re-assigned.

## Kept as it is

**The `SyncDomain` re-export** ([thread](https://github.com/zingolabs/zingolib/pull/2428#discussion_r3535531922)). Removing the re-export would give `SyncDomain` a single canonical path and spare the reader one indirection. We kept it nonetheless: `wallet/traits.rs` and `sync/spend.rs` both import `crate::SyncDomain` through it, and because it is `pub(crate)` it adds nothing to the crate's public surface.

## Tracked onward

**The dust-floor economics** ([thread](https://github.com/zingolabs/zingolib/pull/2428#discussion_r3537094086)). Later work resolved most of this in your direction. The floor is no longer local policy: it is ZIP 318's `MAX_RESIDUAL_VALUE` (0.01 ZEC), imported from the reference crate. Notes at or below `sweep_min`, and any pool too small to fund the smallest denomination, now stay in the pre-Ironwood pool as residual — recorded in the plan and covered by the consent hash — rather than folding into fees. One fee-fold remains: when a pool funds at least one denomination, the sizing surplus below the smallest denomination becomes the final split transaction's fee. That surplus is bounded below 0.01 ZEC per migration, but your underlying point stands — the user should not pay value away silently — so #2573 tracks whether the surplus should become a residual V2 change note instead.

### DIFF
--- a/zingolib/src/lightclient/error.rs
+++ b/zingolib/src/lightclient/error.rs
@@ -120,10 +120,10 @@ pub enum MigrationError {
     /// A migration transaction (a note split or a part) failed or
     /// disappeared from the wallet.
     #[error("Migration transaction {0} failed or disappeared.")]
-    MigrationTransactionFailed(TxId),
+    TransactionFailed(TxId),
     /// Migration transactions were not confirmed within the polling window.
     #[error("Timed out waiting for migration transactions to confirm.")]
-    MigrationConfirmationTimeout,
+    ConfirmationTimeout,
     /// The scheduled flow was asked to start over a plan that still needs
     /// note splitting, which no scheduled-flow driver executes yet.
     #[error(
--- a/zingolib/src/lightclient/error.rs
+++ b/zingolib/src/lightclient/error.rs
@@ -117,13 +117,13 @@ pub enum MigrationError {
     /// Note splitting kept producing new rounds past the round bound.
     #[error("Migration did not converge within {0} rounds.")]
     SplitDidNotConverge(usize),
-    /// A note-splitting transaction failed or disappeared from the wallet.
-    #[error("Note-splitting transaction {0} failed or disappeared.")]
-    SplitTransactionFailed(TxId),
-    /// Note-splitting transactions were not confirmed within the polling
-    /// window.
-    #[error("Timed out waiting for note-splitting transactions to confirm.")]
-    SplitConfirmationTimeout,
+    /// A migration transaction (a note split or a part) failed or
+    /// disappeared from the wallet.
+    #[error("Migration transaction {0} failed or disappeared.")]
+    MigrationTransactionFailed(TxId),
+    /// Migration transactions were not confirmed within the polling window.
+    #[error("Timed out waiting for migration transactions to confirm.")]
+    MigrationConfirmationTimeout,
     /// The scheduled flow was asked to start over a plan that still needs
     /// note splitting, which no scheduled-flow driver executes yet.
     #[error(
--- a/zingolib/src/lightclient/migrate.rs
+++ b/zingolib/src/lightclient/migrate.rs
@@ -584,8 +584,10 @@ impl LightClient {
         &self,
         account: zip32::AccountId,
     ) -> Result<MigrationPlan, LightClientError> {
-        let wallet = self.wallet().read().await;
-        Ok(wallet.plan_ironwood_migration_now(account)?)
+        self.wallet()
+            .read()
+            .await
+            .plan_ironwood_migration_now(account)
     }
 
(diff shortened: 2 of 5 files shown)


Oscar-Pepper (dismissed):
not merging until https://github.com/zingolabs/zingolib/pull/2582#pullrequestreview-4807484754 is resolved

PR #2575, opened 2026-07-28
refactor!: de-stutter the migration errors and pin the NoSyncData contract

zancas:
Stacked on #2574, implementing the two suggestions from its review.

**The variant stutter is gone.** `MigrationError::MigrationTransactionFailed` and `MigrationError::MigrationConfirmationTimeout` repeated the enum's name at every use site. They become `TransactionFailed` and `ConfirmationTimeout`, which reads as Rust convention intends at the match site. The display strings are unchanged, so no user-facing text moves. The rename is breaking on the public `MigrationError` enum, exactly as the base PR's rename was, and rides the same major bump.

**The `NoSyncData` contract is pinned.** #2574 gave `get_migration_heights` the job of typing the absent-sync case once, and documented it, but nothing enforced the promise. A unit test now constructs a wallet that has never synced and asserts the wrapper returns `WalletError::NoSyncData` itself, so every migration site inherits the typed refusal from one place and the doc comment's claim is a tested contract.

### DIFF
--- a/zingolib/src/lightclient/error.rs
+++ b/zingolib/src/lightclient/error.rs
@@ -120,10 +120,10 @@ pub enum MigrationError {
     /// A migration transaction (a note split or a part) failed or
     /// disappeared from the wallet.
     #[error("Migration transaction {0} failed or disappeared.")]
-    MigrationTransactionFailed(TxId),
+    TransactionFailed(TxId),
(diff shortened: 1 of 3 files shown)


Oscar-Pepper (approved):
not merging until https://github.com/zingolabs/zingolib/pull/2582#pullrequestreview-4807484754 is resolved

PR #2579, opened 2026-07-29
Ratify consumer convergence (ADR 0024) and mint the Mixnet Mode wire contract

zancas:
This branch carries the mixnet stack (the net_diag_taxonomy → socks5_typed_timeout → attach_retune line) plus the convergence arc; the reviewable novelty is the top three commits.

`e041f6dd2` ratifies ADR 0024: the three consumers (zingo-cli, zingo-mobile, zingo-pc) converge on a zingolib-owned mixnet surface — one mint, N renderers. zingolib owns the wire contract, the session driver with the single liveness clock, the proxy-path resolver, startup consent, and the netutils re-exports; consumers render typed variants and never restate tokens or match prose. The same commit appends the ADR 0011 amendment reinstating the 2026-07-23 mixnet-only price rule in full (superseding the 2026-07-27 clearnet restoration merged as #2548: with routing left to the callers, both shipping consumers fumbled it identically), and adds the glossary sentence making a startup-time opt-out the explicit act that reaches switched off.

`d0227dd8e` implements the mint: MixnetMode gains the ratified wire tokens (unattached, switched_off, bootstrapping, ready, died) as a serde snake_case representation and an as_str/Display/FromStr trio that agree by test, plus the exhaustive ALL array. The retired token off is rejected by parse and pinned so by test. serde_json enters dev-dependencies only.

`acf0c9822` labels the tokens as minted per the standing documentation rule, naming the mint's four enforcement mechanisms in the enum doc.

Tracking: closes nothing yet; phase 1 continues under #2578, with zingo-mobile#1235/#1236 and zingo-pc#338 downstream.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

**Stacks on #2561** (net_diag_taxonomy → dev), the one open PR in this branch's ancestry: until it merges, this PR's diff includes that content. The rest of the mixnet chain beneath this branch (netutils_time, socks5_typed_timeout, the attach gate, #2569) has already merged upward into the stack line.

Issue #2578, opened 2026-07-29
Converge the consumer mixnet surface in zingolib (ADR 0024 phase 1)


zancas:
ADR 0024 (branch `zingo-cli-over-nym`, `docs/adr/0024-consumers-converge-on-a-zingolib-owned-mixnet-surface.md`) ratifies one mint, N renderers for the three consumers (zingo-cli, zingo-mobile, zingo-pc). Phase 1 is the zingolib side, on the attach_retune line:

- The canonical wire representation of the five-state `MixnetMode`: Display/FromStr (and serde where it already fits) with the ratified strings `"unattached"`, `"switched_off"`, `"bootstrapping"`, `"ready"`, `"died"`, plus a typed status struct. Today the wire strings are minted in zingo-mobile's FFI shim — that mint dies in phase 2.
- A session driver owning forced-on for connected sessions, the recovery predicate, startup consent (a `--no-mixnet`-style opt-out lands `SwitchedOff`, per the glossary's new consent-at-start sentence), and the ONE liveness clock, with status delivered to consumers by subscription (the proxy shim's death callback is the pattern). The CLI's forced-on block (`zingo-cli/src/lib.rs:743-771`) and zingo-mobile's TypeScript `MixnetCoordinator` both retire onto this.
- The proxy-path resolver: the CLI's 4-tier precedence (explicit → `$ZINGO_NYM_PROXY` → bundled-location hint → PATH) moves into `zingolib::nym`, parameterized by platform hints; the driver takes Spawn{hints} or Attach{addr}.
- The reinstated mixnet-only price rule (ADR 0011 amendment 2026-07-28, superseding the 2026-07-27 clearnet restoration): the driver refuses price in every state except Ready; non-nym builds compile no fetch.
- Re-exports of the netutils items consumers legitimately need (`GrpcIndexer`, `Indexer`, the wire types), so consumers funnel through one rev-pinned zingolib dependency (this is the #2566 funnel fix; it removes zingo-mobile's capability-less duplicate netutils copy).

The 2026-07-28 audit's CLI divergence list (21 sites) shrinks to renderers only in the same arc: the duplicated https-only gate, the second proxy-start error mapping, and the path-resolution prose all retire.

---

[...]

PR #2580, opened 2026-07-29
Provisioning moves into zingolib and --no-mixnet records consent (ADR 0024 phase 1)

zancas:
Stacks on #2579; second arc of ADR 0024 phase 1 (#2578).

`909e64e34` moves proxy-path provisioning into `zingolib::nym::provision`: the four-tier precedence (explicit, then `ZINGO_NYM_PROXY`, then the hinted bundled directory, then the bare name on PATH, empty values counting as absent) becomes one rule parameterized by `SpawnHints`, so each platform names only what zingolib cannot know — its bundled-binary directory. The environment-variable name and the binary name become labeled minted tokens with the module as their sole production site. zingo-cli's resolver trio collapses to a one-call delegation passing its executable-sibling hint, and the precedence tests move down with the policy.

`96782eb7b` repairs a semantic conflict the dev back-merge could not see textually: `mixnet_death_report` arrived written against the retired four-state `mixnet_proxy` field and now reads through the five-state slot, as its sibling does. (The same fix is cherry-picked onto the base branch as `6270e3206`.)

`9a30a07f4` closes the `--no-mixnet` regression the convergence audit found: under five states the flag left a session `Unattached` — refusing send and price-fetch — while its documentation promised clearnet. Per ADR 0024's consent-at-start decision (and the glossary's amended Mixnet Mode entry), the startup opt-out is the explicit act: the forced-on block now routes the flag through `disable_mixnet`, landing `SwitchedOff` exactly as an in-session `nym off` would. A new zingolib test pins the disable-from-fresh transition the flag relies on.

Verified with `cargo check`/`clippy` on both crates and 15/15 targeted tests (`provision`, `wire_contract`, the new consent pin) under `--features nym,nym-diary`.


Oscar-Pepper (approved):

Oscar-Pepper:
awaiting conflict resolution of PR chain https://github.com/zingolabs/zingolib/pull/2583

PR #2582, opened 2026-07-29
feat: pin the Sync State Contract with a validator, attack corpus, and read enforcement

zancas:
A session auditing every site that sets `sync_state` found the documented scan-range invariant ("in block height order with no overlaps or gaps") enforced only by construction inside the sync engine, and unchecked at the one production seam that can produce a non-trivial state: `SyncState::read` on wallet load. A wallet file violating the invariants did not fail loudly — it silently corrupted every consumer that trusts them, from `fully_scanned_height` to the spend horizon, and a file carrying an inverted scan range panicked the load inside the `ScanRange::from_parts` assert. This PR states the contract explicitly and makes the seam enforce it.

`SyncState::validate` is a pure, side-effect-free validator over the self-contained invariants: scan-range well-formedness and contiguity, shard-range order per pool, and tree-bound orientation. It reports the first violation in deterministic field-then-index order as a `SyncStateIntegrityError`, one variant per distinct violation, each carrying the convicting evidence. `SyncState::read` now rejects inverted scan-range bytes as `InvalidData` before the constructor assert can panic the process, and validates the assembled state before returning it.

The contract was built test-first from an adversarial corpus: one hostile wallet state per test, each violating a distinct invariant, plus honest controls, two file-seam attacks mounted through serialized bytes, and a maximally adversarial wallet whose violations the closing test peels off one at a time in the documented reporting order. Constructing the corpus proved two boundaries. Inverted scan ranges are unrepresentable — the constructor asserts and the fields are private — so the enum carries no variant for them, and a should-panic test witnesses why. And the previously-scanned counts admit no bound at all: a mid-session reorg truncation legally shrinks the scanned ranges below the counts recorded at session start, so rejecting such states would convict legally produced wallets. The defect there is the status math, whose six `u32` subtractions in `get_sync_status` now saturate instead of underflowing in the post-truncation window; a corpus control pins that truncation drift passes validation.

The base is #2575, whose `NoSyncData` contract work is adjacent; this PR merges after it in the stack.

### DIFF
--- a/pepper-sync/src/wallet/integrity.rs
+++ b/pepper-sync/src/wallet/integrity.rs
@@ -1,0 +1,604 @@
+//! The Sync State Contract: a pure validator over [`SyncState`].
+//!
+//! [`SyncState`] documents invariants on its fields (notably that
+//! `scan_ranges` is "in block height order with no overlaps or gaps")
+//! but, until this module, nothing checked them: the sync engine
+//! preserves them by construction, while [`SyncState::read`] accepted
+//! arbitrary bytes from the wallet file unvalidated. A state violating
+//! the invariants does not fail loudly — it silently corrupts every
+//! consumer that trusts them, from [`SyncState::fully_scanned_height`]
+//! to zingolib's spend horizon.
+//!
+//! [`SyncState::validate`] is the contract's witness: a pure,
+//! side-effect-free function that checks every self-contained invariant
+//! — every claim decidable from a `&SyncState` alone — and reports the
+//! first violation as a [`SyncStateIntegrityError`], one variant per
+//! distinct violation, each carrying the evidence. Violations are
+//! reported in deterministic field-then-index order: scan ranges first
+//! (well-formedness before contiguity at each index), then shard ranges
+//! per pool in sapling, orchard, ironwood order, then tree bounds in
+//! the same pool order.
+//!
+//! [`SyncState::read`] enforces the contract at the wallet-file seam: a
+//! file whose ranges violate it fails to load with `InvalidData`
+//! instead of silently poisoning consumers. Because `read` never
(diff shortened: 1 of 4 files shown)


pepper-sync/src/wallet/integrity.rs
```diff
@@ -0,0 +1,604 @@
...
+                if range.start > range.end {
+                    return Err(SyncStateIntegrityError::InvertedShardRange {
+                        pool,
+                        index,
+                        start: range.start,
+                        end: range.end,
+                    });
+                }
+                if let Some(next) = shard_ranges.get(index + 1)
+                    && range.end > next.start
```
Oscar-Pepper, line 207: in most cases, a single block will contain note commitments for the end of one shard *and* the start of the next shard. this means that the *exclusive* range end will often be 1 higher than the next shard ranges start bound, but sometimes be equal to it where the boundary falls exactly at the end of the block. this means that we should be asserting that each range end is either equal to or 1 larger than the next shard ranges start bound

Oscar-Pepper: to be clear:
- scan ranges are contiguous
- shard ranges may contain the same block twice at the boundary

Oscar-Pepper (changes requested):

pepper-sync/src/wallet/integrity.rs
Oscar-Pepper: another note for context, if there are blocks between shards with no note commitments the end bound is already extended up to the next shard range to prevent gaps during shard range construction. gaps are not permitted but overlaps of 1 block are permitted. this prevents holes in scanning by covering the entire chain by at least one shard range but allows an entire shard to be prioritised, guaranteeing all note commitments for the prioritised shard will be scanned.

PR #2592, opened 2026-07-29
feat(cli)!: nym compiles by default; network on is the in-session consent act

zancas:
This branch makes `nym` a default feature of zingo-cli and renames the REPL command from `nym` to `network`. Turning the network on is now an in-session Connectivity Consent act: `network on` switches an offline session to ONLINE MODE, the help text and command output warn in capitals, the probe still refuses while offline, and the `change_server` refusal pin reads live state. Builds that opt out of the feature are offline-only — they carry no `network` command, their online launch acts refuse, and stored consent is inert. ADR 0026 records the default flip, the consent act, and the rename.

A second commit fixes the testutils clearnet-consent funnel to key on zingolib's own `nym` feature rather than the testutils crate's: feature unification can enable `zingolib/nym` from any workspace member, and a local cfg would compile the consent out exactly when the refusal is compiled in.

Verified on both feature shapes after the rebase onto the ADR 0030/0031 command table: `cargo clippy -D warnings` and `rustfmt` clean, 178/178 default and 159/159 opt-out zingo-cli tests, and 336/336 zingolib lib tests. The `network` body now runs async below the crate's single seam, which resolves review 4852724030's nested-runtime panic structurally.


zingo-cli/src/commands.rs
```diff
@@ -703,42 +704,55 @@ impl Command for CurrentPriceCommand {
...
+            WARNING: in an offline session, `network on` is itself the consent
+            act: it switches the session to ONLINE MODE, for this session only.
+            The session selects an indexer over the same curated ranking that
+            `--online` uses at launch, and only then bootstraps the mixnet
+            (ADR 0026).
+
+            `status` reports off, bootstrapping or ready. `on` starts the
+            nym-proxy child, taking the binary from the given path, else
+            $ZINGO_NYM_PROXY, else one bundled beside this binary, else PATH.
+            `off` reverts to clearnet without leaving ONLINE MODE. `probe` runs
```
Oscar-Pepper: "`off` reverts to clearnet without leaving ONLINE MODE." is this out of date? network off should go offline right? maybe this should be called "disconnect_nym" or "nym_off" or "clearnet"?

Oscar-Pepper (changes requested):
panic reached during testing:

```
$ ./target/release/zingo-cli --data-dir ~/wallets/test_wallet_1/ --offline
Launching save task...
(main) Block:0 [Sync stopped] >> help
Standalone commands (no wallet required):
  help - Lists all available commands
  parse_address - Parse an address
  parse_viewkey - Parse a view_key.
  servers - Show ranked indexer servers and response times
  version - Get version of build code

Wallet commands:
  addresses - List unified addresses in the wallet.
  balance - Return the wallet ZEC balance for each pool (account 0).
  birthday - Returns block height wallet was created
  calculate - Sign the latest proposal without transmitting it.
  change_server - Change indexer server
  check_address - Checks if the given encoded address is derived by the wallet's keys.
  clear - Clear the wallet state, rolling back the wallet to an empty state.
  coins - Show the wallet's coins (transparent outputs).
  confirm - Build and transmit the latest proposal, then resume sync.
  current_price - Updates and returns current price of ZEC.
  delete - Delete wallet file from disk
  drain - Send all Orchard funds into the Ironwood pool now (immediate ZIP 318 path).
  export_ufvk - Export unified full viewing key for the wallet.
  height - Print the chain height as of the wallet's last request to the server.
  info - Get the indexer server's info
  max_send_value - Print the most the wallet can send to a given address.
  memobytes_to_address - Show by address memo_bytes transfers for this seed.
  messages - List memos for this wallet.
  migrate - Migrate all Orchard funds to the Ironwood pool in one interactive run.
  migration - Drive the scheduled Orchard to Ironwood migration
  network - Control the network posture and mixnet transport; `network on` switches an offline session to ONLINE MODE.
  new_address - Create a new unified address.
  new_taddress - Create a new transparent address.
  new_taddress_allow_gap - Create a new transparent address (even if the last one did not receive any funds).
  notes - Show the wallet's notes (shielded outputs).
  quicksend - Send ZEC, fusing `send` and `confirm`.
  quickshield - Shield transparent funds, fusing `shield` and `confirm`.
  quit - Quit the light client, saving state to disk.
  recovery_info - Print the wallet's seed phrase, birthday and account count.
  remove_transaction - Remove a failed transaction from the wallet.
  rescan - Clear all chain-derived wallet data and sync again from the birthday.
  save - Launch the save task. Not meant to be called by hand.
  send - Propose a transfer of ZEC, for 'confirm' to broadcast.
  send_all - Propose a transfer of all shielded ZEC to one address, for 'confirm' to broadcast.
  sends_to_address - Show by address number of sends for this seed.
  settings - Show or set wallet settings.
  shield - Propose a shield of transparent funds, for 'confirm' to broadcast.
  spendable_balance - Display the wallet's spendable balance.
  split - Split Orchard notes into ZIP 318 part sizes, one round per call.
  sync - Sync the wallet to the latest state of the blockchain.
  t_addresses - List transparent addresses in the wallet.
  transactions - List the wallet's transaction summaries by block height.
  transmit - Transmit calculated transactions to the Indexer.
  value_to_address - Show by address value transfers for this seed.
  value_transfers - List all value transfers for this wallet.
  wallet_kind - Displays the kind of wallet currently loaded
(main) Block:0 [Sync stopped] >> help network
Control the session's network posture and its mixnet transport
(the mixnet is Nym; the name is implicit).

With Mixnet Mode on, send and price-fetch route over the mixnet and
fail closed while it bootstraps, never falling back to clearnet.
Turning it off is a deliberate choice to transmit over clearnet.

WARNING: in an offline session, `network on` is itself the consent
act: it switches the session to ONLINE MODE, for this session only.
The session selects an indexer over the same curated ranking that
`--online` uses at launch, and only then bootstraps the mixnet
(ADR 0026).

`status` reports off, bootstrapping or ready. `on` starts the
nym-proxy child, taking the binary from the given path, else
$ZINGO_NYM_PROXY, else one bundled beside this binary, else PATH.
`off` reverts to clearnet without leaving ONLINE MODE. `probe` runs
GetLightdInfo over both routes side by side to tell whether a
failure is mixnet-specific, and its clearnet leg uses your real IP;
an offline session refuses it. `history` shows per-indexer attempts
across sessions, and needs the nym-diary feature plus
--indexer-diary.

Usage:
network status | on [binary_path] | off | probe [uri] | history

(main) Block:0 [Sync stopped] >> network on
No --server specified. Probing 12 indexers...

thread '<unnamed>' (69861) panicked at /home/oscar/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.50.0/src/runtime/scheduler/multi_thread/mod.rs:88:9:
Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
Error executing command network: receiving on a closed channel


thread 'main' (69844) panicked at zingo-cli/src/lib.rs:294:38:
called `Result::unwrap()` on an `Err` value: SendError { .. }
```

Oscar-Pepper:
nym is now default yet i see nothing in the readme to explain the nym setup:

```
$ ./target/release/zingo-cli --data-dir ~/wallets/test_wallet_1/ --online
Error: Failed to start the Nym mixnet proxy: failed to spawn the nym-proxy binary at nym-proxy: No such file or directory (os error 2). Mixnet Mode is required for a connected session; install the nym-proxy binary, pass --nym-proxy <path>, set $ZINGO_NYM_PROXY, or pass --no-mixnet to transmit over clearnet this session.
```

zingo-cli/src/commands.rs
zancas: Good catch!

zancas:
There's still follow-up work required to select indexers..    I think that should stack on this, since it's required for e.g. transmitting.

Oscar-Pepper (approved):

zingo-cli/src/commands.rs
Oscar-Pepper: i dont see this issue resolved as of 80dab5d3d32f6cdd47d2f98730cb1bfac3a5010f

Oscar-Pepper (changes requested):
there are still issues with the help text of the "network" command. "network off" should disconnect all network capability as described here:
"off      Disconnect every network capability of the session, keeping any stored consent" but this is contradicted by other parts of the help text:
"`off` reverts to clearnet without leaving ONLINE MODE" and
"Turning it off is a deliberate choice to transmit over clearnet."

if "off" disconnects all network activity then there should also be a way to turn the mixnet off and transmit over clearnet instead

zancas:
Remaining concerns issued.

PR #2601, opened 2026-07-30
refactor: split the two senses of "batch" into load and decryption batch

zancas:
Pepper-sync used one word for two different units. This change gives each its own name, stacked on #2599.

The batcher's unit is a maximal contiguous sequence of whole blocks whose combined output count stays within a budget — the block that would cross the line starts the next unit. It is now a **load**, built by the `Loader`, bounded by `max_load_outputs` and `MAX_LOAD_NULLIFIERS`. The trial-decryption unit is a flattened, transaction-aligned buffer of one pool's outputs, indifferent to block boundaries. It keeps **batch**, qualified at the types as `DecryptionBatch`, `DecryptionBatchRunner`, and friends, staying continuous with the upstream `zcash_note_encryption::batch` API it wraps; bare "batch" inside the runners module still reads unambiguously as the decryption sense, so module-local prose and locals keep it.

The worker-side budget takes the neutral name `max_outputs`, because `scan` derives both the decryption batch size and the located-tree build size from it; only the value handed to the runners is `decryption_batch_size`, retiring the inverted `trial_decrypt_task_size` name. The crate docs and mermaid diagrams follow the split, and the loading step's description now states the true cut rule — a load is cut at the last whole block that keeps the output budget unexceeded — replacing the previous "a fixed number of outputs" wording, which described the mechanism incorrectly.

The rename is behavior-preserving and entirely `pub(crate)`-internal: no public API surface changes.

### DIFF
--- a/pepper-sync/src/scan.rs
+++ b/pepper-sync/src/scan.rs
@@ -125,7 +125,7 @@ pub(crate) async fn scan<P>(
     consensus_parameters: &P,
     ufvks: &HashMap<AccountId, UnifiedFullViewingKey>,
     scan_task: ScanTask,
-    max_batch_outputs: usize,
+    max_outputs: usize,
 ) -> Result<ScanResults, ScanError>
 where
     P: consensus::Parameters + Sync + Send + 'static,
@@ -195,7 +195,7 @@ where
             &consensus_parameters_clone,
             &ufvks_clone,
             initial_scan_data,
-            max_batch_outputs / 8,
+            max_outputs / 8,
         )
     })
     .await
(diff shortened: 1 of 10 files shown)


pepper-sync/src/scan.rs
```diff
@@ -50,10 +53,20 @@ impl InitialScanData {
...
+                let prev_bounds = prev.tree_bounds();
+                let ironwood_initial_tree_size = if prev_bounds.ironwood_final_tree_size == 0 {
+                    first_block.chain_metadata.map_or(0, |chain_metadata| {
+                        chain_metadata.ironwood_commitment_tree_size.saturating_sub(
+                            block::shielded_output_count(first_block, ShieldedPool::Ironwood),
+                        )
+                    })
+                } else {
+                    prev_bounds.ironwood_final_tree_size
+                };
```
Oscar-Pepper: we must remove this logic otherwise it will hide the key indicator that we have synced ironwood blocks with an ironwood incompatible version previously and therefore not trigger the ironwood rescan.

Oscar-Pepper (changes requested):
looks really good, just one change request

pepper-sync/src/scan.rs
zancas: Fixed.

PR #2603, opened 2026-07-30
fix: treestate ironwood bug

dorianvp:


pepper-sync/src/sync/state.rs
```diff
@@ -932,6 +941,263 @@ pub(super) fn calculate_scanned_blocks(sync_state: &SyncState) -> u32 {
...
+where
+    W: SyncBlocks,
+{
+    [
+        scan_range.block_range().start,
+        scan_range.block_range().end - 1,
+    ]
+    .into_iter()
+    .filter(|&block| wallet.get_wallet_block(block).is_ok())
+}
```
Oscar-Pepper: this fn seems unnecessary, if the boundary blocks of any scan range are not in the wallet this is a critical error and will fail when creating a sync status

pepper-sync/src/sync/state.rs
```diff
@@ -932,6 +941,263 @@ pub(super) fn calculate_scanned_blocks(sync_state: &SyncState) -> u32 {
...
+{
+    let Some(highest) = wallet
+        .get_sync_state()?
+        .scan_ranges()
+        .iter()
+        .flat_map(|scan_range| retained_blocks(wallet, scan_range))
+        .max()
+    else {
+        return Ok(ChainEvidence::Unreadable);
+    };
```
Oscar-Pepper: cant we just take the highest block in the wallet? its a BTreeMap so its ordered by height

pepper-sync/src/chain.rs
```diff
@@ -0,0 +1,115 @@
+//! The chain the sync engine syncs against.
+
+use zcash_protocol::{
+    consensus::{self, BlockHeight, NetworkType, NetworkUpgrade},
+    local_consensus::LocalNetwork,
+};
```
Oscar-Pepper: we have our own network types and activation height types in zingo_common, I want us to avoid using multiple. we will be using the zingo_common versions for the public API anyway so it makes sens to just use one

pepper-sync/src/wallet.rs
```diff
@@ -149,22 +150,45 @@ pub struct SyncState {
     pub(crate) scan_targets: BTreeSet<ScanTarget>,
     /// Initial sync state.
     pub(crate) initial_sync_state: InitialSyncState,
+    /// The chain being synced. Not serialized: it is resolved from the
+    /// wallet's configured chain on load, so a wallet can never carry
+    /// activation heights that disagree with the chain it is opened
+    /// against.
+    pub(crate) chain_parameters: ChainParameters,
```
Oscar-Pepper: the chain params get fed into sync, so im not sure why they are duplicated into the sync state? i think we should avoid multiple sources of truth

Oscar-Pepper (changes requested):
I know this is still draft but i want to leave a snippet of my current concerns with this approach. I think this solution seems heavy-handed, introducing additional scan range priorities for partial rescans is a large bug surface and I feel the additional complexity is a big cost to pay where i believe there could be a more simple solution involving our current rescan/truncation framwork. I will need time to ponder as I understand that truncation is limited to a small window before the user would require a full rescan but I think we need to weigh up the pros and cons here before this merges. i will add more detail later when i have time

pepper-sync/src/wallet/traits.rs
```diff
@@ -359,6 +359,24 @@ pub trait SyncShardTrees: SyncWallet {
...
+    /// this is the rollback a pool needs when its recorded history cannot be
+    /// trusted and no checkpoint survives to roll back to.
+    fn clear_pool_shard_tree(&mut self, pool: ShieldedPool) -> Result<(), SyncError<Self::Error>> {
+        let shard_trees = self.get_shard_trees_mut().map_err(SyncError::WalletError)?;
+        tracing::info!("Clearing {pool:?} shard tree.");
+        match pool {
+            ShieldedPool::Sapling => shard_trees.sapling = empty_shard_tree(),
+            ShieldedPool::Orchard => shard_trees.orchard = empty_shard_tree(),
+            ShieldedPool::Ironwood => shard_trees.ironwood = empty_shard_tree(),
+        }
```
Oscar-Pepper: its very important for our checkpoint logic to insert a zero height checkpoint when creating a new tree. see ShardTrees::new()

Oscar-Pepper: this should be added to `empty_shard_tree` i suppose

Oscar-Pepper: i see that a checkpoint is immediately added for the birthday frontiers, including an empty tree checkpoint if the frontier is empty. so this is not a bug, but i believe this is a foot gun in case `empty_shard_tree` is used in the future without immediately adding a checkpoint so its better safe than sorry

pepper-sync/src/scan/compact_blocks.rs
```diff
@@ -298,59 +298,85 @@ fn check_continuity(
...
 
+/// Checks every pool's commitment tree size against the chain.
+///
+/// A block's metadata size is the chain's own count of that pool's note
+/// commitments up to and including the block, so the wallet's count of the
+/// outputs it was served must equal it. Two ways to disagree, told apart by
+/// what the block actually served:
+///
+/// Metadata reporting nothing while the block serves outputs for that pool
+/// means the server does not serve that pool. Failing sync against such a
```
Oscar-Pepper: im not sure i understand this comment, if there are ironwood outputs in teh block then the server serves ironwood even if the chain metadata tree size reported is incorrect right?

pepper-sync/src/scan/compact_blocks.rs
```diff
@@ -298,59 +298,85 @@ fn check_continuity(
...
-    if metadata_size != 0 && calculated_size != 0 {
-        matched(ShieldedPool::Ironwood, metadata_size, calculated_size)?;
-    } else if metadata_size != calculated_size {
-        tracing::warn!(
-            "ironwood tree size mismatch at block {}.\nwallet block: {}\ncompact block metadata: {}",
-            wallet_block.block_height(),
+        if metadata_size == 0 && served_outputs > 0 {
+            tracing::warn!(
+                "{pool:?} outputs served at block {} against a chain metadata size of zero: \
+                 this server does not serve {pool:?}",
```
Oscar-Pepper: similar to above

pepper-sync/src/scan/compact_blocks.rs
```diff
@@ -717,40 +743,86 @@ mod tests {
...
+    /// block still carries that pool's outputs. Failing sync there would
+    /// strand every wallet using such a server, and the wallet's own record
+    /// is not what is at fault, so this is tolerated.
+    #[test]
+    fn a_server_not_serving_a_pool_does_not_fail_the_scan() {
+        let compact_block = block_with_served_ironwood_actions(5, 0);
+        let wallet_block = wallet_block_with_ironwood_size(5);
+
+        assert!(check_tree_size(&compact_block, &wallet_block).is_ok());
+    }
```
Oscar-Pepper, line 770: this test is related to my comment above

Oscar-Pepper (commented):
non-blocking change requests. I much prefer this approach to the previous solution as its very clear and there is a much smaller potential bug surface for edge cases etc. i will approve when this is ready for review

Oscar-Pepper (approved):

PR #1250, opened 2024-06-26
Test Expected Fee + Fix

fluidvanadium:

PR #2608, opened 2026-07-31
feat: reformat git_description into the zl_ build descriptor

zancas:
This reworks `git_description()` in `zingolib/build.rs` to emit a build descriptor instead of raw `git describe` output, per the format specified in zingolabs/zingo-mobile#1250.

The format is `zl_<ver>[_<numcommit>_<hash5>][_dirty]`, derived from the `zingolib_v<ver>` release tags. The bracketed fields are elided when the build sits exactly on its tag, so a tagged release reads `zl_3.0.1` while today's dev tip reads `zl_3.0.1_962_aa7e8`. The hash keeps five characters and drops describe's `g` prefix. `_dirty` marks a build from an uncommitted tree.

Two behavior notes beyond the reformat. The describe call now passes `--match=zingolib_v*`, so a non-release tag can no longer become the version base. The no-git fallback (published crate, container workspace without `.git`) emits `zl_<crate version>` in the same shape as an on-tag build.

This string reaches every consumer of `zingolib::git_description()`, zingo-cli included. The zingo-mobile side of #1250 appends a `-zm_…` part for the mobile checkout in a separate PR.

Verified by `cargo check -p zingolib` and by inspecting the generated `git_description.rs`, which reads `zl_3.0.1_962_aa7e8_dirty` for the pre-commit working tree (dirty from the then-uncommitted `build.rs` itself).

### DIFF
--- a/zingolib/build.rs
+++ b/zingolib/build.rs
@@ -65,23 +65,60 @@ fn git_path_query(flag: &str) -> Option<PathBuf> {
     Some(PathBuf::from(path))
 }
 
+fn descriptor(raw: &str, tag_prefix: &str, part: &str) -> String {
+    let (body, dirty) = match raw.strip_suffix("-dirty") {
+        Some(stripped) => (stripped, true),
+        None => (raw, false),
+    };
+    let fields: Vec<&str> = body.rsplitn(3, '-').collect();
+    let formatted = match fields.as_slice() {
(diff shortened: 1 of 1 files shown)


dorianvp (approved):
Nice! This deserves a line in the readme but that can come later

PR #2609, opened 2026-08-01
docs: reduce verbosity

dorianvp:
The idea behind this PR is to reduce LLM-generated text verbosity.


zancas (approved):
I want my agents to know this now.. so.. I suppose more of the theme can be in another PR!

PR #2613, opened 2026-08-03
fix: delete the uncalled balance_delta method

### LINKED ISSUE #2612: balance_delta under-reports a Zennies-donating self-send by the donation amount
## What happens

`TransactionSummary::balance_delta` (`zingolib/src/wallet/summary/data.rs:175-185`)
reports only `-fee` for `SendToSelf` transactions. A self-send that carries a
Zennies For Zingo donation still classifies as `SendToSelf`, because
`transaction_kind` (`zingolib/src/wallet/transaction.rs:163-226`) exempts the
donation address from its external-recipient check. pepper-sync's
`total_value_sent` (`pepper-sync/src/wallet.rs:860`) carries no such exemption
and counts the donation as external value sent, so the summary arrives with a
nonzero `value` and a `SendToSelf` kind. The two layers disagree about whether
the donation address is "self", and `balance_delta` sits on the fault line: for
every donating self-send it reports `-fee` while the wallet's balance actually
moved by `-(donation + fee)`.

## Failing test

The branch `balance_delta_misreport` adds a red test to the existing fixture
module in `summary/data.rs`, constructing exactly the summary shape the wallet
produces for a donating self-send and asserting the report equals the amount
actually spent:

```rust
#[test]
fn a_donating_self_send_balance_delta_is_the_amount_actually_spent() {
    let mut summary = self_send_summary(vec![PoolType::ORCHARD], vec![], vec![note(90_000)]);
    summary.value = 10_000;
    summary.fee = Some(5_000);
    let amount_actually_spent = (summary.value + summary.fee.unwrap()) as i64;
    assert_eq!(summary.balance_delta(), Some(-amount_actually_spent));
}
```

```
FAIL [ 0.003s] (1/1) zingolib wallet::summary::data::tests::a_donating_self_send_balance_delta_is_the_amount_actually_spent

thread '...' panicked at zingolib/src/wallet/summary/data.rs:1191:9:
assertion `left == right` failed
  left: Some(-5000)     <- balance_delta: the fee alone
 right: Some(-15000)    <- the amount the wallet actually spent (donation + fee)
```

## Impact

The 2026-08-03 consumer audit found no current consumer logic reading
`balance_delta` in zingo-pc, zingo-mobile, or zingo-cli, so the wrong number is
latent in the API surface rather than user-visible today. It becomes visible to
the first consumer that renders a balance delta.

A follow-up sweep sharpened this: the method has zero callers anywhere. No
Rust call site exists in the zls workspace beyond the definition, zingolib
never serializes the value into any JSON (the summary `json::object!` blocks
carry no such field), zingo-mobile's UDL does not name it (so UniFFI generates
no binding), and no TypeScript in either GUI references it. `balance_delta` is
dead public API: `pub` and `#[must_use]`, called by no one. Under the privacy
ratchet, deletion is therefore a live remediation alongside the fix — though
if it stays, the movement below corrects it structurally.

## Where the fix likely belongs

The exemption is an editorial judgment sitting upstream of fact arithmetic.
Moving it into the zingo-viewmodel projection (per the ADR 0024 amendment
architecture and the `viewmodel_seed` slice, PR #2611) removes the misreport
structurally: `transaction_kind` becomes purely key-derived, a donating
self-send classifies `Sent(Send)`, and the existing `Send` arm computes the
true `-(value + fee)` with no special case. A narrower patch — a
donation-aware `SendToSelf` arm in `balance_delta` — would also fix the number
but adds a third site that knows the donation address.

zancas:
This pull request deletes `TransactionSummary::balance_delta`, which was `pub` and `#[must_use]` yet had no caller anywhere: the 2026-08-03 audit found no Rust call site in this workspace beyond the definition, no serialization into any summary JSON, no UniFFI binding (zingo-mobile's UDL never names it), and no reference in either GUI's TypeScript. The compiler confirms the audit — `cargo check --workspace --all-targets` passes with the method gone.

The method also reported the wrong number for a Zennies-donating self-send: `transaction_kind` exempts the donation address, so the kind reads `SendToSelf` and the middle arm reports `-fee`, while pepper-sync's `total_value_sent` counts the donation as external value — the wallet actually moves the donation plus the fee. Deleting dead API that misreports is the privacy-ratchet resolution of the defect; a future consumer wanting balance deltas should derive them after the Zennies exemption moves to the viewmodel projection, where the `Send` arm arithmetic is already correct.

Closes #2612.


Oscar-Pepper (dismissed):
changelog!

PR #2614, opened 2026-08-04
refactor: delete the test-only paid_fees and txids methods

zancas:
The sibling audit behind issue #2612 found `TransactionSummaries` kept two more `pub`, `#[must_use]` methods with no production caller anywhere: `paid_fees`, whose only caller was the `get_fees_paid_by_client` testutils helper, and `txids`, called only by one mock-chain test and one libtonode twin. Per the privacy ratchet, public API whose sole constituency is test code comes down: the fee summation now lives inside the testutils helper it existed for, and the two `txids` call sites inline the one-line map.

The 33 fee assertions across the mock-chain and libtonode suites are untouched. On this branch the workspace all-targets check, workspace fmt, and the mock-chain and summary-data suites (13 tests) all pass.

Companion to #2613, which deletes the equally-uncalled `balance_delta`; the two are independent and land in either order.


Oscar-Pepper (dismissed):
changelog!

PR #2617, opened 2026-08-04
feat: re-derive the mixnet projection from the typed status snapshot

zancas:
This slice adds the mixnet projection the seed deliberately excluded — "re-derived from the zingolib-owned typed status in its own later slice, never ported," per the seed's claim file and the 2026-08-03 migration rulings. The old branch's `mixnet.rs` proved unportable: it renders the dead tri-state and re-mints the retired `off` token that `mode.rs` now rejects.

`zingo-perspective/src/mixnet.rs` derives everything from `zingolib::nym::MixnetStatus`. The status line absorbs zingo-cli's `render_status` wording verbatim, together with its three wording-pin tests; the disclaimer composition absorbs `render_status_with_disclaimer`, keeping zingolib's canonical `IP_CORRELATION_DISCLAIMER` as the one source of the sentence; the JSON view is a parse-and-render of zingolib's golden-pinned serde wire, so no second JSON vocabulary can drift into existence; and a `MixnetStatusSnapshotExt` trait reads the session's status channel whole. The CLI's `nym status` arm delegates to the projection and its local renderers retire.

Verified with the nym feature on and off: workspace fmt, 15/15 zingo-perspective tests, both doc-tests, and zingo-cli's six nym tests, including the offline-contract test that exercises the delegated path end to end.

Based on `viewmodel_seed` and intended to land after it; it rebases mechanically if the seed advances.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2611, opened 2026-08-03
feat!: the zingo-perspective extraction (funnel + value_transfer slice)

zancas:
The seed slice of the zingo-perspective re-extraction, implementing the rulings recorded in ADR 0024's 2026-08-03 amendment (carried in this PR) and originating in dorianvp's boundary analysis (hackmd HJ0jWQh4zx).

## What this does

The new `zingo-perspective` crate takes two roles. As the editorial layer, it receives the value-transfer derivation, the finsight rollups, and their types, reached through extension traits (`LightWalletPerspectiveExt`, `LightClientPerspectiveExt`) that preserve every method name, signature, and output shape zingolib carried before the extraction. As the funnel of rule 7, its modules re-export zingolib's consumer surface path-for-path, so a governed consumer declares exactly one wallet dependency and repoints by renaming it. The path-for-path guarantee covers the shipped consumer surface only; test scaffolding is re-exported as a convenience with no path-stability promise, as the glossary's Funnel entry records.

zingo-cli is the first tenant: its wallet dependency is now zingo-perspective, every import renamed mechanically, with a single trait import added at the editorial call sites. libtonode-tests repoints its value-transfer imports while keeping both dependencies, since test scaffolding is not a governed consumer.

The finsight rollups ride this slice as a forced dependent: `do_total_*` are built on `value_transfers()` and could not remain in zingolib once the derivation left. The mixnet projection is deliberately absent; it will be re-derived from the zingolib-owned typed status (ADR 0024 rules 1–2) in its own slice, never ported.

## Contract safety

The eight golden fixtures captured from the pre-extraction implementation (branch `viewmodel-crate`) all reproduce byte-identically against this extraction — JSON and Display renderings of `value_transfers`, `messages_containing`, and the three rollups. The capture mechanism was removed once the capture was complete, so the goldens can never be silently regenerated; an intended contract change must re-mint its rows by hand, in review. Ten editorial tests relocated with the code (seven derivation, three mock-chain); canonical tests stayed in zingolib. Five summary primitives (`pools_present`, `shielded_notes_by_pool`, `received_memos`, `self_received_value`, `received_memo_value`) were promoted to `pub`, one rung each, driven by the cross-crate diagnostics.

## Verification

[...]

PR #2623, opened 2026-08-05
feat!: land the editorial layer as zingolib::perspective, not a crate

zancas:
Supersedes #2611 and #2617. The 2026-08-04 grilling session declined the zingo-perspective extraction before merge: the funnel's repoint churn — every governed consumer repointing its manifest and imports across three repositories, then keeping a second pin synchronized forever — outweighs the gate maintenance the in-crate form costs. The `viewmodel_seed` branch joins `viewmodel-crate` as inventory of the editorial layer, mined but never rebased.

## What this does

The editorial layer lands as the `zingolib::perspective` module (children `value_transfer` and `finsight`), compiled behind the crate's first default-on feature, `perspective`. The feature is a boundary marker, not a capability toggle: CI's cargo-hack feature-powerset job compiles the perspective-off subset of every combination under `-D warnings`, so zingolib building without its editorial layer is a standing compiler-checked fact. In the other direction there is deliberately no mechanism — no import allowlist, no workbench gate — and the ADR 0024 amendment carried in this PR records that asymmetry as an accepted cost, along with the reversal's grounds and the analysis inherited from the extraction.

Because the derivation returns to inherent `impl LightWallet` and `impl LightClient` blocks, no call site changes anywhere: consumers rewrite only the `use` lines naming the moved types (`zingolib::perspective::value_transfer::{ValueTransfer, …}`), and rule 7 stands as originally written — consumers declare exactly one wallet dependency, and it is zingolib. The five summary primitives the extraction had promoted stay `pub(crate)`, now cfg-gated to the feature that consumes them. The naming ruling survives the placement reversal: the layer is perspective, singular, one house perspective with N renderers.

The `finsight()` aggregate the seed crate minted rides the port: one derivation yields every rollup, and the `do_total_*` methods delegate through it. The mixnet projection remains deliberately absent; it re-derives from the zingolib-owned typed status in a later module slice, where #2617's content re-homes.

## Contract safety

The eight golden fixtures captured from the pre-extraction implementation port unchanged from the seed branch and reproduce byte-identically against the module form — JSON and Display renderings of `value_transfers`, `messages_containing`, and the three rollups. The goldens remain never-regenerate pins: an intended contract change must re-mint its rows by hand, in review. The seven derivation tests and golden harness land as `zingolib/tests/` integration tests gated on `all(perspective, testutils)`; the three mock-chain editorial tests and the `create_various_value_transfers` fixture gate in place.

## Verification

The workspace compiles with `--all-targets`; clippy (including the `nym,nym-diary` lane), rustfmt, and `cargo doc` are clean; the four perspective/testutils feature combinations compile warning-free. Test runs: 8/8 golden and derivation, 17/17 zingolib summary and mock-chain, 167/167 zingo-cli.


zingolib/Cargo.toml
```diff
@@ -9,7 +9,13 @@ homepage = "https://github.com/zingolabs/zingolib"
 license = "MIT"
 
 [features]
-default = []
+default = ["perspective"]
```
dorianvp, line 12: default?

dorianvp: Hmm once we pull out perspective and into a separate crate we might wanna remove this feature too

zingolib/src/perspective/finsight.rs
```diff
@@ -0,0 +1,98 @@
+//! The financial-insight rollup types, relocated from zingolib's summary
+//! data module when their derivation moved to this crate.
```
dorianvp, line 2: I think we should remove narration of past events in doc comments. How are these useful to a first-time reader in docs.rs?

zingolib/src/perspective/value_transfer.rs
```diff
@@ -0,0 +1,295 @@
+//! The value-transfer editorial types, relocated from zingolib's summary
+//! data module when the derivation moved to this crate.
```
dorianvp, line 2: Same as in finsight

zingolib/src/perspective.rs
```diff
@@ -0,0 +1,401 @@
+//! The editorial layer: the house perspective over the canonical wallet
+//! record, deriving presentation-facing statements that never feed wallet
+//! logic. Compiled behind the default-on `perspective` feature, whose off
+//! build proves core stands without this layer (ADR 0024, amendment
```
dorianvp, line 4: IMO we should also not reference an ADR or amendment in documentation that appears in docs.rs

dorianvp (approved):
Left some comments

PR #2453, opened 2026-07-13
chore!: retire darkside-tests, zingo-testutils, lwd, and zcashd

zancas:
This draft stacks on #2444 and carries the package-simplification work promised in the #2448 review discussion: it retires the `darkside-tests` and `zingo-testutils` crates (commit 72b7d3172).

`darkside-tests` drove darksidewalletd to test wallet reorg behavior, but all nine of its tests had been ignored for months on two lightwalletd darkside-mode bugs. The crate therefore protected nothing while holding the legacy stack in the dependency graph and carrying pepper-sync's subtractive `darkside_test` feature, whose activation under `cargo --workspace` stripped `GetSubtreeRoots` from every crate by feature unification (#2447). Its reorg protections now live in the mock-indexer darkside module (`zingolib/src/lightclient/darkside.rs`), where they run offline in seconds. `zingo-testutils` was already a dead directory: not a workspace member, depended on by nothing; the live utility crate is the untouched `zingolib_testutils`.

The orphaned `darkside_test` and `darkside_tests` cargo features remain declared but unenabled — nothing can turn them on, so #2447's poison is cured — and their cfg sites await a mechanical sweep. ADR 0005 records the decision, the rejected gated-preservation alternative, and what eliding lightwalletd itself still turns on.

The branch also carries the ironwood-migration merge with its decision ledger, and the zingo-cli harness work (a `zcash_local_net` `Wallet` implementation for the `zingo-cli` binary, plus `--activation-heights` TOML support), which the crate retirement builds on.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2626, opened 2026-08-05
refactor(cli)!: typed command table, single async seam, stdout carries only the result

zancas:
This PR realizes ADR 0030 and ADR 0031. Every zingo-cli command is now one row of a static spec table: its name is minted once from its body function's identifier, its capability is a typed `CommandBody` variant (a standalone body never receives the wallet; the REPL-owned `servers` cannot claim a body), and its async body runs below the crate's single sync-to-async seam, which a Clippy `disallowed-methods` rule enforces. This eliminates the nested-runtime panic class that review 4852724030 of PR #2592 reached through `network on`, along with the per-dispatch `HashMap` rebuild, the fifty `impl Command` scaffolds, and the hand-patched `servers` help line.

Breaking, per ADR 0031: a failing command renders once as `Error: …` on stderr and one-shot mode exits nonzero; stdout carries exactly the command's result, with the in-band `{"error": …}` JSON objects deleted; unknown commands are errors; and all narration (transmit heartbeat, server probing, save/sync notices, the quit trailer) moves to stderr. Consumers that parse stdout for errors or assert exit 0 must adapt — the known audit surface is the cli-wallet-harness work in PR #2453.

The command tests move to `commands/tests.rs` (169 passing, including new table invariants: names sorted and unique, and every usage section invokes its command's minted name), the non-test `commands.rs` shrinks from 3,316 lines to 2,867, and clap is bumped to 4.6.5 ahead of a follow-up branch that will delegate subcommand parsing, help, and usage errors to clap's derive API.

### DIFF
--- a/docs/adr/0030-the-cli-crosses-sync-to-async-exactly-once.md
+++ b/docs/adr/0030-the-cli-crosses-sync-to-async-exactly-once.md
@@ -1,0 +1,72 @@
+---
+status: accepted
+date: 2026-08-05
+---
+
+# The CLI crosses sync to async exactly once, dispatching from a static command table
+
+`zingo-cli` accumulated some sixty-four `RT.block_on` call sites because every
+command privately chose where the sync world ended. Any call path that stacked
+two of them panicked with "Cannot start a runtime from within a runtime," and
+review 4852724030 of PR #2592 reached exactly that panic in testing: `network
+on`, already inside `block_on`, called the synchronous `select_servers()`
+wrapper. We decided that command bodies are ordinary `async` functions
+returning typed results, that the sync-to-async crossing happens at exactly
+one seam — `do_user_command` for the string frontends, plus the startup driver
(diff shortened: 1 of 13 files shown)


docs/adr/0030-the-cli-crosses-sync-to-async-exactly-once.md
```diff
@@ -0,0 +1,72 @@
+---
+status: accepted
+date: 2026-08-05
+---
+
+# The CLI crosses sync to async exactly once, dispatching from a static command table
+
+`zingo-cli` accumulated some sixty-four `RT.block_on` call sites because every
+command privately chose where the sync world ended. Any call path that stacked
+two of them panicked with "Cannot start a runtime from within a runtime," and
```
zancas, line 10: There were 0 Command stacks that panicked.  The only command stack was an exit->save stack that didn't nest block_on.

PR #2630, opened 2026-08-06
refactor(cli)!: delegate command parsing to clap derive grammars

zancas:
This PR delegates zingo-cli's input processing to clap's derive API, completing the arc ratified after PR #2626. Four commits, each gate-green under both feature configurations:

1. **Family grammars** (`c80eae1f7`): the migration, nym, drain, split, sync, and save families parse through `#[derive(clap::Parser)]` grammars with typed fields — the plan hash as `[u8; 32]`, spacings as `Duration`s with defaults in the grammar, the probe target as an https-only `http::Uri`. Ten parse-error variants leave the typed error enums; those refusals belong to the grammars now. The byte-identity parser tests retire deliberately with the hand parsers whose strings they pinned.
2. **The table absorbed** (`715cea930`): the static `CommandSpec` table, `CommandBody`, and the name-minting macros dissolve into one alphabetical `#[derive(clap::Subcommand)] CliCommand` enum; dispatch is an exhaustive match, so a variant without a body or a body without a name cannot compile. Bodies take parsed inputs only, and `settings` gains a real subgrammar so no argument is ever inspected under the wallet lock. The send family keeps a trailing string payload for its JSON-or-positional hybrid, the one documented compromise.
3. **Parse once at the boundary** (`b61a83b95`): the outer app absorbs `CliCommand` as real subcommands, so a one-shot command parses in main's argument pass — malformed invocations fail with clap's usage error, exit code 2, before any wallet work — and `zingo-cli --help` lists every command. The REPL parses each line locally, so parse errors never cross the command channel, which now carries the parsed `CliCommand`. `do_user_command` is deleted; `do_user_command_result` remains the string entry point.
4. **CHANGELOG** (`4b583a0f7`), including a correction to the Unreleased entry from #2626 whose `do_user_command`/`do_user_command_result` roles this arc inverts.

Breaking, and recorded in the CHANGELOG: generated help and error prose replaces the pinned hand-parser strings; usage errors exit 2 pre-startup; command names are case-sensitive; `exit` cleanly aliases `quit`; a bare `save` refuses at the parse.

Deferred follow-ups, tracked in `.agent-plans/clap-commands-arc.md`: retiring the hand-written Usage prose from the long help now that clap generates usage (and the invariant test pinning it); a revision of ADR 0030's registry description, which this arc makes historical; the send-family grammar absorption; and the worker's failure interpolations naming commands by their `Debug` rendering.

The branch also rides the post-#2629 base, so the lockfile story is clean: clap's derive feature adds `clap_derive` and syn 3 (both proc-macro, compile-time only), and the workspace otherwise holds a single syn 2.

### DIFF
--- a/zingo-cli/src/commands.rs
+++ b/zingo-cli/src/commands.rs
@@ -1104,7 +1104,6 @@ pub enum NymCommandError {
 
 /// A parsed `nym` command, its arguments parsed completely at the clap
 /// derive grammar before any wallet access.
-#[cfg(feature = "nym")]
 #[derive(clap::Subcommand, Clone, Debug, PartialEq, Eq)]
 pub(crate) enum NymSubCommand {
     Status,
@@ -1121,7 +1120,6 @@ pub(crate) enum NymSubCommand {
 
 /// https-only: the mixnet leg refuses a plaintext target at dial time,
 /// so the grammar rejects it up front with a clear message.
-#[cfg(feature = "nym")]
 fn parse_probe_target(raw: &str) -> Result<http::Uri, String> {
     let uri = raw
         .parse::<http::Uri>()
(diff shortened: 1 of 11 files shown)


zingo-cli/src/commands.rs
```diff
@@ -124,246 +130,155 @@ fn usage(command: &str, detail: impl std::fmt::Display) -> CommandError {
...
+    match lightclient.set_indexer_uri(uri.unwrap_or_default()).await {
+        Ok(()) => Ok("server set".to_string()),
+        Err(e) => Err(CommandError::NotYetTyped(format!(
+            "failed to set server: {e}"
+        ))),
+    }
+}
+
+/// Renders the wallet's judgment of an address into the `check_address` JSON.
+fn address_check_json(address_ref: Option<WalletAddressRef>) -> json::JsonValue {
```
zancas, line 191: Does this handle ironwood properly?

PR #2642, opened 2026-08-07
Show That Migration Operations Run Over The Mixnet

zancas:
A migration part's submission returned a bare txid. Nothing recorded which wire carried it, so the route was knowable only by re-deriving the session policy afterwards. That is the weakest form of a privacy guarantee. It cannot be audited, and it cannot be tested.

## What changes

Submissions now answer with a typed receipt. The receipt carries the txid and the route the submission traveled. A mixnet route names the drawn Broadcast Indexer and the tunnel endpoint. A clearnet route names the endpoint host.

The migration broadcast loop records each receipt route in the cross-session indexer history. The `network history` command then shows the wire each part used. The evidence never enters the wallet file. The persisted grammar is untouched, and no format version moves.

## What the tests pin

A new validation module pins the rule the evidence defends:

1. A ready session resolves the mixnet wire.
2. An unattached session refuses rather than resolving clearnet, because absence is not consent.
3. Every part the lifecycle broadcasts reaches the wire exactly once, through the one seam.
4. A clearnet route reads as clearnet, so the pass is not vacuous.

The rule holds for ironwood-to-ironwood parts exactly as it holds for pool-crossing parts. The route resolver reads the Mixnet Mode and the proxy address alone. No pool can reach that decision. This matters most for note splitting, whose self-send amounts and cadence sketch the schedule.

## Verification

Clippy and rustfmt are clean. All 175 migration tests pass, including the four new ones.

### DIFF
--- a/zingolib/src/lightclient/migrate.rs
+++ b/zingolib/src/lightclient/migrate.rs
@@ -1121,8 +1121,10 @@ impl LightClient {
                     wallet.save_required = true;
                 })
                 .ok_or(MigrationError::NoMigration)?;
+            let started = std::time::Instant::now();
             match client.submit(raw_tx, expiry_height).await {
-                Ok(_) => {
+                Ok(receipt) => {
+                    record_part_route(&self.indexer_history, &receipt.route, started, Ok(()));
                     wallet
                         .with_migration_state(|wallet, state| {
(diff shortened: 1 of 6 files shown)


dorianvp (approved):
utACK!

PR #2665, opened 2026-08-10
The slot tunnel crosses the route as a Shared-category type

zancas:
## What this does

ADR 0039's typed partition (#2663) left one surface uncategorized: the session slot's tunnel. Route resolution handed its SOCKS5 address around as a bare `String`. The types recorded nothing about its exposure.

`MixnetRoute::Mixnet` now carries a `SlotTunnel`. The type offers one accessor: a borrowing `addr`. That accessor is the Shared category's repeated-dial seam. Every slot-tunnel consumer reads the tunnel through it: an attached session's send pulls, the price fallback, the probe fan-out, and the migration transmissions.

## Scope note

The categorization lands at the route seam. The consumers extract the address for wire use, because `zingo-price` and the netutils dial helpers are transport-agnostic by design. `MixnetRoute::socks5_proxy` keeps its signature.

## Breaking change

`MixnetRoute::Mixnet`'s payload changes from `String` to `SlotTunnel`. The zingolib CHANGELOG records it, and ADR 0039's amendment records the slot tunnel's categorization.

## Verification

The zingolib suite passes with the nym feature. The zingo-cli suite passes (230 tests). `cargo hack check --each-feature` passes on both crates.


zancas:
Code review of head `fad081cd`, eight findings confirmed. Commit 6d33827 already resolves two of them: the probe's wrong `Unattached` refusal and the inlined resolver body. The six findings below remain.

**1. ADR 0039 amendment overstates what `SlotTunnel` enforces.**
`docs/adr/0039-…-shared-across-correspondents.md:10`. The amendment files `SlotTunnel` under the lease-type frame: declared at the acquisition site, enforced by type-check. `SlotTunnel` is a Clone-able String newtype. It has no `ExitUse` marker, no `exit_pool::Reservation`, and no acquisition site. Every consumer erases it to `String` on arrival (`send.rs:124`, `lightclient.rs:643`, `lightclient.rs:1092`, `migrate.rs:925`). The untyped bypasses (`mixnet_socks5_addr()`, `MixnetProxy::socks5_addr()`) stay public. An auditor who greps `Member<_, Shared>` misses the slot tunnel entirely. Either narrow the ADR text to what the newtype does, or wrap a real `Member<MixnetProxy, Shared>`.

**2. `SlotTunnel` needs `into_addr(self) -> String`.**
`zingolib/src/nym/route.rs:27`. The type offers only a borrowing `addr()`. All four consumers own the tunnel by value and re-allocate with `.to_string()` where the pre-PR code moved the `String` out for free. `update_current_price` discards the clone unread when the Price Source Pool supplies a pooled member (`lightclient.rs:672`). An `into_addr(self)` (the `Member<T, Exclusive>::dial(self)` shape) removes every clone.

**3. Four match arms hand-roll what `socks5_proxy()` provides.**
`zingolib/src/lightclient.rs:643`. Five copies of the Mixnet-arm extraction now exist (`route.rs:49` plus the four consumers), and the canonical accessor has zero production callers. A future guard added to `socks5_proxy()` misses every real call site. Route the consumers through the accessor, or drop it.

**4. The updated route test asserts a private-field struct literal.**
`zingolib/src/nym/route.rs:133`. The `socks5_proxy()` assertion on the next line already fails on `Err`, on `Clearnet`, and on a wrong address. The literal adds no coverage and breaks the test on any internal change to `SlotTunnel`. Deleting it restores the pre-PR one-line assertion.

**5. `SlotTunnel` is the seam to close the empty-address hole.**
`zingolib/src/nym/route.rs:92`. Pre-existing, not a regression: `MixnetProxy::attach` rejects unparseable addresses (`supervisor.rs:286`), but `drive_state` (`supervisor.rs:519`) accepts the child's `SOCKS5_ADDR=` line with no parse. A broken bundled nym-proxy that emits a bare `SOCKS5_ADDR=` publishes Ready with an empty string, and every surface treats the session as routable until an opaque SOCKS5 connect error. A parsing `SlotTunnel` constructor closes both producers at once.

**6. Stale comment in the send path.**
`zingolib/src/lightclient/send.rs:901`. The `transmit_transactions` route comment still names the old payload shape `Mixnet(addr)`. It is now the only in-tree reference to the pre-PR `String` payload.

PR #2684, opened 2026-08-12
The survey bounds its fan-out, its refusal names its causes, and a teardown harness rides the workbench

zancas:
This PR stacks on #2683 and targets its branch; it merges after #2683 or retargets to dev when #2683 lands.

## The defect

The sweep surveyed all 17 candidates concurrently. Every tunnel rides the one sweep exit's packet pipeline. An adjacent harness measured one GetLatestBlock round trip through that pipe at about 5 seconds, with roughly every other fresh SOCKS5 tunnel reset. Seventeen concurrent TLS handshakes through one mixnet-throughput pipe blow every probe's 20-second budget together. That is the observed all-or-nothing signature: 17/17 on a fast exit, 0/17 on an ordinary one. The indexers answered clearnet grpcurl instantly throughout.

## The fix

The survey opens at most `SURVEY_TUNNEL_WIDTH` (4) tunnels at a time. Two live runs against the full census answered 17 of 17, chose the pinned server, and opened the Sync Session, where the unbounded fan-out answered 0 of 17 minutes earlier on the same machine.

## The self-diagnosis

The sweep stops discarding why probes fail. `SurveyResult` carries the diary's `FailureKind`, and `EmptyCohort` tallies causes into its rendering: "no live indexer: 0 of 17 answered, none within the cohort (17 timeout)". A saturated transport now reads differently from dead indexers in the refusal itself. A falsifier test pins the tally.

## The harness

`tools/workbench/src/bin/sweep-teardown.rs` ports the throwaway that found this: standalone nym-proxy, ncat bridge, grpcurl probes — an independent client per the workbench scripting rule. It requires grpcurl and ncat on PATH.

### DIFF
--- a/tools/workbench/src/bin/sweep-teardown.rs
+++ b/tools/workbench/src/bin/sweep-teardown.rs
@@ -33,8 +33,10 @@ const RPC: &str = "cash.z.wallet.sdk.rpc.CompactTxStreamer/GetLatestBlock";
 /// How long the proxy may bootstrap before the harness gives up.
 const BOOTSTRAP_BUDGET: Duration = Duration::from_secs(150);
 
-/// How long one bridged probe may take, matching the sweep's own leg budget.
-const PROBE_BUDGET: Duration = Duration::from_secs(30);
+/// How long one bridged probe may take: a hand-copied mirror of the
+/// sweep's `PROBE_LEG_TIMEOUT` (this dependency-free workspace cannot
+/// import it), retuned together with that constant.
+const PROBE_BUDGET: Duration = Duration::from_secs(20);
 
 /// How many rounds ride the same exit without `--rounds`.
 const DEFAULT_ROUNDS: usize = 2;
--- a/tools/workbench/src/bin/sweep-teardown.rs
+++ b/tools/workbench/src/bin/sweep-teardown.rs
@@ -1,0 +1,172 @@
+//! Tease apart a failing Server-Selection Sweep with an independent client.
+//!
+//! The harness spawns the bundled `nym-proxy` standalone (it draws its own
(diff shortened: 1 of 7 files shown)


zancas:
Absorbed into #2695. Every commit of `survey_width` is now an ancestor of
`first_healthy_sweep`, merged there in 5e62e679e.

The two branches had converged on the same sweep work by different routes:
both bounded the survey's fan-out, offered an early verdict, and aborted the
health sweep on `network off`. The merge keeps one copy of each and takes
what only this PR carried:

- The refusal diagnosis. `SurveyResult` carries the diary's `FailureKind`,
  and `EmptyCohort` renders the per-kind `RefusalTally`, so a saturated
  transport reads differently from dead indexers.
- The `GetLatestBlock` tip probe, as `Socks5Indexer::get_latest_block`. The
  attach readiness round trip and live-indexer discovery use it.
- The `sweep-teardown` workbench harness.

The runner in #2695 stays the first-healthy design, which the 2026-08-12
ruling ratified over the wave-and-median gate this PR was built against.

One deliberate departure from this PR: the survey itself keeps
`GetLightdInfo` rather than the tip probe. A tip carries no chain identity,
and the first-healthy verdict binds without a median gate, so the chain a
candidate names is the only guard against a mainnet wallet binding a testnet
indexer.

Verified on the merge: clippy clean, 232 of 232 zingo-cli and 497 of 497
zingolib unit tests pass with `--features nym`, and zingo-netutils is clean
in its standalone workspace.

Note for #2683, which this PR stacked on: four of its commits are still not
in #2695 — the `--sync-server` rename, the ADR 0034 rulings, and its dev
merge — so it stays open on its own.

PR #2697, opened 2026-08-13
Continuous sync

Oscar-Pepper:


zingolib/src/config.rs
```diff
@@ -504,8 +504,9 @@ impl Default for ClientConfigBuilder {
                 chain_height: 1,
                 wallet_settings: WalletSettings {
                     sync_config: SyncConfig {
-                        transparent_address_discovery: TransparentAddressDiscovery::minimal(),
+                        transparent_address_discovery: TransparentAddressDiscovery::default(),
                         performance_level: pepper_sync::config::PerformanceLevel::High,
+                        shutdown_on_completion: false,
```
juanky201271, line 509: [blocker] With `false` as default, `sync_and_await` / `await_sync` never return. `migrate.rs` calls `sync_and_await` internally (L1689, L2043, L2219), so immediate migration / drain hangs forever for any wallet in continuous mode. Suggest forcing bounded mode in `sync_and_await` regardless of config, or making the mode a `sync()` argument instead of wallet config.

pepper-sync/src/sync.rs
```diff
@@ -442,214 +457,333 @@ where
...
+                    if check_for_new_blocks && scanner.is_verified() {
+                        continuous_sync_interval.reset();
+                        check_for_new_blocks = false;
+                        first_verification_complete = true;
+                        continue 'continuous_sync;
+                    }
+
+                    scanner.update(&mut *wallet.write().await, nullifier_map_limit_exceeded).await?;
+
+                    if matches!(scanner.state, ScannerState::Complete) && config.shutdown_on_completion {
```
juanky201271, line 755: [blocker] Migration part witnesses are only captured when sync returns (`refresh_part_witnesses` in `await_sync`, and consumers doing it on `PollReport::Ready`). In continuous mode sync never returns, so the checkpoint window is missed and a scheduled migration can't complete. Needs a hook inside the loop.

pepper-sync/src/config.rs
```diff
@@ -86,12 +86,21 @@ pub struct SyncConfig {
...
+    /// If set, sync will still check for any newly mined blocks during scanning. But when the wallet is completely
+    /// up-to-date with the latest chain, sync will shutdown.
+    pub shutdown_on_completion: bool,
 }
 
 #[cfg(feature = "wallet_essentials")]
 impl SyncConfig {
     fn serialized_version() -> u8 {
-        1
+        2
```
juanky201271, line 103: [must-fix] Extra byte in `SyncConfig` without bumping the `LightWallet` version, and the old `SyncConfig::read` has no max-version check. A wallet saved with this and opened by an older build misreads everything after it (min_confirmations, ...). Please bump the wallet version so older builds fail cleanly.

pepper-sync/src/sync.rs
```diff
@@ -997,8 +1114,13 @@ async fn publish_sync_status<W>(wallet: &W, progress: &watch::Sender<Option<Sync
 where
     W: SyncWallet + SyncBlocks,
 {
-    if let Ok(status) = sync_status(wallet).await {
-        let _ = progress.send(Some(status));
+    match sync_status(wallet).await {
+        Ok(status) => {
+            progress.send(Some(status)).unwrap();
```
juanky201271, line 1119: [must-fix] This used to be ignored. Now `unwrap()` / `panic!` kill the sync task, and `poll_sync` does `expect("task panicked")`, so the panic reaches the consumer (FFI in mobile/pc). `send` also fails if the `LightClient` is dropped while sync is running. Please keep ignoring the error or return it.

pepper-sync/src/sync.rs
Oscar-Pepper: is it not possible to set the config according to the sync settings you want for that particular sync? i.e. during migration syncs set the config to shutdown on completion but then when we are syncing "normally" we set the sync config to not shutdown? im not sure why this is different to changing the boolean we pass into the method if we added one. there may be many more of these settings as we develop the sync engine and sync() / sync_and_wait() currently has no parameters, is it not best to keep it this way?

zingolib/src/config.rs
Oscar-Pepper: this is related to my answer above. if we force sync_and_await() to shutdown on completion we are not able to provide a helper to the user that starts sync and polls for errors which i think is very valuable. I would prefer the consumer sets the config correctly for before they start sync unless there is an issue with this which means its not a good idea or cant be done?

pepper-sync/src/sync.rs
Oscar-Pepper: yes this should not panic, sorry. however, i had a difficult time debugging because it was silently erroring. I think i need to update this to return an error so the consumer is aware is has failed but doesnt crash

juanky201271:
Reviewed from the zingo-pc side. Nothing blocking for us — the note below is for whoever picks the migration up later.

**`migrate_to_ironwood` cannot run under continuous sync.** It calls `sync_and_await` (migrate.rs:2043), and so does `await_migration_confirmations` in its poll loop (migrate.rs:2219). `await_sync` waits for the sync task to exit, which with `shutdown_on_completion: false` it never does. The CLI's fix — `shutdown_on_completion = filled_template.waitsync` — isn't available to a library consumer, since the flag lives in the wallet's stored settings rather than in the call. The immediate path already has `migrate_immediately_presynced` for consumers that own the sync lifecycle; the split-and-confirm path has no equivalent. Not urgent for us: that path is wired but unreachable in zingo-pc today.

juanky201271 (dismissed):

juanky201271:
Following my note above, the merge of `dev` brought the piece that would solve it: `subscribe_sync_status` (#2773).

`wait_for_sync` has the same shape as `await_sync` — it waits for the engine to return, which under continuous sync it does not. The subscription does not: a caller can watch the published status and resolve when the wallet is at the tip. That is what `migrate_to_ironwood` and `await_migration_confirmations` need in place of `sync_and_await`, and it keeps the split-and-confirm path working whether or not the consumer runs the engine continuously.

Still not blocking for zingo-pc.

juanky201271 (approved):
nice job.

PR #2713, opened 2026-08-19
feat!: bump the Zcash stack to zcash_primitives 0.30

zancas:
The workspace compiled two copies of zcash_primitives. Cargo keys a package by source rather than by version, so the git dependency on zcash_pool_migration dragged in git-sourced copies of zcash_primitives, zcash_protocol and zcash_transparent that no version bump could unify with the registry stack. This moves the whole stack to 0.30 and sources the migration crate from the registry at 0.1.0-rc.7, leaving one copy of each crate and no librustzcash git source in the lockfile.

Collapsing the divide retires the insulation it forced. The migration schedule passed heights across the version boundary as u32; with one shared zcash_protocol it passes BlockHeight directly. ADR 0020's branch-move tripwire watched a git revision that no longer exists, so it now pins the adjudicated release instead, preserving the loud-movement property the ADR asked of it.

Upstream moved ZIP 318's ratified constants into zcash_protocol::zip318 and changed one of them: ANCHOR_AGE_CAP falls from 16 boundaries to 4. The conformance tripwire adopts that value deliberately, as ADR 0020 requires. MigrationParams keeps version 2, because the anchor cap never fed the consent hash. MIGRATION_MAX_DENOMINATION_ZEC and RESIDUAL_MIGRATION_MIN become DENOM_CAP and MAX_RESIDUAL_VALUE, both Zatoshis rather than a count of whole ZEC.

The client backend's own surface moved further. WalletWrite now requires OutputLockStore, so LightWallet implements it over an in-memory lock set held beside the pending proposal, which expires with the process exactly as ADR 0006 rules the proposal slot does. WalletRead gained get_wallet_recover_until, WalletWrite gained prune_scan_queue_below, and WalletCommitmentTrees gained readers for the Sapling and Orchard subtree roots, answered through the tree's public root query. InputSource gained anchor_computable, answered from the shard-scan machinery that already decides whether a witness can be built. Five selection methods gained a lock filter, which this accepts without consulting: zingolib requests no lock at any call site, so its lock set is empty in every flow it drives, and honouring the filter belongs with adopting the locking feature.

The transaction builder replaced orchard_pool_bundle_type with paired orchard and ironwood padding, and three call sites grew an argument that this declines.

The seeded draw_anchor_boundary golden vectors were captured under the old anchor cap and are expected to diverge; their adjudication against the divergence ledger is not part of this commit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -872,15 +872,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "equihash"
-version = "0.3.0"
-source = "git+https://github.com/zcash/librustzcash.git?rev=e12f1d0ff7be5e5bfd2e4bcbb8d9a863a405f031#e12f1d0ff7be5e5bfd2e4bcbb8d9a863a405f031"
-dependencies = [
- "blake2b_simd",
- "corez",
-]
-
-[[package]]
 name = "equivalent"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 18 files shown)


dorianvp (approved):
utACK

PR #2716, opened 2026-08-20
feat(workbench): compare two builds' sync rate through a real CLI session

zancas:
## What this adds

`makers sync-ab [cli] <arm> <arm> --birthday <height>` compares two builds'
sync rate. It reports outputs per second, because scan cost tracks outputs
rather than blocks.

An arm is a commit and a build kind. Write it as `<commit>[:clearnet]`. The
default kind builds the CLI with the mixnet and bundles the proxy. The
`:clearnet` kind builds without the mixnet and bundles nothing.

## Why it drives the CLI

An in-process harness measures the sync engine alone. It says nothing about
a session where four proven exits bootstrap beside the scan. So each run
spawns a real session. It uses `makers run-cli --online` where that task
exists. It builds and launches `zingo-cli` directly where the task does not
exist, which reaches commits from before the task landed.

Sync scans over clearnet in both build kinds. The sync engine takes a plain
gRPC indexer and never a SOCKS5 route. So one commit against itself in the
two kinds measures what the mixnet's presence costs a clearnet sync.

## Where the numbers come from

The tool grafts a span into each arm's worktree. It does not require the arm
to carry one. The span times the sync call and logs the outputs scanned.

A fixed patch cannot do this. The call's argument list grew across the range
this compares. So the graft anchors on the spawn that every era shares, walks
to the matching brace, and surrounds the text between them. It never reads
the call.

The graft refuses in two cases. It refuses when the anchor does not appear
exactly once. It refuses when the anchored block does not call the sync
engine. A step that would instrument the wrong work therefore fails loudly.

Boot is the one wall-clock number. It spans the spawn to the span's opening.
It carries the build and the bootstrap, and it is reported separately so a
launch cost cannot enter the scan rate.

## What is held fixed

A benchmark that varies more than one thing measures nothing. This holds the
indexer, the seed, the window, the performance level, and the release profile
constant. Each arm gets its own worktree, so neither arm meets the other's
build artifacts. A warm-up per arm pays for the build and is discarded. The
measured rounds alternate, so warm-up and machine load fall on both arms
equally.

The window is the caller's. `--birthday` is required, and every arm receives
the one value. Nothing here asks a moving tip where to start. The report
prints the output counts and warns when the arms scanned different work.

## Measurements taken with it

The tool found a 23.8% sync-rate gain between `zingolib_beta_ironwood` and
`dev`, over a 10,000-block window:

```
zingolib_bet  mean 4480 outputs/s  stdev 228  [4217, 4601, 4622]
0770cb68e     mean 5546 outputs/s  stdev 242  [5274, 5736, 5626]
outputs scanned: 70608 to 70615 (0.01% drift)
difference: +23.8%
boot: 0.1s against 43.1s
```

A 1,000-block window agreed on the direction at +25.7%. It could not
separate the arms, because one arm's spread exceeded the gap. It also saved
only 8% of wall time, because builds dominate. Prefer the larger window.

### Where the gain sits

Points across the 78 first-parent commits, each candidate measured against
the tag in the same interleaved run:

```
  idx      gap     reference      candidate   commit
  ---   ------   -----------   ------------   -------------------------------------
   10    +3.8%      4050±391       4205±215   8778fc076  #2579 zingo-cli-over-nym
   20    -1.8%      4323±142       4246±242   14fb412d7  #2618 selfsend_canonical_value
   30    -3.0%      4227±216       4101±518   451db7abd  #2627 network_observer
   38    +6.9%      4196±437       4485±469   da283eaba  #2647 cargo_update
   48    +5.2%      4496±287       4732±349   f422c80c2  #2657 no_magic_numbers
   58        —             —              —   d08848907  #2670 sync_progress_watch (failed)
   68    +9.0%      4373±387       4769±584   fcc1c0fd9  #2689 sync_progress_saturation
   69    +9.8%      4213±345       4625±255   7a2cf0041  #2695 first_healthy_sweep
   70   +17.8%      4329±402       5101±338   4d9f2ece8  #2677 transparent_gap_limit
   71   +14.6%      4289±605       4916±386   3eff192ea  #2707 sweep_assigns_every_indexer
   73   +13.2%      4297±482       4862±373   dff7f8eac  #2710 cli_cleanup
   77   +23.8%      4480±228       5546±242   0770cb68e  dev (head)
```

The same points drawn at one sigma, on a shared scale of outputs per second:

```
           3500      3900     4300     4700      5100      5500
   10  ref     [--------o--------]
      cand             [----o----]                          +3.8%  overlap   #2579 zingo-cli-over-nym

   20  ref                 [---o--]
      cand             [-----o----]                         -1.8%  overlap   #2618 selfsend_canonical_value

   30  ref             [----o----]
      cand   [-----------o------------]                     -3.0%  overlap   #2627 network_observer

   38  ref       [----------o---------]
      cand             [----------o-----------]             +6.9%  overlap   #2647 cargo_update

   48  ref                  [------o-----]
      cand                      [-------o--------]          +5.2%  overlap   #2657 no_magic_numbers

   68  ref             [--------o--------]
      cand                 [-------------o-------------]    +9.1%  overlap   #2689 sync_progress_saturation

   69  ref          [-------o-------]
      cand                      [-----o-----]               +9.8%  overlap   #2695 first_healthy_sweep

   70  ref           [---------o--------]
      cand                               [-------o-------] +17.8%  SEPARATE  #2677 transparent_gap_limit

   71  ref     [--------------o-------------]
      cand                         [---------o--------]    +14.6%  overlap   #2707 sweep_assigns_every_indexer

   73  ref        [-----------o----------]
      cand                        [--------o--------]      +13.1%  overlap   #2710 cli_cleanup

   77  ref                   [----o-----]
      cand                                    [-----o----] +23.8%  SEPARATE  dev (head)
```

Two rows separate at one sigma. One is the head. The other is index 70,
whose whole diff is 23 lines of CLI argument parsing.

Note how far the reference bar itself moves between rows. The same tag reads
from 4,050 to 4,496 across these runs. That is the machine, not the code, and
it is why each point re-measures the reference beside its candidate.

Every point through index 68 sits inside its own spread. The gain therefore
lives in the last nine merges.

### A bisect over those nine gave a false answer

The bisect converged on index 70. That merge cannot be the cause. It adds one
CLI command, changes no default, and touches nothing the sync engine runs.
The whole diff between index 69 and index 70 is 23 lines in
`zingo-cli/src/commands.rs`, a CHANGELOG entry, and tests.

The table shows why the convergence is an artifact. Indices 71 and 73 read
lower than index 70, which a step change cannot do. The readings wander
upward inside overlapping spreads, and the cutoff crossed at 70 because that
run came in high.

This is worth stating in the tool's own PR. The tool measures honestly, and
the numbers still do not support a single-commit cause at this precision.
Separating those nine merges needs more rounds per point.

### One failure worth reading

Index 58 produced no reading:

```
Server-Selection Sweep: 0 of 17 candidates answered; judging the live cohort...
Server-Selection Sweep: no sync indexer selected: pinned server
  'https://zec.rocks:443/' is not live over the mixnet. This Sync Session
  does not open; the mixnet posture stands, and send and price-fetch continue.
sync-ab: budget of 900s exceeded
```

In a mixnet build, sync does not open unless the Server-Selection Sweep
reaches the pinned server over the mixnet. So the mixnet gates whether the
scan starts, and does not merely run beside it.

## Tests

Five tests cover the graft. They pin both eras' spawn shapes, both refusals,
the arm suffix, and a truncated reading.


zancas (commented):
Self Approved.

PR #2719, opened 2026-08-21
Make the conduit dial a proper guard

zancas:
## What this changes

`ConduitDial` was a guard in shape only. Its one method returned the
`SocketAddr` it wrapped. `zingo-price` read that address, formatted a
`socks5h://` proxy URL, and built a `reqwest` client around it. The type said a
price fetch is mixnet-only. The body discharged that claim into a string.

This moves the HTTP leg below the seam. `zingo-netutils` gains a `socks5-fetch`
feature and a `socks5_fetch` module. The module carries `Socks5FetchError`, the
`zingo-net-diag` classification table, and `ConduitDial::fetch_text`. A caller
reaches the wire by holding the guard. Holding the guard keeps the conduit's
in-flight count correct for as long as the work runs.

`zingo-price` no longer names a network type. It loses `reqwest` and `rustls`
from its dependency set. It also loses `fetch_over`, the proxy-URL formatting,
the crypto-provider install, and the classification table. What remains is the
price data model and the per-source parsers.

## Why this matters

ADR 0046 states that the wallet cannot read an address out of a
`MixnetConduit`. ADR 0047 records that the earlier move relocated the types
without making them opaque. This change makes one caller match the record. It is
the smallest complete proof of the pattern.

The second commit records a related ruling in `zingolib/CONTEXT.md`. A conduit
label carries `Display` and nothing else. A caller that must compare two labels
has shown that its responsibility belongs below the seam.

## TLS roots

The `reqwest` feature string moved across verbatim rather than being
re-derived. It selects `rustls-tls-webpki-roots-no-provider`. The fetch
therefore verifies against the compiled-in Mozilla bundle, and it installs the
`aws-lc-rs` provider itself. This preserves the behaviour the price fetch
already had. It makes no new ruling about trust roots.

Whether this fetch should use platform roots instead is a live question. It
belongs to ADR 0021, not to this PR. zingo-mobile PR #1276 hands the app
`Context` to a real `rustls-platform-verifier` inside the proxy shim, so the
mobile side now verifies against Android's own store. ADR 0021 rejected that
shape for reasons that shipped code no longer supports, and it needs a
superseding record.

## Breaking change

`PriceError::RequestFailed` carries a boxed `Socks5FetchError` where it carried
a `reqwest::Error`. The `failure` field does not change. The box keeps the enum
at its previous size, which `result_large_err` requires. Both crates carry
CHANGELOG entries.

## Not done here

`ConduitDial::socks5` and `MixnetConduit::over` stay public. Fourteen `zingolib`
sites still take the raw address: `route.rs:86,120`, `speed.rs:108,329`,
`mode.rs:196`, `supervisor.rs:1421`, `lightclient/mixnet.rs:988,1063,1115,1146`,
`select.rs:193`, `send.rs:817`, and `transmission_route.rs:105,130`. Sealing
both methods requires collapsing the desktop and mobile fork at
`lightclient/mixnet.rs:1135`, which ADR 0049 covers.

## Tests added

`docs/agents/net-diag-design.md` requires that every classifier be tested with
fabricated error chains covering every stage. Nothing had ever supplied that for
the reqwest table. This adds one case per stage the table can produce, plus the
assertion that the set of stages reached is the whole set.

Two further cases pin where the implementation has drifted from the design's
rows. The TLS arm fires on chain text alone, where the design conjoins it with
`is_connect()`. That arm also outranks `is_status()`, so a status error whose
chain names a certificate never reaches `RemoteHttp`. Both drifts are now
visible failures if anyone changes them. Both remain open questions for the
design.

## Verification

`cargo check` and `cargo clippy -- -D warnings` pass for `zingo-price`
all-targets, `zingo-price --no-default-features`, netutils under default,
`socks5-transmit`, and `socks5-fetch` all-targets, and `zingolib` with and
without `nym`. Tests are not run here.


zingo-netutils/src/socks5_fetch.rs
```diff
@@ -0,0 +1,335 @@
...
+    if signals.is_decode_or_body {
+        return NetOpStage::PayloadDecode;
+    }
+    if signals.is_connect {
+        return NetOpStage::RemoteConnect;
+    }
+    NetOpStage::TunnelTransport
+}
+
+fn classify_request(
```
dorianvp: This should probably be named something like `ClassifyError`

tools/workbench/src/bin/feature-census.rs
```diff
@@ -0,0 +1,551 @@
...
+    // transmit legs, which is where this crate's own feature declarations are.
+    "zingo-netutils",
+    "socks5-transmit,socks5-fetch,testutils",
+)];
+
+/// One dependency feature the census can probe.
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+struct Candidate {
+    /// The crate directory whose manifest declares it, or `.` for the root.
+    krate: String,
```
dorianvp:
```suggestion
    crate_dir: String,
```

dorianvp (dismissed):
Approved but left some comments

zingo-netutils/src/socks5_fetch.rs
zancas: Fixed.

PR #2741, opened 2026-08-24
test: speed up send_shield_cycle from 58s to 44s

zancas:
Profiling showed that zebra's getblocktemplate dominates this test. Each template builds a fresh shielded coinbase with a proof. An empty template costs about 2.2 seconds. A transaction-bearing template costs about 4.5 seconds.

This branch lands three speed-ups and the instrumentation that found them.

1. **Infrastructure pin bump.** The pin moves to 89cf104a2, the merge of zingolabs/infrastructure#284. `Zebrad::generate_blocks` now parks a long-poll getblocktemplate after each mined block. Zebrad precomputes the next coinbase during harness work. Mining drops from 28.3 to 23.2 seconds mean.
2. **Converge before the confirmation sync.** `follow_proposal` raced the indexer with a bare `sync_and_await`. The wasted confirm round cost about nine seconds. `sync_client_to_tip` barriers on indexer convergence first.
3. **Proving-key warmup.** The PostNu6_3 Orchard proving key now builds during environment setup, in parallel with the network launch. This removes over a second from the measured send path.
4. **Phase timers.** A `timed` helper logs each phase's label and elapsed seconds across the whole call graph. These timers produced the measurements above.

The measured mean run time falls from 58.4 seconds to about 44 seconds. A run through the merged pin confirmed 45.6 seconds.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3268,7 +3268,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "32497e9a4c7b38532efcdebeef879707aa9f794296a4f0244f6f69e9bc8574bd"
 dependencies = [
  "fastrand",
- "getrandom 0.4.3",
+ "getrandom 0.3.4",
  "once_cell",
  "rustix",
  "windows-sys 0.61.2",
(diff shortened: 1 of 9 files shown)


zancas:
Test only PR, tested and benchmarked.

PR #2747, opened 2026-08-27
fix!: restore clearnet price fetch when Mixnet Mode is switched off

dorianvp:
This PR restores the price fetch when nym is switched off, which was not working.

### DIFF
--- a/docs/adr/0011-nym-mixnet-transmission.md
+++ b/docs/adr/0011-nym-mixnet-transmission.md
@@ -490,3 +490,28 @@ workspace and generates its own bindings (zingo-mobile PR #1251). The
 UniFFI, and the runtime-boundary generalization remains in force; only
 the shim's hosting repository changes. This workspace keeps the desktop
 `nym-proxy` binary and its `bundle-nym-proxy` workbench tool.
+
+## Amendment (2026-08-26): the switched-off consent covers the price fetch
+
+The 2026-07-28 ruling above, that the switched-off consent covers
+Transmission and never the price fetch, is reversed. The deliberate
+SwitchedOff now consents to a clearnet price fetch exactly as it consents
+to a clearnet send. The mobile sessions supplied the deciding evidence. A
+user who declines the mixnet starts every session SwitchedOff, and the
+price display never works for that user, in any session, by design. The
+refusal was meant to prevent an unconsented leak, but the state it fires
+in is the one state the user reached by an explicit choice. The same
+session's sends already travel clearnet to an indexer that knows the
+wallet's address set, and a price API learns less than that indexer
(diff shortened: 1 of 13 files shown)


zancas (approved):

zancas:
The review finished with ten verified findings. I'll report them through the findings UI, then summarize.

● Code review(high · 10 findings)
zingolib/src/lightclient/mixnet.rs
  ● 1159 [privacy]         clearnet_price_race collects all nine RACED_SOURCES into an unbounded FuturesUnordered, so the IP-disclosing route fans out nine-wide while the mixnet route is capped at SURVEY_WAVE_WIDTH (4).
  ● 1223 [test-coverage]   switched_off_mode_consents_to_a_clearnet_fetch is a plain #[tokio::test] firing nine live HTTPS requests, and its empty Err(PriceError) arm accepts exactly the failure a routing regression produces.
  ● 1107 [correctness]     The ConduitDial guard is taken before the acquirer branch that never uses it (then shadowed by a String named dial), so a spawned-session price race pins the standing conduit's in_flight count for a race travelling a freshly drawn exit.
  ● 1154 [duplication]     clearnet_price_race restates PriceRun's declared policy and re-implements run_wave's drive-and-settle loop inline, leaving zingo-price's first_quote doc ('the racing lives in one place') false.
  ● 1093 [doc-comments]    Six diff-authored item doc-comments are two sentences, violating the governing hard rule that every /// is exactly one sentence with no exceptions.
  ● 1104 [imports]         All five diff-authored PriceFetchRoute use sites are inline full paths though both files already import siblings from those exact modules, violating the partial-imports rule.
  ● 1165 [efficiency]      The clearnet legs reuse REQUEST_TIMEOUT (20s) and CONNECT_TIMEOUT (10s), bounds sized against the mobile watchdog for a tunnel-traversing fetch, so a switched-off fetch on a dead or captive-portal network holds update_current_price 10-20 seconds.
zingo-cli/src/commands.rs
  ● 2151 [stale-docs]      Six surviving statements of the reversed mixnet-only price rule contradict the PR's behavior, led by the CurrentPrice long_about in the very file the PR edits.
docs/agents/net-diag-design.md
  ●  217 [breaking-change] The design doc says via_socks5 on the price-fetch success payload is consumed by zingo-mobile and must not be disturbed, and the PR removes that field without updating the doc or naming the mobile migration.
zingo-netutils/src/socks5_fetch.rs
  ●  157 [api-surface]     Dropping #[cfg(feature = "testutils")] makes fetch_text_untunneled pub to every socks5-fetch consumer, so fail-closed transitional states are protected only by one match in update_current_price rather than the compiler.

PR #2749, opened 2026-08-28
fix(zingolib): constrain wallet reader & recovery, and add validation method

dorianvp:


zancas (dismissed):
I particularly like the CountingReader pattern.

(re) running tests locally.

juanky201271 (approved):

PR #2750, opened 2026-08-28
fix(`pepper-sync`): retain ZIP 318 anchor checkpoints

dorianvp:
Sends of exactly 1, 2, 5 × 10^n ZEC failed with `AnchorNotFound`: the canonical-crossing path anchors to a ZIP 318 grid boundary 147–290 blocks behind the tip, but pepper-sync retained only the newest 100 checkpoints, and the `anchor_computable` guard only checked scan completeness, so the doomed attempt was not rejected until after proposal acceptance.

- zingolib: `anchor_is_computable` now also requires a checkpoint at the anchor height, so a canonical attempt at a pruned boundary degrades to the ordinary send path.
- pepper-sync: create and pin grid-boundary checkpoints in all three pool trees during scanning (window of `ANCHOR_AGE_CAP + 2` intervals from the Ironwood activation), re-derive the pins at sync start, exempt pinned checkpoints from the newest-100 serialization cap, and iterate the full checkpoint count when truncating to the next checkpoint.

Wallets with already-pruned boundaries send via the ordinary path immediately; canonical crossings resume once sync passes the next grid boundary.

### DIFF
--- a/pepper-sync/src/wallet/serialization.rs
+++ b/pepper-sync/src/wallet/serialization.rs
@@ -1356,6 +1356,8 @@ impl ShardTrees {
 
 #[cfg(test)]
 mod tests {
+    use crate::witness::ANCHOR_RETENTION_INTERVALS;
+
     use super::*;
 
     // Helper: build a minimal v3 SyncState byte blob (no ironwood_shard_ranges).
@@ -1465,9 +1467,9 @@ mod tests {
     }
 
     /// The checkpoint set of a synced wallet decomposes into exactly two parts:
-    /// [`MAX_REORG_ALLOWANCE`] rolling checkpoints, which serve ordinary reorg handling and
-    /// near-tip spends, plus [`MAX_BOUNDARY_CHECKPOINTS`] pinned ZIP 318 grid boundaries,
-    /// which serve pool crossings. Their sum is [`MAX_SHARDTREE_CHECKPOINTS`].
+    /// [`MAX_SHARDTREE_CHECKPOINTS`] rolling checkpoints, which serve ordinary reorg handling and
+    /// near-tip spends, plus [`ANCHOR_RETENTION_INTERVALS`] pinned ZIP 318 grid boundaries,
+    /// which serve pool crossings.
     ///
     /// The two parts are disjoint and independently bounded: pinning a boundary must not
     /// consume a rolling slot (that would shrink the reorg window), and the rolling budget must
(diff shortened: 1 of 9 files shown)


pepper-sync/src/wallet/serialization.rs
```diff
@@ -1341,9 +1341,21 @@ impl ShardTrees {
                 Ok(())
             })
             .expect("Infallible");
-        if checkpoints.len() > MAX_REORG_ALLOWANCE as usize {
```
Oscar-Pepper, line 1344: this is good logic it just needs to use MAX_SHARDTREE_CHECKPOINTS instead

Oscar-Pepper (changes requested):
we also need to check spend migrations only use boundary anchors as only these are retained. im not sure what happens in the chain tip i.e. top 100 blocks etc.

pepper-sync/src/wallet/serialization.rs
```diff
@@ -1467,6 +1467,112 @@ mod tests {
...
+                Ok(())
+            })
+            .expect("infallible");
+
+        assert!(
+            pinned.iter().all(|height| height % INTERVAL == 0),
+            "every pinned checkpoint must be a grid boundary, got {pinned:?}"
+        );
+        assert_eq!(rolling.last().copied(), Some(TIP));
+        // TODO: this fails with:
```
dorianvp: @Oscar-Pepper take a look at this. I think this means that the extra 6 checkpoints we added are not really used, and that whatever the shardstore has, has it's own separate storage to take this into account...

Oscar-Pepper: i have pushed an adjustment to account for this. running tests locally now.

pepper-sync/src/wallet/serialization.rs
```diff
@@ -1603,98 +1593,81 @@ mod tests {
...
+        }
+        assert_window(sapling_store, oldest_kept);
+        assert_window(orchard_store, oldest_kept);
     }
 
-    /// A pinned anchor checkpoint must survive the write cap. The pinned set itself is not
-    /// persisted, being re-derived at the start of every sync.
+    /// A pinned anchor checkpoint survives serialization even once it has aged out of the
+    /// rolling window. The pinned set itself is not persisted, being re-derived at the start of
+    /// every sync.
```
Oscar-Pepper, line 1630: this comment is thought provoking. how does the shard tree rebuild the retained list on serialization? there needs to be a test on a higher layer that tests a roundtrip and asserts the retained list is restored also

Oscar-Pepper (changes requested):

pepper-sync/src/wallet/serialization.rs
Oscar-Pepper: ok it rebuilds it on the first sync after reload from file before pruning. i have added a test to prove that shard tree serialization is stable on the highest level and also checks for all three stores

Oscar-Pepper (dismissed):
great work!

Oscar-Pepper (approved):

PR #2757, opened 2026-09-06
fix(`pepper-sync`): add height-0 checkpoint when a loaded shard tree is empty

dorianvp:
This PR solves an issue where wallets with checkpoint-less shard trees abort at launch.


dorianvp:
Superseded by #2758.

PR #2759, opened 2026-09-07
fix(propose): size send-all with propose_send_max_transfer

dorianvp:


juanky201271:
The `zennies_for_zingo` path in `create_send_all_proposal` corrects the recipient amount once and retries. If that corrected amount shifts input selection and the fee moves again, the second attempt just fails — no third correction, where the loop this PR removes kept going until it converged.

Worth a comment on why one correction always converges, or a test that forces the fee to move twice.

juanky201271 (approved):

PR #2761, opened 2026-09-10
test(mock-indexer): enforce consensus checks on submitted transactions

dorianvp:
This PR expands the mock indexer to do enforce all the relevant consensus checks for zingolib.


zingolib/src/wallet/keys.rs
```diff
@@ -459,6 +459,41 @@ impl LightWallet {
             .map(|id| id.address_index())
     }
 
+    pub(crate) fn truncate_failed_refund_addresses(&mut self) {
```
juanky201271, line 462: **This isn't scoped to an account, so a failed send in one account can delete another account's refund addresses.**

The candidate is the highest refund index across *all* accounts, and `truncate_refund_addresses` then removes refund addresses wallet-wide. Example with two accounts:

1. Account 0 has a confirmed TEX send that used refund index 0.
2. Account 1's first TEX send (also refund index 0) fails to transmit.
3. Both addresses have index 0. `transparent_addresses()` is a `BTreeMap`, so `max_by_key` returns the last one, which is account 1's. It's paid and not live, so `below` is `None` and `truncate_refund_addresses(None)` removes every refund address, including account 0's.

Account 0's next TEX send then derives index 0 again and reuses an ephemeral address that's already on chain. That breaks ZIP 320's one-time-use rule and links the two payments. The previous logic in `send()` only truncated back to the pre-send highest index, so it never removed addresses that already existed.

Suggestion: run this per `account_id` (filter on `id.account_id()` here and in the truncation), and add a two-account test for the case above.

zingolib/src/testutils/mock_indexer.rs
```diff
@@ -210,20 +792,162 @@ impl MockChain {
...
+            return Ok(Some(Zatoshis::ZERO));
+        };
+        if bundle.is_coinbase() {
+            return Ok(Some(Zatoshis::ZERO));
+        }
+        let mut total = Zatoshis::ZERO;
+        let mut all_known = true;
+        for txin in &bundle.vin {
+            let outpoint = txin.prevout();
+            let Some(mined) = self.transactions.get(outpoint.txid()) else {
```
juanky201271, line 904: **Inputs are only looked up in mined transactions, so with `transparent_inputs` on, spending an output that's still in the mempool is always rejected as `UnknownInput`.**

The second step of a TEX send spends the first step's output while that step is still in the mempool. So:

- a successful TEX send can't be tested against the strict mock, and
- the failed-step tests can't tell an injected failure from this false rejection, so they may pass for the wrong reason.

`mempool_state` already parses the mempool and download queue, so it could also collect their outputs for this lookup. `spent_outpoints` / `mempool_spends` already handle double spends. zcashd accepts spends of unconfirmed outputs; I believe zebra does too, but I haven't confirmed it.

.github/workflows/ci-pr.yaml
```diff
@@ -184,4 +184,4 @@ jobs:
   test:
     uses: ./.github/workflows/test.yaml
     with:
-      nextest-flags: "-E 'not test(slow)'"
+      nextest-flags: "-E 'not test(slow) and not package(libtonode-tests)'"
```
juanky201271, line 187: This stops running `libtonode-tests` on pull requests. That's the suite that runs against real zainod/zebrad, and the PR description doesn't mention it. It still runs nightly, but breakage would only show up after merging to `dev`. Was this intentional (CI time?)? If so, could you say so in the description, or keep a subset on PRs?

zingolib/src/testutils/mock_indexer.rs
```diff
@@ -781,51 +1692,109 @@ impl CompactTxStreamer for MockIndexerService {
...
-        _request: Request<GetAddressUtxosArg>,
+        request: Request<GetAddressUtxosArg>,
     ) -> Result<Response<Self::GetAddressUtxosStreamStream>, Status> {
-        Err(Status::unimplemented("get_address_utxos_stream"))
+        let address_utxos = self
+            .get_address_utxos(request)
+            .await?
+            .into_inner()
+            .address_utxos;
+        Ok(Response::new(stream_with_fault(address_utxos, None)))
```
juanky201271, line 1779: **Stream faults injected for `Rpc::AddressUtxos` are lost here.**

This calls `get_address_utxos`, which runs `self.fault_for(Rpc::AddressUtxos).await?`. That consumes a queued `TruncateStream`/`EndStream` fault and discards it (only `Fail` propagates). The stream is then built with `None` and comes back complete, even though the fault is gone from the queue. The `Rpc::AddressUtxos` doc says it covers the stream variant too.

Suggestion: take the fault once here and pass it to `stream_with_fault`, with the UTXO collection factored into a helper that doesn't call `fault_for`.

juanky201271 (commented):
Nice work on the consensus checks overall: moving `expire_transactions` after scanning, the fee/expiry/anchor/coinbase-maturity rules and the new failure handling in `send.rs` all look right to me. A few issues below; the first one is in production wallet code.

pepper-sync/src/sync.rs
```diff
@@ -503,8 +503,6 @@ where
     )
     .await?;
 
-    expire_transactions(&mut *wallet.write().await)?;
```
Oscar-Pepper, line 506: where do we now expire transactions during sync?

zingolib/src/lightclient/send.rs
```diff
@@ -1026,11 +1003,7 @@ impl LightClient {
                 .transaction()
                 .write(&mut transaction_bytes)
                 .map_err(|e| {
-                    pepper_sync::set_transactions_failed(
-                        &mut wallet.wallet_transactions,
-                        vec![*txid],
-                    );
-                    wallet.save_required = true;
+                    fail_unsent(&mut wallet, &calculated_txids, index);
```
Oscar-Pepper, line 1006: we cant fail all the txs here. we must only fail the tx that failed. take a tex send as an example:
- the first step is successfuly transmitted
- the second step fails and sets the first to to failed even though its on chain
- the user no longer has the ability to re-shield and recover the lost funds in the ephemeral address

Oscar-Pepper: oh sorry i see the `from` field that only fails all txs after the failed tx. ok this is great!

Oscar-Pepper (dismissed):
i am concerned we are no longer expiring transactions. great work adding test coverage and improving mock indexer

pepper-sync/src/sync.rs
juanky201271: Agree. The only call left is at the end of sync, so the `Shutdown` path and interrupted syncs never expire. Using `fully_scanned_height()` instead of `last_known_chain_height` in `expire_transactions` would make it safe to call again at sync start and after each `process_scan_results`.

dorianvp: thanks! good comments, now addressed

juanky201271 (approved):

PR #2762, opened 2026-09-11
feat: split the send route from the price route behind a transmit policy

dorianvp:
this pr splits the send route from the price route behind a transmit policy

### DIFF
--- a/zingo-cli/src/commands.rs
+++ b/zingo-cli/src/commands.rs
@@ -305,12 +305,8 @@ async fn confirm(lightclient: &mut LightClient) -> Result<String, CommandError> 
 async fn current_price(lightclient: &mut LightClient) -> Result<String, CommandError> {
     match lightclient.update_current_price().await {
         Ok(fetch) => {
-            let route = match &fetch.route {
-                zingolib::lightclient::PriceFetchRoute::Mixnet { via_socks5 } => {
-                    format!("over the mixnet via {via_socks5}")
-                }
-                zingolib::lightclient::PriceFetchRoute::Clearnet => "over clearnet".to_string(),
-            };
+            let zingolib::lightclient::PriceFetchRoute::Mixnet { via_socks5 } = &fetch.route;
+            let route = format!("over the mixnet via {via_socks5}");
             Ok(format!(
                 "current price: {} USD (source: {}, rtt: {} ms, fetched {})",
                 fetch.usd,
@@ -1348,7 +1344,8 @@ fn render_status(
              `network on` to enable the mixnet, or `network off` to use clearnet."
             .to_string(),
         Indicator::SwitchedOff => {
-            "Mixnet Mode: switched off (send and price-fetch use clearnet)".to_string()
+            "Mixnet Mode: switched off (price-fetch refuses; send follows the transmit policy)"
+                .to_string()
         }
         Indicator::Bootstrapping => match bootstrap_detail {
             Some(detail) => format!(
--- a/zingolib/src/lightclient/mixnet.rs
+++ b/zingolib/src/lightclient/mixnet.rs
@@ -1029,16 +1029,53 @@ impl LightClient {
             .and_then(|proxy| proxy.death_report())
     }
 
-    /// Resolve the fail-closed route every mixnet-only surface must obey —
-    /// the mixnet proxy when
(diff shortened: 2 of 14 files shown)


zingolib/src/lightclient/mixnet.rs
```diff
@@ -1029,16 +1029,53 @@ impl LightClient {
...
+    /// resolved. Never persisted: every session starts under
+    /// [`TransmitPolicy::Mixnet`](crate::mixnet::TransmitPolicy).
+    pub fn set_transmit_policy(&self, policy: crate::mixnet::TransmitPolicy) {
+        self.transmit_over_mixnet.store(
+            policy == crate::mixnet::TransmitPolicy::Mixnet,
+            std::sync::atomic::Ordering::Release,
+        );
     }
 
     /// Runs the mixnet liveness probe — concurrent `GetLightdInfo` calls
```
juanky201271, line 1081: Nothing in production ever calls this with `Clearnet`. `MixnetStartPolicy::OptedOutThisSession` (mixnet.rs:925) lands `SwitchedOff` via `disable_mixnet` but leaves the policy at the `Mixnet` default, so the ADR 0024 startup opt-out now refuses every send with `MixnetNotReady::Unattached` instead of transmitting over the indexer. The only caller that sets `Clearnet` is `consent_to_clearnet_for_tests`, which makes the clearnet send route unreachable outside tests.

Either `disable_mixnet` or the `OptedOutThisSession` arm should set `TransmitPolicy::Clearnet`.

zingo-cli/src/commands.rs
```diff
@@ -1348,7 +1344,8 @@ fn render_status(
              `network on` to enable the mixnet, or `network off` to use clearnet."
```
juanky201271, line 1344: Stale now: `network off` calls `go_offline`, which lands `Unattached`, and under this PR `Unattached` refuses send regardless of the policy. The CLI also exposes no way to set the transmit policy, so there is no path from here to clearnet at all. Pinned by `zingo-cli/src/commands/tests.rs:748`.

juanky201271 (commented):
The split itself is coherent: the two resolvers, the `AtomicBool` policy, all three constructors updated, and no stale references to `resolve_route` / `mixnet_route` / `ProbeRequiresMixnet` / `PriceFetchRoute::Clearnet` left in the tree.

The gap is where `SwitchedOff` used to carry the clearnet consent: the new policy is never wired into the acts that produced it. Inline comments cover the functional one and the stale CLI string. Two more, in files this PR doesn't touch, so no inline anchor:

- `zingolib/src/mixnet/mode.rs:102,113` — `MixnetSlot::SwitchedOff` is still documented as "The one slot state that consents to clearnet". After this PR no slot state consents to clearnet.
- `zingolib/src/mixnet/driver.rs:157` — `OptedOutThisSession` still says it records "the same per-session clearnet consent as an in-session toggle-off". Same staleness in the `disable_mixnet` doc at `zingolib/src/lightclient/mixnet.rs:889`.

Oscar-Pepper (dismissed):
UTAck

zingolib/src/lightclient/mixnet.rs
```diff
@@ -922,6 +926,10 @@ impl LightClient {
         match policy {
             crate::mixnet::MixnetStartPolicy::OptedOutThisSession => {
                 self.disable_mixnet().await;
+                // The opt-out is the user's explicit clearnet consent for
+                // this session, so it is the one production act that sets
+                // the policy. The price fetch stays mixnet-only regardless.
+                self.set_transmit_policy(crate::mixnet::TransmitPolicy::Clearnet);
```
juanky201271, line 957: Nothing resets this. Opt out at start, then `enable_mixnet` / `enable_mixnet_via_host` / `attach_mixnet` mid-session: indicator goes `Ready` but sends keep going over clearnet. The enable paths still claim they revoke clearnet consent (L506, L758, `a_failed_attach_revokes_clearnet_consent`), which now only holds for the indicator.

Not blocking (no consumer uses the opt-out today), but mobile will hit the same thing once its toggle-off sets `Clearnet`. Suggest the enable/attach paths set `TransmitPolicy::Mixnet`.

juanky201271 (dismissed):

juanky201271 (dismissed):

juanky201271 (approved):

PR #2763, opened 2026-09-12
fix(summary): attribute a failed transaction's spends to the wallet

### LINKED ISSUE #2756: A failed outgoing transaction is reported as received
## Summary

A failed outgoing transaction is reported as `received`. Direction is inferred from the wallet's recorded spends, and a transaction that never reached a block never gets any — so every failed send, shield or self-send reads to the user as money arriving.

## What a user sees

A wallet holding only transparent funds shields twice; both attempts fail. History then shows two **received** transfers, each for the transparent balance minus the fee, each carrying a transaction fee.

A received transfer that paid a fee is not a thing. That is the tell.

## Why

`transaction_kind` (`zingolib/src/wallet/transaction.rs:174`) calls it `Received` when the wallet spends nothing in it:

```rust
if spends.transparent.is_empty() && spends.spends_no_shielded() && sends_no_notes {
    TransactionKind::Received
}
```

`spends` comes from `find_spends`, which keeps an output only when `output.spending_transaction()` names this txid. That field is written by sync alone, from what it finds on chain — `update_spent_notes` from nullifiers and `update_transparent_spends` from outpoints, both in `pepper-sync/src/sync/spend.rs`.

A `Failed` transaction was never mined, by definition: transmit failed, the mempool refused it, or it expired. There is no chain evidence of its spends, so the field stays empty and the classifier sees a transaction whose outputs the wallet owns and whose inputs it apparently does not.

The wallet has not lost the inputs — it built the thing. `calculate_transaction_fee` reads them straight off the raw transaction and succeeds, which is why the fee is on screen next to the word "received".

## A cheap discriminator

`calculate_transaction_fee` resolves prevouts among the wallet's own transparent outputs and fails when they are not there, so a genuinely received transaction cannot have a computable fee. `kind == Received` together with `Some(fee)` is a contradiction the wallet can already detect.

Whether the fix is to key off that, to treat a `Failed` status as evidence of authorship on its own, or to record intended spends at build time, is yours — this is the report, not the design.

## Scope

Not specific to shielding. Any failed outgoing transaction takes this path.

`remove_failed_transaction` lets the user clear the rows, and refuses anything not `Failed`, so no confirmed history is at risk. Balances are unaffected: `output_spend_status` already maps a failed spender to `Unspent`. The damage is that the user is told they received money they did not.

## Environment

zingo-pc, zingolib pinned to `opreturn_on_proposal` @ `e353931`, mainnet.

dorianvp:
This PR fixes #2756


Oscar-Pepper (approved):
great fix! this will correctly link the outputs that were attempted to be spent after the tx failed instead of the spending txs being cleared which results in incorrect tx kind

PR #2766, opened 2026-09-13
fix(send): derive Destinations from the indexer registry per chain

### LINKED ISSUE #2764: On testnet, transactions sent over the mixnet go to mainnet indexers and never broadcast
On a testnet wallet with Mixnet Mode on, every transaction fails to broadcast with `unknown Sapling anchor`. The same send with Mixnet Mode off broadcasts fine.

### Cause

[`mixnet_escalating_transmit`](https://github.com/zingolabs/zingolib/blob/3035fe5d53d3a2bca76cfef221789f964ca7b388/zingolib/src/lightclient/send.rs#L354-L368) draws its destinations from [`eligible_destinations`](https://github.com/zingolabs/zingolib/blob/3035fe5d53d3a2bca76cfef221789f964ca7b388/zingolib/src/destination.rs#L177-L190), which has no chain parameter and builds the pool from [`DESTINATION_INDEXERS`](https://github.com/zingolabs/zingolib/blob/3035fe5d53d3a2bca76cfef221789f964ca7b388/zingolib/src/destination.rs#L108-L138), a mainnet-only list. A testnet transaction is therefore sent to mainnet indexers, which reject its anchor, and the testnet sync indexer is excluded from the pool by design (ADR 0022), so no testnet node ever receives it.

The clearnet route submits to the configured indexer and is not affected.

### Reproduce

1. Testnet wallet syncing from `testnet.zec.rocks`, Mixnet Mode on.
2. Send to a single recipient.

### Observed (trimmed)

```
Send error. ← Transmission error. ← Transmission failed. no indexer confirmed delivery after contacting 6 of them:
lwd.z0n.jp: failed at remote-http: indexer rejected the transaction: ... unknown Sapling anchor: Root("bd6af55e…")
zec-node.cakewallet.com: failed at remote-http: ... unknown Sapling anchor
us.zec.stardust.rest: failed at remote-http: ... unknown Sapling anchor
zcashlw.devshore.ovh: failed at tunnel-transport: deadline has elapsed
webhighway.website: failed at tunnel-transport: tls handshake eof
znode.roamerx.win: failed at tunnel-transport: tls handshake eof
```

### Expected

The destination pool follows the wallet's chain. There is no testnet destination list today, so testnet needs one, or an explicit decision about what a mixnet transmission does there; it should not reach mainnet indexers.

Found in zingo-pc on `dev` (`3035fe5`).

dorianvp:
- Fix: testnet sends over the mixnet reach testnet indexers.
- Destinations come from the indexer registry.
- Indexers are classified by role, trust, and location.
- Local indexers are trusted; remote indexers are untrusted on mainnet by default.
- A reachable trusted indexer receives broadcasts alone.
- Configured broadcast indexers receive sends first.
- Mixnet sends skip local indexers.
- Migration refuses a local target over the mixnet.
- CLI transmit JSON: the clearnet `indexer` key is renamed to `destination`.
- Fixes #2764.


juanky201271:
With the mixnet off, a user syncing against their own node now broadcasts through public registry operators instead of that node. That's a privacy regression for self-hosters until the trusted set is wired.

The ADR plans the settings toggle for mobile only; zingo-pc lets users pick a custom server too, so desktop needs it as well. Could `ClientConfig` feed `with_trusted` in this PR or right after, so both apps can expose it?

juanky201271 (approved):
great job!

PR #2768, opened 2026-09-14
fix: max_send_value TEX

### LINKED ISSUE #2767: max_send_value returns 0 for TEX addresses
Reported in zingolabs/zingo-mobile#1400.

`max_send_value` returns 0 when the recipient is a TEX address, so the app shows 0 available and blocks any send to TEX.

**Cause**

`recipient_amount` (`wallet/propose.rs`) reads payment 0 of the first step only:

```rust
proposal.steps().first().transaction_request().payments().get(&0)
```

For TEX, `zcash_client_backend` builds a 2-step proposal and removes the TEX payment from `tr0` (it goes to `tr1`, spending the ephemeral output). So `get(&0)` is `None` → `Zatoshis::ZERO`.

The zennies branch of `create_send_all_proposal` hits the same thing (sizing gives 0 → `InsufficientFunds`). Also `input_total` there only counts the first step fee, missing the `tr1` fee for TEX.

**Fix**

- Look up the recipient payment across all steps (for TEX it's in the last step).
- Include every step's fee in `input_total`.
- Add a `max_send_value` test to a TEX address (`send_to_tex` can be the base).

dorianvp:
Fixes #2767. Introduced in #2759.


juanky201271 (approved):
nicely done.

PR #2769, opened 2026-09-16
fix(migrate): report a refused part submission as Failed, not Slid

juanky201271:
When the endpoint refuses a migration part's submission, `transmit_due_parts_selected` logged a warning and returned no txids, so `execute_due_parts` reported the part as `Slid` with `halted: None`. The transaction was already built and recorded (clients show it as pending), so users were told the batch went out or slid while nothing left, and only a manual resend recovered it.

- The loop now returns the submission error with the sent txids.
- `execute_due_parts` reports that part as `Failed` and halts, as `PartSendResult::Failed` documents. The part stays `Signed` and due, so the next batch resubmits it.
- `transmit_due_parts` and `catch_up_migration` behave as before. No public type changes.

Test: `a_refused_submission_is_reported_failed_not_slid` (refused → `Failed` + halt, part still `Signed`; endpoint recovers → `Sent`).


zingolib/src/lightclient/migrate.rs
```diff
@@ -483,6 +483,16 @@ pub struct BatchReport {
     pub halted: Option<String>,
 }
 
+/// What one pass of the due-part transmission loop achieved.
+struct SelectedTransmission {
```
dorianvp: Hmm I think that this makes it possible to represent an invalid state. what about something like this:

```rust
#[must_use]
enum Transmission {
    Complete(Vec<TxId>),
    Halted {
        sent: Vec<TxId>,
        part: PartId,
        error: PartTransmissionError,
    },
}
```

juanky201271: Done in 3f932391f: `Transmission` enum as suggested. `Halted` carries the failed part and the batch report takes its id from there.

dorianvp (approved):

PR #2771, opened 2026-09-18
feat(migrate)!: replace stepping migration API with explicit functions

dorianvp:


dorianvp:
Converting to draft due to concern around code size. It may be a much better idea to simply create a separate crate for handling the ZIP-318 private migration path.

PR #2773, opened 2026-09-20
feat(lightclient): add sync status subscription and shared-reference sync waiting

dorianvp:


zingolib/src/lightclient.rs
```diff
@@ -263,17 +263,14 @@ impl LightClient {
...
-                    if !overwrite && config.get_wallet_path().exists() {
-                        return Err(LightClientError::FileError(std::io::Error::new(
-                            std::io::ErrorKind::AlreadyExists,
-                            format!(
-                                "Cannot save to given data directory as a wallet file already exists at:\n{}",
-                                config.get_wallet_path().display()
-                            ),
-                        )));
-                    }
+                if !overwrite && config.get_wallet_path().exists() {
```
juanky201271: Removing the ios/android cfg guard is unrelated to this PR and breaks mobile: zingo-mobile always calls `LightClient::new(config, false)`, and the salvage flow restores from seed with the damaged wallet file still on disk → `AlreadyExists`. Split it out or handle the file on the mobile side first.

zingolib/src/lightclient/sync.rs
```diff
@@ -210,6 +220,16 @@ impl LightClient {
         }
     }
 
+    /// Waits until the engine of the Sync Session the latest [`Self::sync`] launched has
+    /// returned, and resolves at once when no engine runs.
+    ///
+    /// The session's result remains with the client for [`Self::await_sync`], which collects
+    /// it and refreshes the migration part witnesses.
+    pub async fn wait_for_sync(&self) {
```
juanky201271: nit: never resolves while sync is paused (sender stays alive). Worth a line in the doc. Also `wait_for_sync` vs `await_sync` is easy to mix up.

juanky201271 (approved):

